PR #3, opened 2018-09-20
Dev cc docker

MrColeC:
Addressing shortfalls in the Dockerfile that were encountered when building the image within a different environment (Ubuntu 18.04 vs Mac)


MrColeC:
FYI @arose

PR #6, opened 2018-11-30
Lingrad2

luna215:
# Summary 
Added linear-gradient to the color-scale parameters.

# Example
screen shot 2018-11-30 at 9 45 03 am


arose:
NIce, thanks @luna215!

PR #8, opened 2018-12-14
User can continue dragging after leaving canvas & Bug fixes

luna215:
# Summary

User is able to continue dragging a point while they go out the bounds of the canvas. I also improved performance while user is dragging point, and you no longer get pauses.


arose:
thanks, next year lets revisit the handling of the ghost point

luna215:
@arose sure thing. Anything advise to approach it differently?

arose:
ask me again next year :)

Issue #15, opened 2019-04-05
Interval quantization encoder should use max as upper bound


JonStargaryen:
The interval quantization encoder uses numSteps as upper bound.

https://github.com/molstar/molstar-proto/blob/50ebd5467cbf80da6898bf307a8062f9cd300b6d/src/mol-io/common/binary-cif/array-encoder.ts#L175

The line should probably read
`else if (v >= max) output[i] = max; `
otherwise the example at https://github.com/dsehnal/BinaryCIF#interval-quantization cannot be recreated.

arose:
ping @dsehnal, we just discussed this and now think it should be `numSteps - 1`. Do you agree?

dsehnal:
Yes, you are right, nice find!

PR #16, opened 2019-06-05
fixes upper bound in interval quantization encoding

JonStargaryen:
Fix for #15 - use `numSteps - 1` during IntervalQuantization.


arose:
That is the same bug you fixed in the Java impl already, right?

arose:
Thanks @JonStargaryen

JonStargaryen:
> That is the same bug you fixed in the Java impl already, right?

Yeah, Mol* and the Java implementation are now in agreement.

PR #17, opened 2019-06-05
Encoding and precision config

JonStargaryen:
- allows to specify encoding strategy for arbitrary columns
- specify number of decimal places for specific float columns

Changes/implementation:
- adds a browser-test in encoding-config
- creates an `EncodingProvider` from a set of instructions

Things to discuss:
- use of precision property: right now it has to specified to identify float columns - could make it optional, so that precision is inferred if needed
- encoding could be optional: specify only the precision and let the encoding chain be determined automatically


src/mol-io/writer/cif.ts
```diff
@@ -53,5 +53,58 @@ export namespace CifWriter {
...
+                    const hint = hints[i];
+                    if (hint.categoryName === c && hint.columnName === f) {
+                        return resolveEncoding(hint);
+                    }
+                }
+            }
+        }
     }
+
+    function resolveEncoding(hint: EncodingStrategyHint): ArrayEncoder | undefined {
```
arose: You can add

```
type EncodingType = 'pack' | 'rle' | 'delta' | 'delta-rle'
```

and then you can return `ArrayEncoder` because via the switch statement the compiler will check that all cases of the union type are handled.

JonStargaryen: Addressed in 
df6b163

JonStargaryen:
Other changes:

- drop browser-test, created .spec
- added filtering by white-/blacklisting categories and columns
- removed uppercase String parameters

src/mol-io/writer/cif.ts
```diff
@@ -53,5 +53,60 @@ export namespace CifWriter {
...
+                    return E.by(E.delta).and(E.integerPacking);
+                case 'delta-rle':
+                    return E.by(E.delta).and(E.runLength).and(E.integerPacking);
+            }
+        }
+        throw new Error('cannot be reached');
+    }
+}
+
+// defines the information needed to encode certain fields: category and column name as well as encoding tag, precision is optional and identifies float columns
```
arose: tip: if you want a comment to show up in editor tooltip as documentation for a function, field, ... use the /** */ style comments.

JonStargaryen: Done in 4fa04f0

JonStargaryen:
Removed the json-definition of category/field filtering as discussed. Now behaves like the original filtering text files: specify `category_name.field_name` to whitelist stuff. Retained previous support from json-impl to allow whitelisting by `category_name` alone. Furthermore, categories/fields can be explicitly blacklisted by starting lines with `!`.

arose:
Thanks @JonStargaryen!

PR #35, opened 2020-04-24
Forkdev

corredD:
Support both full membrane and struct_assembly of each residues. Example is "Influenza complete".
Support for ingredient colors if provided.
I created a PluginStateTransform to generate and group all the assembly but I am not sure if it's the best way to do it.
I will clean up the other model to use the light membrane files.


src/extensions/cellpack/state.ts
```diff
@@ -94,4 +96,113 @@ const StructureFromCellpack = PluginStateTransform.BuiltIn({
             }
         }
     }
-});
\ No newline at end of file
+});
+
+export { GetAllAssamblyinOne };
```
arose, line 101: this one is not used, right?

src/extensions/cellpack/model.ts
```diff
@@ -108,7 +109,7 @@ async function getStructure(plugin: PluginContext, model: Model, source: Ingredi
     }
     let query;
     if (source.selection){
-        const asymIds: string[] = source.selection.replace(' :', '').split(' or');
+        const asymIds: string[] = source.selection.replace(' ', '').replace(':', '').split('or');
```
arose, line 112: How about changing, in the files, the `selection` property to `authAsymIds` with type `string[]`? Or do you need it compatible with something else?

src/extensions/cellpack/state.ts
corredD: Its not, I wasnt sure if it was best to do it from the Model or the Structure. You can also change the name if you want.

src/extensions/cellpack/model.ts
corredD: Definitly, I will have to make sure first that all the pipeline will work without issues. How this will look like in json ? like this : {"selection":["A","B",...]} ?

arose:
Hi @corredD, merged it, thank you.

I separated the color themes into two, one for the provided color (is there a model that provides colors?) case and one for generated colors. For `Ingredient.confidence` we can add another color theme. Note the the ColorThemeProviders can have a isApplicable method to show only the color themes that are applicable in the UI.

For the Structure from all assemblies I changed it to use the one creating it from the model, no need to create a structure first.

PR #47, opened 2020-05-19
ANVIL / Membrane placement prediction implementation

JonStargaryen:
It's been some time that we talked about functionality to predict the placement of the membrane layer for transmembrane proteins.

This implements the ANVIL algorithm. There are a number of TODOs in the code. Questions include:
- ANVIL requires predicted or computed ASA values - what is the most elegant way of requesting that information?
- Should there be a dedicated representation for membranes? That would also be nice to have for the creation of static Mol* images with membrane layers (e.g., derived from the OPM database).


arose:
Nice, quick comments below

>ANVIL requires predicted or computed ASA values - what is the most elegant way of requesting that information?

Calculate (via attach) and get it 'inline'.

```ts
await AccessibleSurfaceAreaProvider.attach(ctx, structure);
const asa = AccessibleSurfaceAreaProvider.get(structure).value;
```

> Should there be a dedicated representation for membranes? That would also be nice to have for the creation of static Mol* images with membrane layers (e.g., derived from the OPM database).

How about a `MembraneOrientationProvider` with a param for where/how to get/calculate it like we do in `SecondaryStructureProvider`. Then we can have a `MembraneRepresentation = StructureRepresentationProvider({ ... })` that obtains the data via the `MembraneOrientationProvider`.

src/mol-model-props/computed/membrane-orientation/ANVIL.ts
```diff
@@ -0,0 +1,350 @@
...
+                hphob++;
+            } else {
+                hphil++;
+            }
+        }
+        return of(hphob, hphil);
+    }
+
+    // ANVIL-specific (not general) definition of membrane-favoring amino acids
+    const HYDROPHOBIC_AMINO_ACIDS = ['ALA', 'CYS', 'GLY', 'HIS', 'ILE', 'LEU', 'MET', 'PHE', 'SER', 'THR', 'VAL'];
```
arose, line 339: maybe use a `Set`

JonStargaryen: Good call, realized in efc5bf6

src/extensions/membrane-orientation/representation.ts
```diff
@@ -0,0 +1,80 @@
...
+    return PD.clone(MembraneOrientationParams);
+}
+
+export type MembraneOrientationRepresentation = StructureRepresentation<MembraneOrientationParams>
+export function MembraneOrientationRepresentation(ctx: RepresentationContext, getParams: RepresentationParamsGetter<Structure, MembraneOrientationParams>): MembraneOrientationRepresentation {
+    return Representation.createMulti('Membrane Orientation', ctx, getParams, StructureRepresentationStateBuilder, MembraneOrientationVisuals as unknown as Representation.Def<Structure, MembraneOrientationParams>);
+}
+
+const MembraneOrientationVisuals = {
+    'membrane-orientation': (ctx: RepresentationContext, getParams: RepresentationParamsGetter<MembraneOrientation, MembraneOrientationParams>) => ShapeRepresentation(getMembraneSpheres, Spheres.Utils),
```
arose, line 42: any chance you could do
- bilayer-spheres (what you have now)
- bilayer-planes (two planes, mesh geo)
- bilayer-rims (two rims, lines geo)

For planes and rims, see the https://github.com/molstar/molstar/blob/master/src/mol-geo/primitive/polygon.ts to get points on a circle. Planes and rims are separate so that we can have a transparent plane and an opaque plane. See https://github.com/molstar/molstar/blob/master/src/mol-repr/shape/loci/angle.ts for how overwrite the opacity/alpha for a single visual.

src/mol-model-props/computed/membrane-orientation/ANVIL.ts
arose: thanks

src/extensions/membrane-orientation/representation.ts
arose: please :)

src/mol-model-props/computed/membrane-orientation.ts
```diff
@@ -0,0 +1,54 @@
...
+import { Structure } from '../../mol-model/structure';
+import { CustomStructureProperty } from '../common/custom-structure-property';
+import { CustomProperty } from '../common/custom-property';
+import { CustomPropertyDescriptor } from '../../mol-model/custom-property';
+import { ANVILParams, ANVILProps, computeANVIL } from './membrane-orientation/ANVIL';
+import { AccessibleSurfaceAreaProvider } from './accessible-surface-area';
+import { MembraneOrientation } from '../../mol-model/structure/model/properties/membrane-orientation';
+
+function getMembraneOrientationParams(data?: Structure) {
+    let defaultType = 'anvil' as 'anvil' | 'db'; // TODO flip - OPM (or some other db-source) should be default if PDB identifier is known
```
arose, line 18: see https://github.com/molstar/molstar/blob/master/src/mol-model-props/computed/secondary-structure.ts#L18-L38 for example code

src/extensions/membrane-orientation/representation.ts
JonStargaryen: Sure thing, I'll have a go at it.

src/mol-model-props/computed/membrane-orientation.ts
JonStargaryen: Is there an example that shows how parameters can be passed to this kind of provider? Along the lines of: `await MembraneOrientationProvider.attach(ctx, structure, ...);` The signature calls for `Mapped` and `NamedParams` but I struggle to construct the correct parameter object.

arose: for

```js
{
        type: PD.MappedStatic(defaultType, {
            'model': PD.EmptyGroup({ label: 'Model' }),
            'dssp': PD.Group(DSSPComputationParams, { label: 'DSSP', isFlat: true })
        }, { options: [['model', 'Model'], ['dssp', 'DSSP']] })
}
```

it is

```js
{ type: { name: 'model', params: {} } }
```

and

```js
{ type: { name: 'dssp', params: { oldDefinition: true, oldOrdering: true } } }
```

JonStargaryen:
Planes and rim visuals are in place. I think a combination of both provides a really nice result. Some basic options are available to modify the visuals.

example

src/mol-model-props/computed/membrane-orientation.ts
JonStargaryen: Oh, I wanna say duh. Didn't realize that it's so straight forward to pass the params in. Thanks!

src/extensions/membrane-orientation/representation.ts
```diff
@@ -0,0 +1,168 @@
...
+};
+export type BilayerPlanesParams = typeof BilayerPlanesParams
+export type BilayerPlanesProps = PD.Values<BilayerPlanesParams>
+
+const BilayerRimsParams = {
+    ...SharedParams,
+    ...Lines.Params,
+    lineSizeAttenuation: PD.Boolean(true),
+    linesSize: PD.Numeric(1, { min: 0.01, max: 50, step: 0.01 }),
+    alternate: PD.Boolean(true)
```
arose, line 55: maybe 'dashes' or 'dashedLines' would more telling?

src/mol-model-props/computed/membrane-orientation/ANVIL.ts
```diff
@@ -0,0 +1,349 @@
...
+import { AccessibleSurfaceAreaProvider } from '../accessible-surface-area';
+import { ANVILContext } from './anvil/common';
+import { MembraneOrientation } from '../../../mol-model/structure/model/properties/membrane-orientation';
+
+export const ANVILParams = {
+    numberOfSpherePoints: PD.Numeric(120, { min: 35, max: 700, step: 1 }, { description: 'Number of spheres/directions to test for membrane placement. Original value is 350.' }),
+    stepSize: PD.Numeric(1, { min: 0.25, max: 4, step: 0.25 }, { description: 'Thickness of membrane slices that will be tested' }),
+    minThickness: PD.Numeric(20, { min: 10, max: 30, step: 1}, { description: 'Minimum membrane thickness used during refinement' }),
+    maxThickness: PD.Numeric(40, { min: 30, max: 50, step: 1}, { description: 'Maximum membrane thickness used during refinement' }),
+    afilter: PD.Numeric(40, { min: 10, max: 100, step: 1 }, { description: 'Absolute ASA cutoff above which residues will be considered' })
```
arose, line 25: may `asaCutoff`?

arose: or `minAccessibleSurfaceArea`

arose:
How about adding a preset "Annotation/Membrane Orientation" with the membrane orientation shown and the structure colored by hydrophobicity?

I was thinking, maybe this could be added just with the ANVIL implementation and no external call/dependency. For integration with RCSB, there could be an extension `rcsb/membrane-orientation` that makes calls to rcsb apis to get more infos before showing the membrane orientation, like checking if a structure is actually a membrane protein or taking the orientation from a homologous structure... The rcsb extension could then make use of the types and representation already available in this PR.

src/mol-model-props/computed/membrane-orientation/ANVIL.ts
JonStargaryen: Changed to `asaCutoff` in 72bdd5b

src/extensions/membrane-orientation/representation.ts
JonStargaryen: Renamed to `dashedLines` in 255cc62

JonStargaryen:
> How about adding a preset "Annotation/Membrane Orientation" with the membrane orientation shown and the structure colored by hydrophobicity?
I've created a present/behavior for that. Is there a simple way to test that code or make the present appear in the UI? Tried adding it as extension to viewer/index.ts but no dice.

> I was thinking, maybe this could be added just with the ANVIL implementation and no external call/dependency. For integration with RCSB, there could be an extension `rcsb/membrane-orientation` that makes calls to rcsb apis to get more infos before showing the membrane orientation, like checking if a structure is actually a membrane protein or taking the orientation from a homologous structure... The rcsb extension could then make use of the types and representation already available in this PR.
Yeah, this seems like the way to go. I've removed the param stub for a future api/database integration again. So, now there is only ANVIL supported and rcsb-specific extensions can be added in the future.

src/extensions/membrane-orientation/behavior.ts
```diff
@@ -0,0 +1,124 @@
...
+    category: 'custom-props',
+    display: {
+        name: 'Membrane Orientation',
+        description: 'Initialize orientation of membrane layers. Data calculated with ANVIL algorithm.' // TODO add ' or obtained via RCSB PDB'
+    },
+    ctor: class extends PluginBehavior.Handler<{ autoAttach: boolean }> {
+        private provider = MembraneOrientationProvider
+
+        register(): void {
+            DefaultQueryRuntimeTable.addCustomProp(this.provider.descriptor);
```
arose, line 36: this (and also in unregister) is only needed when you actually have `this.provider.descriptor.symbols` defined (see rcsb/validation-report for example)

arose: How about moving the `isTransmembraneSymbol` to the file with the provider following the setup as rcsb/validation-report. Also, add the symbol to the custom-property Descriptor there.

arose:
Could you move `computed/membrane-orientation.ts`, `computed/membrane-orientation/ANVIL.ts`, `computed/membrane-orientation/anvil/common.ts` and `structure/model/properties/membrane-orientation.ts` also to `extensions/membrane-orientation/*`, so everything but the `browser/test`.

> I've created a present/behavior for that. Is there a simple way to test that code or make the present appear in the UI? Tried adding it as extension to viewer/index.ts but no dice.

Adding it to https://github.com/molstar/molstar/blob/master/src/apps/viewer/index.ts#L34-L40 should work. Is that what you tried? The preset should show up under `Components/Preset/Annotation`.

Screen Shot 2020-06-13 at 6 25 32 PM

Regarding the extra Hydrophobicity color theme. Do you think that is required or would it be enough to reuse the existing Hydrophobicity theme?

JonStargaryen:
I moved everything to `extensions`. The symbols now live with the property provider.

> Adding it to https://github.com/molstar/molstar/blob/master/src/apps/viewer/index.ts#L34-L40 should work. Is that what you tried? The preset should show up under `Components/Preset/Annotation`.

Tried that initially. It just didn't show up due to some error.

> Regarding the extra Hydrophobicity color theme. Do you think that is required or would it be enough to reuse the existing Hydrophobicity theme?

I think it's perfectly fine to use the existing color theme. However, when doing so it's not possible to explicitly register that color theme because it was added elsewhere (causing an error: `hydrophobicity already registered. at ThemeRegistry../lib/mol-theme/theme.js.ThemeRegistry.add`). Is it safe to assume that the `colorThemeRegistry` will be always be aware of the `hydrophobicity` color theme or should I add an explicit check to add it if needed?

JonStargaryen:
Furthermore, the `hydrophobicity` and `residue-name` color themes were broken after `label_comp_id` was moved from `StructureProperties.residue` to `StructureProperties.atom`. 4c0d138 should fix that.

Before:
coloring

After:
fixed

arose:
> Is it safe to assume that the colorThemeRegistry will be always be aware of the hydrophobicity color theme or should I add an explicit check to add it if needed?

Yeah, that is fine. The build-in names should by type checked, so if it gets removed for some reason there would be a type error.

arose:
> Furthermore, the hydrophobicity and residue-name color themes were broken after label_comp_id was moved from StructureProperties.residue to StructureProperties.atom. 4c0d138 should fix that.

thanks!!!

arose:
> Tried that initially. It just didn't show up due to some error.

So it works now?

arose:
Thanks, this is really great. Did a few tweaks, see 181646f. There seems to be an issue with the spheres-based visual.

JonStargaryen:
> Did a few tweaks, see [181646f](https://github.com/molstar/molstar/commit/181646f0527be1dd24b702fbae8c53421dfe388f).

Thanks a lot for your refinements!

> So it works now?

It should. 

What still annoys me is that the results are somewhat unpredictable for non-membrane proteins. Sometimes nothing will be shown, in other cases the membrane orientation will appear (but, of course, be meaningless). I tried to use embedding statistics and the obtained scores to make an educated guess. However, that's not trivial.

> There seems to be an issue with the spheres-based visual.

With the current implementation, the spacing between spheres depends on the orientation of the membrane bilayers. I'd propose to remove the sphere visual for now and reintroduce it to visualize data from OPM or MemProtMD (where individual spheres actually correspond to lipids of the bilayer). The rims/planes visuals that you proposed look way more modern.

PR #71, opened 2020-06-26
SDF/MOL/MOL2 export by ModelServer

JonStargaryen:
**Aim:** 
Use ModelServer to export a single selected residue/ligand as SDF/MOL/MOL2

**Changes:**
- extended `ResponseFormat` by `Encoding` property that can be `CIF` | `BCIF` | `SDF` | `MOL` | `MOL2` - default value is still CIF
- new `ligand` query type
- SDF/MOL/MOL2 encoder that resemble original CIF encoder
- SDF reader (not needed but for good measure)

**Problems/Questions/Improvements:**
- How to use custom properties in ModelServer/config? I'd propose an optional argument to specify the path to `ccb.bcif`. If it's absent no ligand export is available.
- Is it possible to use the query itself to ensure only 1 residue is returned? Right now, I check residue identifiers but there might be a more simple way.
- Only coordinates and bond order are currently being exported. Nice to have would be stereo-flags, charge information, and orbital annotations.
- Would be nice to enable `SDF` | `MOL` | `MOL2` only for `ligand` queries

Feedback would be highly appreciated when you find the time. Guess there are quite some things that could be refined in this draft.


dsehnal:
Hi Sebastian,

nice update.

> 
> **Problems/Questions/Improvements:**
> 
>     * How to use custom properties in ModelServer/config? I'd propose an optional argument to specify the path to `ccb.bcif`. If it's absent no ligand export is available.

Would adding a "config variables" array property to the config help solve this issue? I.e. there would be a global object as part of ``mol-model`` which would set these variables which custom properties would then be able access? Similar to this https://github.com/molstar/molstar/blob/master/src/mol-plugin/config.ts

> 
>     * Is it possible to use the query itself to ensure only 1 residue is returned? Right now, I check residue identifiers but there might be a more simple way.

Yeah, this can be done by using "group-by" in the atom groups query with ``residueKey`` and then ``first`` query on top of that.

David

JonStargaryen:
Thank you @dsehnal 

> Would adding a "config variables" array property to the config help solve this issue? I.e. there would be a global object as part of `mol-model` which would set these variables which custom properties would then be able access? Similar to this https://github.com/molstar/molstar/blob/master/src/mol-plugin/config.ts

It's all good. I wrongly assumed that the ModelServer config must specify the `chemCompBondTablePath` in order to access component-bond data. It works out of the box.

> Yeah, this can be done by using "group-by" in the atom groups query with `residueKey` and then `first` query on top of that.

That's convenient. Done in 61390cb and 9bde4c4 respectively. 

Out of curiosity, is it possible to make the selection unique with respect to model_nums (i.e. return only the first group if there are multiple models)? I tried using `unitTest: ctx => StructureProperties.unit.model_num(ctx.element) === 1` but that will leave behind PDB entries such as 1ezc that don't start with model_num 1.

dsehnal:
> Out of curiosity, is it possible to make the selection unique with respect to model_nums (i.e. return only the first group if there are multiple models)? I tried using `unitTest: ctx => StructureProperties.unit.model_num(ctx.element) === 1` but that will leave behind PDB entries such as 1ezc that don't start with model_num 1.

I suppose the best way would be to modify https://github.com/molstar/molstar/blob/master/src/servers/model/server/api.ts#L36 and pass all the models as 3rd argument (or modify it to takes single object param with multiple arguments).

That way you could decide to return empty result for certain models.

JonStargaryen:
Followed the 3rd argument suggestion. Now `model_nums` will be honored if present, otherwise only first model will be considered.

dsehnal:
Looks good. I've added a couple of comments to the code and after should be ready to merge.

JonStargaryen:
Sorry, where are your comments at @dsehnal ? Can't seem to find/see them.

dsehnal:
The code review for the request.

- remove the single residue check in the writer
- move the exporters to mol-io and/or mol-model/structure/export

JonStargaryen:
Redundant check removed in 37fdbfe. Moved all encoders to `mol-io` as of latest commit.

Let me know if there's anything else to do. Can you please create a release to npm after merging? Thank you in advance!

PR #77, opened 2020-08-10
Ext dev

MadCatX:
First commit is a bit of a code drop but it mostly just moves some code around and replaces hardcoded values with user parameters.

Second commit fixes missing pyramids on some unusual structures such as [1jbs - https://blackbox.ibt.biocev.org/devel_molstar/RCSB/cif_dnatco_updated/1jbs_v32C35A23.cif.gz(https://blackbox.ibt.biocev.org/devel_molstar/RCSB/cif_dnatco_updated/1jbs_v32C35A23.cif.gz)

Third commit adds labels with NtC information for highlighted steps. The code requires for the given loci to represent a valid step and the only way how to highlight a valid step from the UI right now is to pick the connecting stick between two residues.

I have some other work in the pipe that will make this code more useful but they require a bit more work and perhaps some discussion. I figured that merging this first would make sense.

### DIFF
--- a/src/extensions/dnatco/common.ts
+++ b/src/extensions/dnatco/common.ts
@@ -1,0 +1,111 @@
+/**
+ * Copyright (c) 2018-2020 mol* contributors, licensed under MIT, See LICENSE file for more info.
+ *
+ * @author Michal Malý <michal.maly@ibt.cas.cz>
+ * @author Jiří Černý <jiri.cerny@ibt.cas.cz>
+ */
+
+import { Column, Table } from '../../mol-data/db';
+import { toTable } from '../../mol-io/reader/cif/schema';
+import { Model } from '../../mol-model/structure';
+import { MmcifFormat } from '../../mol-model-formats/structure/mmcif';
+import { Color } from '../../mol-util/color';
+
+export namespace DnatcoCommon {
+    export const AColor = Color(0xFFC1C1);
+    export const BColor = Color(0xC8CFFF);
+    export const BIIColor = Color(0x0059DA);
+    export const miBColor = Color(0x3BE8FB);
+    export const ZColor = Color(0x01F60E);
+    export const ICColor = Color(0xFA5CFB);
+    export const OPNColor = Color(0xE90000);
+    export const SYNColor = Color(0xFFFF01);
+    export const NColor = Color(0xF2F2F2);
+
+    export interface NtCObject {
+        PDB_model_number: number,
+        name: string,
+        auth_asym_id_1: string,
+        auth_seq_id_1: number,
+        label_comp_id_1: string,
+        label_alt_id_1: string,
+        PDB_ins_code_1: string,
+        auth_asym_id_2: string,
+        auth_seq_id_2: number,
+        label_comp_id_2: string,
+        label_alt_id_2: string,
+        PDB_ins_code_2: string,
+        confal_score: number,
+        NtC: string
+    }
+
+    export const CifSchema = {
+        ndb_struct_ntc_step: {
+            id: Column.Schema.int,
+            name: Column.Schema.str,
+            PDB_model_number: Column.Schema.int,
+            label_entity_id_1: Column.Schema.int,
+            label_asym_id_1: Column.Schema.str,
+            label_seq_id_1: Column.Schema.int,
+            label_comp_id_1: Column.Schema.str,
+            label_alt_id_1: Column.Schema.str,
+            label_entity_id_2: Column.Schema.int,
+            label_asym_id_2: Column.Schema.str,
+            label_seq_id_2: Column.Schema.int,
+            label_comp_id_2: Column.Schema.str,
+            label_alt_id_2: Column.Schema.str,
+            auth_asym_id_1: Column.Schema.str,
+            auth_seq_id_1: Column.Schema.int,
+            auth_asym_id_2: Column.Schema.str,
+            auth_seq_id_2: Column.Schema.int,
+            PDB_ins_code_1: Column.Schema.str,
+            PDB_ins_code_2: Column.Schema.str,
+        },
+        ndb_struct_ntc_step_summary: {
+            step_id: Column.Schema.int,
+            assigned_CANA: Column.Schema.str,
+            assigned_NtC: Column.Schema.str,
+            confal_score: Column.Schema.int,
+            euclidean_distance_NtC_ideal: Column.Schema.float,
+            cartesian_rmsd_closest_NtC_representative: Column.Schema.float,
+            closest_CANA: Column.Schema.str,
+            closest_NtC: Column.Schema.str,
+            closest_step_golden: Column.Schema.str
+        }
+    };
+
+    export function getCifData(model: Model) {
+        if (!MmcifFormat.is(model.sourceData)) throw new Error('Data format must be mmCIF');
+        if (!hasNdbStructNtcCategories(model)) return undefined;
+        return {
+            steps: toTable(CifSchema.ndb_struct_ntc_step, model.sourceData.data.frame.categories.ndb_struct_ntc_step),
+            stepsSummary: toTable(CifSchema.ndb_struct_ntc_step_summary, model.sourceData.data.frame.categories.ndb_struct_ntc_step_summary)
+        };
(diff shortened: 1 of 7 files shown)


arose:
Hi @MadCatX, I will have a look later this week.

src/extensions/dnatco/confal-pyramids/color.ts
```diff
@@ -5,152 +5,50 @@
...
+    Z: PD.Color(Color(0x01F60E)),
+    IC: PD.Color(Color(0xFA5CFB)),
+    OPN: PD.Color(Color(0xE90000)),
+    SYN: PD.Color(Color(0xFFFF01)),
+    N: PD.Color(Color(0xF2F2F2))
+};
 export type ConfalPyramidsColorThemeParams = typeof ConfalPyramidsColorThemeParams
 export function getConfalPyramidsColorThemeParams(ctx: ThemeDataContext) {
-    return ConfalPyramidsColorThemeParams; // TODO return copy
+    return PD.clone(ConfalPyramidsColorThemeParams);
```
dsehnal, line 44: You don't need to clone the param definition, it's never mutated.

src/extensions/dnatco/confal-pyramids/util.ts
```diff
@@ -296,4 +310,95 @@ export namespace ConfalPyramidsUtil {
...
+        }
+    }
+
+    export function lociToPyramid(loci: Loci) {
+        if (loci.kind !== 'element-loci') return void 0;
+        if (loci.elements.length < 1) return void 0;
+        if (!DC.isApplicable(loci.structure.model)) return void 0;
+        if (!Unit.isAtomic(loci.elements[0].unit)) return void 0;
+
+        return (new PyramidGetter(loci.elements[0].unit, loci)).get();
```
dsehnal, line 402: This seems terribly inefficient, triggering very non-trivial computation each time a property is accessed.

A better solution is to create a ``FormatPropertyProvider`` which gets assigned the first time the user tries to access it which contains some index computed from the ``ndb_struct_ntc_*`` categories.

dsehnal (changes requested):
``lociToPyramid`` should be fixed, it's too much work to do each time a label is requested.

src/extensions/dnatco/confal-pyramids/color.ts
MadCatX: Okay.

src/extensions/dnatco/confal-pyramids/util.ts
MadCatX: I took a look at `FormatPropertyProvider` and I am not sure how would it help me in this particular situation.

In order to fetch a pyramid the code has to check if a given `Loci` represents a valid NtC step. Since this check allows me to get `ResidueIndex` of both residues I suppose a could create a map with indices of both residues as key and map them to pyramids. Do you have something this in mind?

MadCatX (commented):
Hi @dsehnal , thanks a lot for your comments! I will see what I can do to address them

src/extensions/dnatco/confal-pyramids/util.ts
dsehnal: Well, you scan thru the entire tables in the categories. You could at least create an index on the tables so you don't have to effectively parse the data each time?

MadCatX: Since a step maps to two residues I could create a `ResidueIndex` to `Pyramid` mapping right away in `createPyramidfFromCif()`, get the residue indices from `Loci` and just fetch the pyramid from the map. Would this be an acceptable solution?

dsehnal: Yes, you can create the map when you create the pyramid representation and store it using the ``FormatPropertyProvider`` on the model. That way the pyramid label access would be instant.

MadCatX: If this idea turns out as workable (proper NtC steps mapping can get pretty awkward with weird structures) it should allow me to eliminate most of the clumsy lookup code from util. I will give it a try and see where it breaks :)

MadCatX: I considered the possibility of organising the NtC data in a way that would allow for more efficient lookups. Unfortunately, it is not that straightforward. I will try to explain what the problem is...

Handling of NtC steps is awkward because of alternate positions. While altpos is an atom propery, NtC steps need to "inflict" the fact that there are some atoms with altposes onto the entire residue. If I were to use `ResidueIndex` (actually two of them, one for each residue in the step) as a key in a map, such a key could possibly map to multiple steps and it would not be appararent which step to take without some extra logic.

When the current code gets two residues that may form a step, it has to:
- Check if there are any atoms with altpos on any of the residues
- Create a list of all possible step names, considering the altpos
- Check if any of the possible step names is present in the structure
- Return a pyramid for the found step

While this is not particularly nice it seems to be the only reliable way how to handle this. NtC pyramids are stored in an array and there is a "step name" -> "array index" map to allow for at least somewhat efficient access.

When the `PyramidGetter.get()` gets a `Loci`, it needs to:
- Check if the contained elements form two consecutive residues
- Walk through the elements of both residues to look for altposes
- Check if the contained elements actually form a step and return the corresponding pyramid if true

Given the explanation above, is there still something I should be looking to optimise?

dsehnal: Can you share a file that has the ``ndb_struct_ntc_*`` categories?

dsehnal: Also, should the ``lociToPyramid`` contain a check if the ``ConfalPyramidsProvider`` is assigned?

Also ``if (!Unit.isAtomic(loci.elements[0].unit)) return void 0;`` this check is cheaper than the one above it, so should move it there.

src/extensions/dnatco/common.ts
```diff
@@ -0,0 +1,111 @@
...
+        for (let j = 0; j < i; j++) {
+            if (id === step_id.value(j)) return { _NtC: assigned_NtC.value(j), _confal_score: confal_score.value(j) };
+        }
+        throw new Error('Inconsistent mmCIF data');
+    }
+
+    export function hasNdbStructNtcCategories(model: Model): boolean {
+        if (!MmcifFormat.is(model.sourceData)) return false;
+        const names = (model.sourceData).data.frame.categoryNames;
+        return names.includes('ndb_struct_ntc_step') && names.includes('ndb_struct_ntc_step_summary');
```
dsehnal, line 105: This should be just:

```ts
const categories = model.sourceData.data.frame.categories;
return !!categories['ndb_struct_ntc_step'] && !!categories['ndb_struct_ntc_step_summary'] 
```

src/extensions/dnatco/confal-pyramids/util.ts
MadCatX: There you go: Each of these structures contains at least a few NtC steps with altposes

https://blackbox.ibt.biocev.org/devel/RCSB/cif_dnatco_updated/3vok_v32C35A23.cif.gz
https://blackbox.ibt.biocev.org/devel/RCSB/cif_dnatco_updated/1dk1_v32C35A23.cif.gz
https://blackbox.ibt.biocev.org/devel/RCSB/cif_dnatco_updated/6r93_v32C35A23.cif.gz
https://blackbox.ibt.biocev.org/devel/RCSB/cif_dnatco_updated/3ok4_v32C35A23.cif.gz

src/extensions/dnatco/common.ts
MadCatX: Yes, there should be a check whether `ConfalPyramidsProvider` is attached so that the label is not displayed at all times.

arose:
Hi @MadCatX would you like us to keep this PR open or would you like to do a new one with your updated code?

MadCatX:
Things have evolved quite a lot since I opened this PR so it'd make a lot more sense to do this in a new PR. Thanks.

arose:
> Things have evolved quite a lot since I opened this PR so it'd make a lot more sense to do this in a new PR. Thanks.

Thanks, looking forward to a new PR.

PR #89, opened 2020-09-29
Add transform functionality to ModelServer

JonStargaryen:
- adds `transform` parameter to (single) ModelServer queries that will be applied to coordinates in `_atom_site`
- honors rest of query normally
- e.g. useful to select a binding site and superimpose it onto a 2nd binding site if alignment was determined externally

### DIFF
--- a/src/servers/model/server/api.ts
+++ b/src/servers/model/server/api.ts
@@ -46,7 +46,8 @@ export const CommonQueryParamsInfo: QueryParamInfo[] = [
     { name: 'model_nums', type: QueryParamType.String, description: `A comma-separated list of model ids (i.e. 1,2). If set, only include atoms with the corresponding '_atom_site.pdbx_PDB_model_num' field.` },
     { name: 'encoding', type: QueryParamType.String, defaultValue: 'cif', description: `Determines the output encoding (text based 'CIF' or binary 'BCIF'). Ligands can also be exported as 'SDF', 'MOL', or 'MOL2'.`, supportedValues: ['cif', 'bcif', 'sdf', 'mol', 'mol2'] },
     { name: 'copy_all_categories', type: QueryParamType.Boolean, defaultValue: false, description: 'If true, copy all categories from the input file.' },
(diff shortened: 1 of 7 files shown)


src/servers/model/server/api.ts
```diff
@@ -288,7 +291,8 @@ export function normalizeRestCommonParams(params: any): CommonQueryParamsInfo {
         model_nums: params.model_nums ? ('' + params.model_nums).split(',').map(n => n.trim()).filter(n => !!n).map(n => +n) : void 0,
         data_source: params.data_source,
         copy_all_categories: Boolean(params.copy_all_categories),
-        encoding: mapEncoding(('' + params.encoding).toLocaleLowerCase())
+        encoding: mapEncoding(('' + params.encoding).toLocaleLowerCase()),
+        transform: params.transform ? createMat4(('' + params.transform).split(',').map(n => n.trim()).map(n => +n)) : Mat4.identity()
```
arose, line 295: You can just cast `as Mat4`, `Mat4` is just a number array of length 16.

Why convert to a string and parse it again?

JonStargaryen: I followed the example of `model_nums` which will also create a number[]. `typeof params.transform` reports string at the start of the method. I understood the explicit conversion to string as safety net (though unsure against which scenario).

`as Mat4` will compile but fail at runtime because the type of matrix elements is still string.

dsehnal: He meant that instead of calling ``createMat4`` (which can be removed) you can just do ``as Mat4``. Parsing the numbers is still required.

dsehnal (commented):
Looks good overall.

Please increment the version in ``version.ts`` as well.

src/servers/model/server/api.ts
JonStargaryen: Oh, sure thing. Done in 6f05179.

JonStargaryen:
Bumped version and updated CHANGELOG in 73fa675. 
1f2812b ensures consistent type definition in `CommonQueryParamsInfo`.

PR #92, opened 2020-10-13
Add chromosome and region query for g3d format

lidaof:
add chromosome and region query for g3d format with examples below.

show struct from chr7:
```js
components.apply(StateTransforms.Model.StructureSelectionFromExpression, { expression: g3dChromosomeQuery('chr7'), label: 'chr7' })
        .apply(StateTransforms.Representation.StructureRepresentation3D, reprChrom);
```
show struct from chr7:2000000-7000000:
```js
    components.apply(StateTransforms.Model.StructureSelectionFromExpression, { expression: g3dRegionQuery('chr7', 2000000, 7000000), label: 'chr7:2000000-7000000' })
        .apply(StateTransforms.Representation.StructureRepresentation3D, reprRegion);
```

### DIFF
--- a/src/extensions/g3d/model.ts
+++ b/src/extensions/g3d/model.ts
@@ -153,7 +153,8 @@ async function getTraj(ctx: RuntimeContext, data: G3dDataBlock) {
         haplotypes: Object.keys(data.data),
         haplotype: normalized.haplotype,
         resolution: data.resolution,
-        start: normalized.start
+        start: normalized.start,
+        chroms: normalized.chromosome,
     });
 
     return models;
@@ -224,7 +225,8 @@ export interface G3dInfoData {
     haplotypes: string[],
     haplotype: string[],
     start: Int32Array,
-    resolution: number
+    resolution: number,
+    chroms: string[]
 };
 
 export const G3dLabelProvider: LociLabelProvider = {
--- a/src/extensions/g3d/model.ts
+++ b/src/extensions/g3d/model.ts
@@ -178,7 +178,7 @@ export const G3dSymbols = {
     chromosome: QuerySymbolRuntime.Dynamic(CustomPropSymbol('g3d', 'chromosome', Type.Str),
         ctx => {
             if (Unit.isAtomic(ctx.element.unit)) return '';
-            const {  asym_id } = ctx.element.unit.model.coarseHierarchy.spheres;
+            const { asym_id } = ctx.element.unit.model.coarseHierarchy.spheres;
             return asym_id.value(ctx.element.element) || '';
         }
     ),
@@ -202,15 +202,21 @@ export function g3dHaplotypeQuery(haplotype: string) {
     });
 }
 
-export function g3dChromosomeQuery(chromosome: string) {
+export function g3dChromosomeQuery(chr: string) {
     return MS.struct.generator.atomGroups({
-        'chain-test': MS.core.rel.eq([G3dSymbols.chromosome.symbol(), chromosome]),
+        'chain-test': MS.core.logic.and([
+            MS.core.rel.eq([MS.ammp('objectPrimitive'), 'sphere']),
+            MS.core.rel.eq([G3dSymbols.chromosome.symbol(), chr])
+        ])
     });
 }
 
 export function g3dRegionQuery(chr: string, start: number, end: number) {
     return MS.struct.generator.atomGroups({
-        'chain-test': MS.core.rel.eq([G3dSymbols.chromosome.symbol(), chr]),
+        'chain-test': MS.core.logic.and([
+            MS.core.rel.eq([MS.ammp('objectPrimitive'), 'sphere']),
+            MS.core.rel.eq([G3dSymbols.chromosome.symbol(), chr])
+        ]),
         'residue-test': MS.core.rel.inRange([G3dSymbols.region.symbol(), start, end])
     });
 }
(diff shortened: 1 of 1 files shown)


src/extensions/g3d/model.ts
```diff
@@ -174,6 +174,23 @@ export const G3dSymbols = {
             const seqId = ctx.element.unit.model.coarseHierarchy.spheres.seq_id_begin.value(ctx.element.element);
             return info.haplotype[seqId] || '';
         }
+    ),
+    chromosome: QuerySymbolRuntime.Dynamic(CustomPropSymbol('g3d', 'chromosome', Type.Str),
```
arose, line 178: @dsehnal, looks like we don't expose the coarse properties, probably should

src/extensions/g3d/model.ts
```diff
@@ -174,6 +174,23 @@ export const G3dSymbols = {
...
+        ctx => {
+            if (Unit.isAtomic(ctx.element.unit)) return '';
+            const {  asym_id } = ctx.element.unit.model.coarseHierarchy.spheres;
+            return asym_id.value(ctx.element.element) || '';
+        }
+    ),
+    region: QuerySymbolRuntime.Dynamic(CustomPropSymbol('g3d', 'region', Type.Str),
+        ctx => {
+            if (Unit.isAtomic(ctx.element.unit)) return '';
```
arose, line 187: Should this function always return a number?

src/extensions/g3d/model.ts
```diff
@@ -185,6 +202,19 @@ export function g3dHaplotypeQuery(haplotype: string) {
...
 
+export function g3dChromosomeQuery(chromosome: string) {
+    return MS.struct.generator.atomGroups({
+        'chain-test': MS.core.rel.eq([G3dSymbols.chromosome.symbol(), chromosome]),
+    });
+}
+
+export function g3dRegionQuery(chr: string, start: number, end: number) {
+    return MS.struct.generator.atomGroups({
+        'chain-test': MS.core.rel.eq([G3dSymbols.chromosome.symbol(), chr]),
```
arose, line 213: You can filter for sphere units in the chain test so the residue test would not be called needlessly (same for g3dChromosomeQuery to catch the edge case of an empty chain/chromosome name better). 

```ts
'chain-test': MS.core.logic.and([
    MS.core.rel.eq([MS.ammp('objectPrimitive'), 'sphere']),
    MS.core.rel.eq([G3dSymbols.chromosome.symbol(), chr])
])
```

src/extensions/g3d/model.ts
dsehnal: Yeah, we should. Not in scope of this PR tho.

src/extensions/g3d/model.ts
lidaof: Thank you for comments @arose.  can you please give an example or modify the PR? I am not familiar with the mol* codebase, this was got from @dsehnal's example code before. good it's working :) Thank both of you!

arose: Well, just swap out the two 'chain-test's with the one I gave above and try to see if it still does what you want

lidaof: I see, thanks @arose, yes, it still works as what I want!

src/extensions/g3d/model.ts
```diff
@@ -174,6 +174,23 @@ export const G3dSymbols = {
...
         }
+    ),
+    chromosome: QuerySymbolRuntime.Dynamic(CustomPropSymbol('g3d', 'chromosome', Type.Str),
+        ctx => {
+            if (Unit.isAtomic(ctx.element.unit)) return '';
+            const { asym_id } = ctx.element.unit.model.coarseHierarchy.spheres;
+            return asym_id.value(ctx.element.element) || '';
+        }
+    ),
+    region: QuerySymbolRuntime.Dynamic(CustomPropSymbol('g3d', 'region', Type.Str),
```
arose, line 185: I think this should be `Type.Num` not `Type.Str` to match the type of info.start[seqId]

src/extensions/g3d/model.ts
arose: this should return a number or throw

src/extensions/g3d/model.ts
```diff
@@ -174,6 +174,23 @@ export const G3dSymbols = {
...
+        ctx => {
+            if (Unit.isAtomic(ctx.element.unit)) return '';
+            const info = (G3dInfoDataProperty as any).get(ctx.element.unit.model);
+            if (!info) return '';
```
arose, line 189: should return a number or throw

src/extensions/g3d/model.ts
```diff
@@ -174,6 +174,23 @@ export const G3dSymbols = {
...
+            const seqId = ctx.element.unit.model.coarseHierarchy.spheres.seq_id_begin.value(ctx.element.element);
+            const s = info.start[seqId];
```
arose, line 191: maybe just `return info.start[seqId] || 0;`

lidaof: thank you @arose for these nice comments!

lidaof:
trying added chromosome array to the G3dInfoData object, as I need the chr and start and resolution for data query to bigwig and bigbed files for decoration. please review, thank you so much!

lidaof:
Hi Alex and David,

do you think this PR could be merged? Thank you so much!

lidaof:
Thank you!!

PR #96, opened 2020-11-12
handle metal ions correctly for ModelServer ligand export

JonStargaryen:
- `bondMap` is undefined if metal ions are requested
- this PR ensures that those cases are handled gracefully and SDF/MOL/MOL2 files without bond information are returned


arose:
lgtm

PR #98, opened 2020-11-16
Wboit

AronKovacs:


src/apps/viewer/index.ts
```diff
@@ -117,6 +118,7 @@ export class Viewer {
             config: [
                 [PluginConfig.General.DisableAntialiasing, o.disableAntialiasing],
                 [PluginConfig.General.PixelScale, o.pixelScale],
+                [PluginConfig.General.EnableWboit, o.enableWboit],
```
dsehnal, line 121: Could the feature be enabled/disabled during runtime? And if so, do you think it would be better to have this is a canvas/renderer prop?

dsehnal:
Is it (more) ready for merge now @AronKovacs ?

Also, thanks a lot for the contribution!

AronKovacs:
Alexander said yes.

PR #116, opened 2021-01-07
ModelServer: Add option to download text files

JonStargaryen:
- currently, text files will be shown in browser, this PR allows to download files directly by adding `download=true` to the URL
- fixes a small bug in the normalization of rest params as `Boolean('false')` returns `true` for some wild reason (happens for the `copy_all_categories` flag)
- also, there is an issue with the swagger-ui: executing the example at https://models.rcsb.org/#/General/assembly gets the response from ModelServer but swagger will freeze for several minutes - seems to be related to the media type of the response and doesn't happen for BCIF


src/servers/model/server/api.ts
```diff
@@ -290,9 +292,11 @@ export function normalizeRestCommonParams(params: any): CommonQueryParamsInfo {
     return {
         model_nums: params.model_nums ? ('' + params.model_nums).split(',').map(n => n.trim()).filter(n => !!n).map(n => +n) : void 0,
         data_source: params.data_source,
-        copy_all_categories: Boolean(params.copy_all_categories),
+        // Boolean('false') return true
+        copy_all_categories: params.copy_all_categories === 'true',
```
dsehnal, line 296: This is a breaking change! 

Some people use it as ``copy_all_categories=1``. 

The ``Boolean('false') => true``  because non-empty string is truthy in JS.

The use case if you don't want the param to be true is not to include it or use empty string as the value.

But I agree this is not very intuitive. Maybe add a function 

```ts
function isTrue(v: any) {
  const b = Boolean(v);
  if (!b) return false;
  if (typeof b === 'string') return b !== '0' && b.toLowerCase() !== 'false';
  return true;
}
```

and use that? (please test it before you push it).

src/servers/model/server/api.ts
```diff
@@ -290,9 +292,11 @@ export function normalizeRestCommonParams(params: any): CommonQueryParamsInfo {
...
         encoding: mapEncoding(('' + params.encoding).toLocaleLowerCase()),
-        transform: params.transform ? ('' + params.transform).split(',').map(n => n.trim()).map(n => +n) as Mat4 : Mat4.identity()
+        transform: params.transform ? ('' + params.transform).split(',').map(n => n.trim()).map(n => +n) as Mat4 : Mat4.identity(),
+        download: params.download === 'true'
```
dsehnal, line 299: Same thing.

src/servers/model/server/api-web.ts
```diff
@@ -45,11 +45,11 @@ async function processNextJob() {
     }
 }
 
-export function createResultWriter(response: express.Response, encoding: string, entryId?: string, queryName?: string) {
+export function createResultWriter(response: express.Response, encoding: string, download: boolean, entryId?: string, queryName?: string) {
```
dsehnal, line 48: If we add any more params here perhaps it should be params object instead. Would you might doing that refactoring now?

dsehnal (changes requested):
Thanks for the PR. A few issues that need to be addressed before merging tho.

src/servers/model/server/api.ts
JonStargaryen: Again done in d60c3dd.

src/servers/model/server/api-web.ts
JonStargaryen: Done in ea9e25b. Let me know if you'd like to see `ResultWriterParams` moved somewhere else.

dsehnal:
Look good, thanks. Do you need new NPM now?

JonStargaryen:
> Look good, thanks. Do you need new NPM now?

Thanks. No rush in terms of a NPM release, whenever it's convenient for you.

PR #119, opened 2021-01-13
ParamDefinition.normalizePrams

dsehnal:
Adds the ability to "normalize" parameter values based on the definition.

### DIFF
--- a/src/mol-util/param-definition.ts
+++ b/src/mol-util/param-definition.ts
@@ -550,7 +550,7 @@ export namespace ParamDefinition {
         }
     }
 
-    function normalizeParam(p: Any, value: any, defaultIfUndefined: boolean): any {
+    function normalizeParam(p: Any, value: any, defaultIfUndefined: boolean, prune: boolean): any {
         if (value === void 0 || value === null) {
             return defaultIfUndefined ? p.defaultValue : void 0;
         }
@@ -563,13 +563,13 @@ export namespace ParamDefinition {
         if (p.type === 'value') {
             return value;
         } else if (p.type === 'group') {
-            const ret = { ...value };
+            const ret = prune ? Object.create(null) : { ...value };
             for (const key of Object.keys(p.params)) {
                 const param = p.params[key];
-                if (ret[key] === void 0) {
+                if (value[key] === void 0) {
                     if (defaultIfUndefined) ret[key] = param.defaultValue;
                 } else {
-                    ret[key] = normalizeParam(param, ret[key], defaultIfUndefined);
+                    ret[key] = normalizeParam(param, value[key], defaultIfUndefined, prune);
                 }
             }
             return ret;
@@ -592,7 +592,7 @@ export namespace ParamDefinition {
             const param = p.map(v.name);
             return {
                 name: v.name,
-                params: normalizeParam(param, v.params, defaultIfUndefined)
+                params: normalizeParam(param, v.params, defaultIfUndefined, prune)
             };
         } else if (p.type === 'select') {
             const options = !!p._optionSet?.has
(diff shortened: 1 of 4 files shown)


src/mol-util/param-definition.ts
```diff
@@ -548,6 +550,91 @@ export namespace ParamDefinition {
         }
     }
 
+    function normalizeParam(p: Any, value: any, defaultIfUndefined: boolean, prune: boolean): any {
```
arose, line 553: with prune = true, the group params return object will only contain current properties, without it, all old properties and there values are copied over

what is the use-case for prune = false? why keep the old properties around? for testing?

src/mol-util/param-definition.ts
```diff
@@ -548,6 +550,91 @@ export namespace ParamDefinition {
...
+            return value.map(v => normalizeParams(p.element, v, defaultIfUndefined ? 'all' : 'skip', prune));
+        }
+
+        // TODO: validate/normalize all param types "properly"??
+
+        return value;
+    }
+
+    export function normalizeParams(p: Params, value: any, defaultIfUndefined: 'all' | 'children' | 'skip', prune: boolean) {
+        if (typeof value !== 'object') {
```
arose, line 622: add `|| value === null` since typeof null is 'object'?

src/mol-util/param-definition.ts
```diff
@@ -548,6 +550,91 @@ export namespace ParamDefinition {
...
+            if (!Array.isArray(value)) return p.defaultValue;
+            const ret = value.filter(function (this: Set<any>, v: any) { return this.has(v); }, options);
+            if (value.length > 0 && ret.length === 0) return p.defaultValue;
+            return ret;
+        } else if (p.type === 'object-list') {
+            if (!Array.isArray(value)) return p.defaultValue;
+            return value.map(v => normalizeParams(p.element, v, defaultIfUndefined ? 'all' : 'skip', prune));
+        }
+
+        // TODO: validate/normalize all param types "properly"??
```
arose, line 616: definitely - in time :)

src/mol-util/param-definition.ts
```diff
@@ -77,7 +77,8 @@ export namespace ParamDefinition {
         type: 'select'
         /** array of (value, label) tuples */
         options: readonly (readonly [T, string] | readonly [T, string, string | undefined])[]
-        cycle?: boolean
+        cycle?: boolean,
+        _optionSet?: Set<T>
```
arose, line 81: Is this worthwhile? how many options need there be for this to be needed? how often normalized?

It should be excluded in PD.clone

arose (commented):
Super useful, thanks!

src/mol-util/param-definition.ts
dsehnal: Because of things like this https://github.com/molstar/molstar/blob/master/src/mol-theme/color/operator-name.ts#L46 (not pointing finders, but ... :D ) 

I wonder if this part deserves refactoring and how would we go around with that.

src/mol-util/param-definition.ts
dsehnal: You're right, probably not worth it at this point.

src/mol-util/param-definition.ts
```diff
@@ -548,6 +550,91 @@ export namespace ParamDefinition {
         }
     }
 
+    function normalizeParam(p: Any, value: any, defaultIfUndefined: boolean, prune: boolean): any {
+        if (value === void 0 || value === null) {
+            return defaultIfUndefined ? p.defaultValue : void 0;
+        }
+
+        // TODO: is this a good idea and will work well?
```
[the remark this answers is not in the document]
dsehnal: Will leave this commented out for now, still not 100% convinced about edge cases.

src/mol-util/param-definition.ts
arose: > Because of things like this https://github.com/molstar/molstar/blob/master/src/mol-theme/color/operator-name.ts#L46

I think we can move those into the getParams function.

>  (not pointing finders, but ... :D )

guilty as charged

arose: moved the label options out, see 8d6557e51cf8367fc01d30ebf7ebe8bb5cc9ac2f

src/mol-util/param-definition.ts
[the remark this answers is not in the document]
arose: ok, anything specific that leaves you unconvinced?

PR #135, opened 2021-03-04
Move part of PluginContext to mol-plugin-ui

dsehnal:
Moves all React dependencies from ``mol-plugin`` to ``mol-plugin-ui`` by introducing ``PluginUIContext``.

### DIFF
--- a/src/mol-plugin/context.ts
+++ b/src/mol-plugin/context.ts
@@ -8,44 +8,51 @@
 import produce, { setAutoFreeze } from 'immer';
 import { List } from 'immutable';
 import { merge, Subscription } from 'rxjs';
+import { filter, take } from 'rxjs/operators';
 import { Canvas3D, Canvas3DContext, DefaultCanvas3DParams } from '../mol-canvas3d/canvas3d';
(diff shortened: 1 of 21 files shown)


arose:
Great, I will try it tomorrow.

src/mol-plugin/context.ts
```diff
@@ -176,9 +174,8 @@ export class PluginContext {
 
     readonly customModelProperties = new CustomProperty.Registry<Model>();
     readonly customStructureProperties = new CustomProperty.Registry<Structure>();
-    readonly customParamEditors = new Map<string, StateTransformParameters.Class>();
 
-    readonly customStructureControls = new Map<string, { new(): PluginUIComponent<any, any, any> }>();
+    readonly customStructureControls = new Map<string, { new(): any /* contructible react components */ }>();
```
arose, line 178: typo: constructible

arose (commented):
This works well, no issues with including react or UI related code. Thanks!. The only aspects that's annoying is that you now have register Actions manually if you don't use the UI version.

dsehnal:
Ok I will move the actions back to the original spec, there is no UI there.

PR #141, opened 2021-03-14
Surrounding Ligands query

dsehnal:
Adds a new query to the Model Server.


src/mol-model/structure/model/properties/utils/residue-set.ts
```diff
@@ -0,0 +1,107 @@
...
+import { StructureElement } from '../../../structure/element';
+import { StructureProperties } from '../../../structure/properties';
+
+export interface ResidueSetEntry {
+    label_asym_id: string,
+    label_comp_id: string,
+    label_seq_id: number,
+    label_alt_id: string,
+    ins_code: string,
+    // 1_555 by default
```
arose, line 16: I usually use /** */ comments here so it shows up in editor tool tips.

arose:
Trying to understand what surroundingLigands does exactly. Does it add ligands if they are within x Angstrom and (are prd or have struct conn)? So it it ignores ligands that are not bonded, right?

What is the use-case?

src/mol-model/structure/model/properties/utils/residue-set.ts
```diff
@@ -0,0 +1,107 @@
...
+    private checkOperator: boolean = false;
+
+    add(entry: ResidueSetEntry) {
+        let root = this.index.get(entry.label_asym_id);
+        if (!root) {
+            root = new Map();
+            this.index.set(entry.label_asym_id, root);
+        }
+
+        let entries = root.get(entry.label_seq_id);
```
arose, line 31: Can this be extended to be with residues that don't have a label_seq_id (water, non-polymer, branched)?

dsehnal: This should work for these to. They will have unique ``label_asym_id`` and ``label_seq_id`` will evaluate to 0. And the index contains ``ResidueSetEntry[]`` for each ``(asym_id, seq_id)`` pair, so I think it should work fine. Or am I missing something?

arose: I think it will break for branched entities, e.g., https://files.rcsb.org/view/1B5F.cif, multiple residues, single asym_id, no label_seq_id.

dsehnal: image
image

works fine.

dsehnal: To compare, this is 5A residues surroundings without the struct_conn expand:

image

arose: Yeah, it works for this use case because you want the whole branched entity anyway and the residues in branched entities are connected with with struct con records as well.

I thing if you use ResiudueSet for another use case where you just want to add a single residue of a branched entity it will add all of them.

Issue #147, opened 2021-03-17
CSS transform:scale cause molstar canvas to have incorrect size


TomasKulhanek:
We use the pdbe-molstar web component using molstar as plugin in our interactive documentation. 
In some cases we need the whole part of the web page to be smaller, therefore, we use CSS `transform: scale(0.5);transform-origin:top left;width:200%`  This cause the molstar canvas to be broken. The viewport (msp-viewport) have correct size. 
But, the canvas with molecule in white background has incorrect size (even smaller) as seen in the following screenshot:

![firefox_pTJ9hFF77T](https://user-images.githubusercontent.com/1429487/110948864-295d6980-8342-11eb-8312-b6ff7f209562.png)

Prepared plunker example of this error at https://plnkr.co/edit/pdmHZTULh3vUPq8e?preview

The canvas size should fit the div. This issue is present in chrome 88.0.4324.190, firefox 86.0, ms edge 88.0.705.74

PR #148, opened 2021-03-17
FIX issue #147 CSS transform:scale cause molstar canvas to have incorrect size

TomasKulhanek:
FIX issue #147 CSS transform:scale cause molstar canvas to have incorrect size
When CSS transform is used e.g.: scale(0.5);transform-origin:top left;width:200% molstar canvas is broken. 

The introduced fix uses HTMLElement offsetWidth/offsetHeight properties containing correct size of element when css transform:scale is used.

### DIFF
--- a/src/mol-canvas3d/util.ts
+++ b/src/mol-canvas3d/util.ts
@@ -12,13 +12,13 @@ export function setCanvasSize(canvas: HTMLCanvasElement, width: number, height: 
 }
 
 /** Resize canvas to container element taking `devicePixelRatio` into account */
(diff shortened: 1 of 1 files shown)


dsehnal:
Thanks for the fix, tested it and seems to work well.

hlvlad:
HI, are there plans to backport this fix to 1.3.X version (and so npm package)? We have the same problem in our app.

arose:
> 
> 
> HI, are there plans to backport this fix to 1.3.X version (and so npm package)? We have the same problem in our app.

released 1.3.2 with this and a few other fixes (https://github.com/molstar/molstar/commits/v1.3.1)

PR #150, opened 2021-03-21
[wip] basic support for bond stubs

arose:
- line and ball & stick repr
- stubs support in link visual helper
- getData and mustRecreate methods for structure repr provider
- Structure.WithChild helper (needs Proxy support)

### DIFF
--- a/src/mol-plugin-state/transforms/representation.ts
+++ b/src/mol-plugin-state/transforms/representation.ts
@@ -126,14 +126,15 @@ const StructureRepresentation3D = PluginStateTransform.BuiltIn({
         return Task.create('Structure Representation', async ctx => {
             const propertyCtx = { runtime: ctx, assetManager: plugin.managers.asset };
             const provider = plugin.representation.structure.registry.get(params.type.name);
-            if (provider.ensureCustomProperties) await provider.ensureCustomProperties.attach(propertyCtx, a.data);
+            const data = provider.getData?.(a.data, params.type.params) || a.data;
+            if (provider.ensureCustomProperties) await provider.ensureCustomProperties.attach(propertyCtx, data);
             const repr = provider.factory({ webgl: plugin.canvas3d?.webgl, ...plugin.representation.structure.themes }, provider.getParams);
-            await Theme.ensureDependencies(propertyCtx, plugin.representation.structure.themes, { structure: a.data }, params);
-            repr.setTheme(Theme.create(plugin.representation.structure.themes, { structure: a.data }, params));
+            await Theme.ensureDependencies(propertyCtx, plugin.representation.structure.themes, { structure: data }, params);
+            repr.setTheme(Theme.create(plugin.representation.structure.themes, { structure: data }, params));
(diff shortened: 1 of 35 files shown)


src/mol-plugin-state/transforms/representation.ts
```diff
@@ -126,41 +126,44 @@ const StructureRepresentation3D = PluginStateTransform.BuiltIn({
...
+            await repr.createOrUpdate(props, data).runInContext(ctx);
+            return new SO.Molecule.Structure.Representation3D({ repr, sourceData: a.data }, { label: provider.label });
         });
     },
     update({ a, b, oldParams, newParams, cache }, plugin: PluginContext) {
         return Task.create('Structure Representation', async ctx => {
             if (newParams.type.name !== oldParams.type.name) return StateTransformer.UpdateResult.Recreate;
 
             const provider = plugin.representation.structure.registry.get(newParams.type.name);
+            if (provider.mustRecreate?.(oldParams.type.params, newParams.type.params)) return StateTransformer.UpdateResult.Recreate;
```
dsehnal, line 145: Is default of ``mustRecreate`` true or false?

If it's true, then should do something like this https://github.com/molstar/molstar/commit/324820890a9cb66f30ba0420ad25e15133688697

dsehnal (commented):
The Parent Proxy is a really nice idea. How about implementing it with something along these lines?

```ts
class Structure {
    // ....

    // replace new Struture with Structure.fromUnits
    static fromUnits(units: ArrayLike<Unit>, props: Structure.Props = {}) {
        // ...
        return new Structure(units, unitMap, unitIndexMap, props);
    }

    private constructor(units: ReadonlyArray<Unit>, unitMap: Structure['unitMap'], unitIndexMap: Structure['unitIndexMap'], props: Structure['_props']) {
        // ...
    }

    private _child: Structure | undefined;
    private _target: Structure | undefined;
    asParent() {
        if (!this.parent) return this;

        const ret = new Structure(....);
        // Optionally, make these two props part of the new constructor so that we help the JS runtime
        // to not create new internal class for Structure when assigning the props
        ret._child = this;
        ret._target = this.parent;
        return ret;
    }

    get child(): Structure | undefined {
        return this._child;
    }

    // instead of Structure.WithChild.getTarget
    get target(): Structure {
        return this._target ?? this;
    }
    // ...
}
```

Refactoring wise, this would be a simple change with string replace ``new Structure(``  with ``Structure.fromUnits(``.

I think this approach would fit into the API more naturally. E.g.

```ts
const s = Structure.WithChild.fromStructure(s);
...
const t = Structure.WithChild.getTarget();
```

vs

```ts
const s = s.asParent();
...
const t = s.target;
```

arose:
> The Parent Proxy is a really nice idea. How about implementing it with something along these lines?
> [...]
> I think this approach would fit into the API more naturally. 

Good point. I integrated the functionality directly into the Structure class. We already had a factory method Structure.create, so not much to change.

Issue #28, opened 2020-04-05
Chances for Three.js interop/wrapper?


jussty:
What a great project!

Since this is a lot more low-level than NGL, how are the chances that there will be an official interface to Three.js at some point? It would make sense to include molecule geometry into existing tools, given the huge ecosystem. E.g. implementing advanced controls for XR with raycasting, colliders etc. will be hard (and unnecessary) to do from scratch for this project. 

Mol* seems to move fast paced at the moment, so maintaining a wrapper may currently be a liability. Could you give a an estimate when in the future will be a good point that such a thing may be implemented?

(I am coming over from a [related issue with NGL](https://github.com/arose/ngl/issues/728#issuecomment-609423107))

dsehnal:
I don't think we will ever have an official interface for Three.js.

However, we have a sub-project planned for generating library-independent geometries. These could then be imported in Three.js and other solutions.

We have a theming refactor planned for the ``mol-repr`` module soon and after that I think the code will be stable enough to write the exporter. Maybe you want to help with that?

awiddess-ccdc:
@dsehnal is it with spending a little time on the sub-project to support the library-independent geometries then allow this project to utilise the api?

arose:
> is it worth spending a little time on the sub-project to support the library-independent geometries then allow this project to utilise the api?

the current low-level work we are doing is to create an exporter interface for RenderObjects (create as part of representations and containing all geometry and coloring data) to transform them into something useful elsewhere. The first exporter will be a OBJ exporter but once we have a stable exporter interface exporters for THREE.js geometries or other formats can be added.

We want to use it for the Viewer app to offer download of OBJ files for the whole scene. But one can also load structures, create representations using low-level mol* functions and then pipe the representations' renderobjects through an exporter programmatically.

jussty:
> the current low-level work we are doing is to create an exporter interface for RenderObjects (create as part of representations and containing all geometry and coloring data) to transform them into something useful elsewhere. 

That sounds great, should be easy to derive Three.js geometries off this data. This is a very good approach, since it allows to ship geometry loaders without the need to carry whole Web Applications around as "library".

> The first exporter will be a OBJ exporter but once we have a stable exporter interface exporters for THREE.js geometries or other formats can be added.
[...]

PR #159, opened 2021-04-09
OBJ Export

sukolsak:
This PR adds the initial version of OBJ export through a new extension named geo-export. (#28) The code is structured like the mp4-export extension and 97e4edc1357e2f229c647d9b225553d71c9ee860, as advised by @arose over emails.

You can try the demo at https://stanford.edu/~sukolsak/molstar/. Click on “Export Geometries” at the right side of the screen. Lines and points are not supported yet.

If this sounds like a good idea, I plan to support other formats in the future, including
- PLY: there is an importer for it in Mol*.
- GLTF/USDZ: for mobile augmented reality on Android/Apple devices. I created a demo that works with PDB at https://stanford.edu/~sukolsak/ar/


src/extensions/geo-export/export.ts
```diff
@@ -0,0 +1,319 @@
...
+            const mesh = MeshBuilder.getMesh(state);
+            const vertices = mesh.vertexBuffer.ref.value;
+            const indices = mesh.indexBuffer.ref.value;
+            const normals = mesh.normalBuffer.ref.value;
+            const groups = mesh.groupBuffer.ref.value;
+            this.addMeshWithColors(vertices, indices, normals, groups, vertices.length / 3, indices.length, values, instanceIndex);
+        }
+    }
+
+    add(renderObject: GraphicsRenderObject) {
```
dsehnal, line 287: Do you think it would be worthwhile to add ``RuntimeContext`` reference here to get partial progress/cancellation for large meshes?

```ts
async add(renderObject: GraphicsRenderObject, ctx: RuntimeContext) {
  ...
  // pass the ctx to addMesh and other functions and do this every N elements (for large enough N)
  if (i % N === 0 && ctx.shouldUpdate) {
    await ctx.update(...);
  }
  ...
}
```

dsehnal (commented):
Hi @sukolsak, this is really nice, thank you.

I think there will be a considerable interest in this feature (we've received multiple requests to be able to export models for 3d printing as well) and using this extension for a "geometry server" service.

src/extensions/geo-export/export.ts
sukolsak: Hi David. Thank you for the feedback. Yes, it's definitely worthwhile. I have made these functions asynchronous. Users can now cancel the exporting for large meshes.

I'm not sure what's the best way to show partial progress though. I can do something like
```js
for (let i = 0; i < drawCount; i += 3) {
    if (i % 3000 === 0 && ctx.shouldUpdate) await ctx.update({ message: 'Rendering...', current: i, max: drawCount });
    ...
}
```
But then the number is reset for every RenderObject. Another approach that I can think of is to compute beforehand the total number of faces of all RenderObjects. But it seems complicated.

dsehnal: I can see two more options:
- Show show the current render object in the message. E.g. ``update({ message: `Exporting object ${i}/${objCount}`, current: ..., max: ... })``
- Return a ``Task`` by the ``add`` function and run it with ``runAsChild``. Since there won't event be that many objects to render, this should not have any negative perf. implications.
```ts
add(renderObject: GraphicsRenderObject, ctx: RuntimeContext) {
    return Task.create('Export Object', ctx => {
        if (!renderObject.state.visible) return;

        switch (renderObject.type) {
            case 'mesh':
                // can return the promise without using async function here instead of
                // async ctx => .... await add(..., ctx); break;
                return this.addMesh(renderObject.values as MeshValues, ctx);
            // ...
        }
    });
}

//////////////////////////

for (let i = 0, il = renderObjects.length; i < il; ++i) {
    await objExporter.add(renderObjects[i]).runAsChild(ctx);
    await ctx.update({ current: i });
}
```

sukolsak: Thank you for your suggestions. I've updated the code to show partial progress.

dsehnal:
Would  you like it merged now and continue in a separate PR or do more changes here 1st?

sukolsak:
Having it merged now would be great. Thank you!

PR #163, opened 2021-04-14
Fix incorrect deflate progress

sukolsak:
The `deflate` function sometimes reports partial progress where the current value is larger than the max value, such as "Deflating... [417630611/252028364]". This usually happens when exporting a large mesh such as 1RB8. This PR fixes the issue by using the correct current value.


dsehnal:
Thanks.

PR #173, opened 2021-05-03
Color smoothing

arose:
- [X] test on webgl1 and without extra extensions
- [ ] support grid/gridInstance color-type in geometry export (fyi @sukolsak), see #184

This adds volumetric color-smoothing for Mesh and TextureMesh geometries. It works by creating a 3d grid, accumulating all vertex colors weighted by their distance for each nieghbouring grid cell, then normalize each grid cell by the number of accumulated colors. Smooth vertex colors are then sampled from the 3d grid trilinearly either on the GPU or CPU depending on 3d grid size.

A `smoothColors` option is added to MolecularSurface and GaussianSurface representations.

Color themes can opt-in with a new property `preferSmoothing: true` to automatic color smoothing.

To keep performance acceptable for high resolution surfaces 1) the color volume is by default half the surfaces resolution and 2) only every 6th vertex is accumulated.

image
image


src/mol-gl/webgl/texture.ts
```diff
@@ -407,8 +407,8 @@ export function loadImageTexture(src: string, cell: ValueCell<Texture>, texture:
 
 //
 
-export function createNullTexture(gl: GLRenderingContext, kind: TextureKind): Texture {
-    const target = getTarget(gl, kind);
+export function createNullTexture(gl?: GLRenderingContext): Texture {
+    const target = 3553;
```
dsehnal, line 411: Are you sure ``gl.TEXTURE_2D`` will be the same across all platforms? Maybe replace this with ``gl?.TEXTURE_2D ?? 3553`` ?

src/mol-repr/structure/complex-visual.ts
```diff
@@ -198,17 +200,30 @@ export function ComplexVisual<G extends Geometry, P extends StructureParams & Ge
...
+
     return {
         get groupCount() { return locationIt ? locationIt.count : 0; },
         get renderObject () { return locationIt && locationIt.count ? renderObject : undefined; },
         createOrUpdate(ctx: VisualContext, theme: Theme, props: Partial<PD.Values<P>> = {}, structure?: Structure) {
             prepareUpdate(theme, props, structure || currentStructure);
             if (updateState.createGeometry) {
                 const newGeometry = createGeometry(ctx, newStructure, newTheme, newProps, geometry);
-                return newGeometry instanceof Promise ? newGeometry.then(update) : update(newGeometry);
+                if (newGeometry instanceof Promise) {
```
dsehnal, line 216: I know this was here before, but in general it might be safer to use a helper function akin to

```ts
function isPromiseLike<T = any>(x: any): x is Promise<T> {
  return typeof x?.then === 'function';
}
```

in the case ``Promise`` comes from an unknown library. There seem to be 3 instance of this ``instanceof Promise`` in ``mol-repr``.

src/mol-repr/structure/visual/util/color.ts
```diff
@@ -0,0 +1,106 @@
...
+
+export function getColorSmoothingProps(props: PD.Values<ColorSmoothingParams>, theme: Theme, resolution?: number, webgl?: WebGLContext) {
+    if ((props.smoothColors.name === 'on' || (props.smoothColors.name === 'auto' && theme.color.preferSmoothing)) && resolution && resolution < 3 && webgl) {
+        let stride = 6;
+        if (props.smoothColors.name === 'on') {
+            resolution *= props.smoothColors.params.resolutionFactor;
+            stride = props.smoothColors.params.sampleStride;
+        } else {
+            if (resolution > 0.5 && resolution < 1) {
+                resolution *= 2 * resolution;
```
dsehnal, line 37: Why ``v <- 2 * v * v``  for values between 0.5 and 1? This seems quite random and there is a weird discontinuity with the other options.

E.g. if resolution = 0.5, the final value will be 1, but if resolution = 0.500001, it will end up being ~0.5 (0.5 * 0.5 * 2).

src/mol-theme/color.ts
```diff
@@ -45,9 +45,12 @@ interface ColorTheme<P extends PD.Params> {
     readonly granularity: ColorType
     readonly color: LocationColor
     readonly props: Readonly<PD.Values<P>>
-    // if palette is defined, 24bit RGB color value normalized to interval [0, 1]
-    // is used as index to the colors
+    /**
+     * if palette is defined, 24bit RGB color value normalized to interval [0, 1]
+     * is used as index to the colors
+     */
```
dsehnal, line 51: I feel personally attacked by this change :D

dsehnal (approved):
Visuals look really nice, couple of comments on the code.

src/mol-gl/webgl/texture.ts
arose: They are guaranteed to be the same value but this way it is less of a magic number :)

src/mol-repr/structure/complex-visual.ts
arose: Good point, added the helper

src/mol-repr/structure/visual/util/color.ts
arose: Yeah, that needed some more work. Changed it to https://graphtoy.com/?f1(x,t)=(2-smoothstep(0,1.1,x))*x&coords=0.7,0.6,1.8

src/mol-theme/color.ts
arose: Can't help myself when I see comments :)

PR #177, opened 2021-05-06
Add per-group transparency support to geometry exporters

### LINKED ISSUE #176: Add per-group transparency support to geometry exporters
Add per-group transparency support to geometry exporters if applicable for target format.

The transparency data is stored in a texture similar to color data and accessed by group-id

```ts
uTransparencyTexDim: UniformSpec('v2'),
tTransparency: TextureSpec('image-uint8', 'alpha', 'ubyte', 'nearest'),
dTransparency: DefineSpec('boolean'),
transparencyAverage: ValueSpec('number'), // can be used to check if there is an transparency at all
```
The transparency should be combined with the per-object alpha value.

sukolsak:
This should resolve #176.

Am I supposed to use `transparencyAverage` somewhere?

I don't think it's a big enough change to warrant an update in the changelog. But let me know.

### DIFF
--- a/src/extensions/geo-export/glb-exporter.ts
+++ b/src/extensions/geo-export/glb-exporter.ts
@@ -36,16 +36,14 @@ export class GlbExporter extends MeshExporter<GlbData> {
     private binaryBuffer: ArrayBuffer[] = [];
     private byteOffset = 0;
 
-    private static vec3MinMax(a: NumberArray, stride: number) {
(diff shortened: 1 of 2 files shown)


dsehnal:
Looks good, thanks for the PR!

> Am I supposed to use `transparencyAverage` somewhere?

I don't think so, but let's wait for @arose to confirm.

> I don't think it's a big enough change to warrant an update in the changelog. But let me know.

No need I guess.

arose:
Thank you @sukolsak, looks good as it is.

You are right, no need for `transparencyAverage` here.

Issue #175, opened 2021-05-04
Investigate non-normalized normals


arose:
Investigate non-normalized normals observer by @sukolsak in #174.

arose:
Hey @sukolsak, not sure if I remember correctly: was there another source of non-normalized normals?

sukolsak:
@arose  There are two sources that I have found so far:
- Normals in arrowheads. See #179
- The `add*Cylinder` methods. These methods call `MeshBuilder.addPrimitive(state, tmpCylinderMat, cylinder)`, which does this
```js
const n = mat3directionTransform(tmpMat3, t);
...
// normal
v3transformMat3(tmpV, v3fromArray(tmpV, na, i), n);
```
`tmpV` is not normalized because of the non-uniform scaling of cylinders.

PR #179, opened 2021-05-08
Fix normals in sheet arrows

sukolsak:
When drawing an arrowhead, the current code applies `normalOffset` to the torsion vectors instead of the normal vectors. This results in incorrect lighting as shown below. Notice the sudden change in the lighting of the arrowheads.

1TQN

This PR applies `normalOffset` to the normal vectors.
1TQN-2

This PR also sets `normalOffset` to zero when drawing parts that are not an arrowhead.

Note that `normalOffset` causes some normal vectors to not be normalized (#175). This PR doesn't fix that. I don't know yet how to fix it without using `Vec3.normalize()`.


arose:
This is a very nice find, thanks.

Another example that is fix by this is 1GRM

1GRM
1GRM (1)

PR #180, opened 2021-05-08
Fix normals in sheet caps

sukolsak:
This PR makes the normal vectors in sheet caps unit vectors while also using one fewer statement.

Part of #175.


arose:
lgtm

PR #189, opened 2021-05-16
Geometry export: support volume & volumeInstance color types

### LINKED ISSUE #184: Geometry export: support volume & volumeInstance color type
#173 will add new `dColorType` values: "volume" | "volumeInstance". Before merging it would be nice to add support for these to the geometry exporters.

New `ColorData` fields:

```
tColorGrid: ValueCell.create(colors),
uColorGridDim: ValueCell.create(Vec3.clone(dimension)),
uColorGridTransform: ValueCell.create(Vec4.clone(transform)),
```

`getTrilinearlyInterpolated` (src\mol-geo\geometry\mesh\color-smoothing.ts) can be used to get trilinearly interpolated colors on the CPU.

sukolsak:
#173 will add new `dColorType` values: "volume" | "volumeInstance". This PR adds support for them in the geometry exporters and resolves #184.

Most changes in this PR are moving the instance loop inside `addMeshWithColors` so that we can call `getTrilinearlyInterpolated` just once. This might also facilitate a potential plan to use the instancing feature in GLTF in the future.

Since vertex coloring [doesn't](https://gamedev.stackexchange.com/questions/21303/how-can-i-include-vertex-color-information-in-obj-files) [seem](https://blender.stackexchange.com/questions/90890/how-to-display-and-use-vertex-color-from-obj-model) to be officially supported in OBJ, exporting raw volume colors can create hundreds of thousands of colors in the MTL file and make some 3D software like Blender take forever to load. So I quantize the colors using the [median cut algorithm](https://en.wikipedia.org/wiki/Median_cut), as suggested by @arose, to limit the number of colors to at most 1024.

This PR also converts sRGB to linear when exporting to GLB.

### DIFF
--- a/src/extensions/geo-export/glb-exporter.ts
+++ b/src/extensions/geo-export/glb-exporter.ts
@@ -94,7 +94,7 @@ export class GlbExporter extends MeshExporter<GlbData> {
                 indices = mesh.indexBuffer.ref.value;
                 groups = mesh.groupBuffer.ref.value;
                 vertexCount = mesh.vertexCount;
-                drawCount = indices.length;
+                drawCount = mesh.triangleCount * 3;
             }
 
             Mat4.fromArray(t, aTransform, instanceIndex * 16);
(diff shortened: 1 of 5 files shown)


dsehnal (approved):
Code looks good, thanks for the contribution!

arose:
Thanks @sukolsak , nice work!

One thing that might be worth testing is if moving the switch statement out of the for loop during getting colors is significantly/noticeably faster.

PR #192, opened 2021-05-20
Geometry export: use instancing in GLB

sukolsak:
glTF/GLB supports [instancing](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#instantiation); The same mesh can be used by multiple nodes, which can have different transformations. The same vertex/normal/index/color buffers can also be used by multiple meshes. This PR tries to reuse meshes and buffers whenever possible.
- Use the same vertex/normal/index buffers for all instances if all instances have the same geometry. That is, the `GraphicsRenderObject`'s type is `mesh` or `texture-mesh`.
- Use the same color buffer for all instances if the color type is not one of the `*Instance` color types and per-group transparency is not used.
- If both conditions are true, use the same `mesh` object for all instances.

Exporting models with many instances such as 1RB8, 6CO8, 6B1T, etc. to GLB should now be very fast. The file size is also much smaller. Without instancing, 1RB8 uses ~900 MB. With instancing, it uses 15 MB. (However, if you change the color theme from "Chain Id" to "Chain Instance", it uses ~300 MB because we have to create a new color buffer for each instance.)

Other changes in this PR:
- Add "Mol* [version]" to the header.
- Remove `min` and `max` from normal/index/color accessors to save some space. They are only required for vertex accessors.
- Remove `indices` from TextureMesh primitives to save some space. They are arrays of serial numbers from 0 to n-1. The [glTF spec](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#primitiveindices) says "When [`primitive.indices`] is not defined, the primitives should be rendered without indices using `drawArrays()`."
- Refactor duplicate code.

### DIFF
--- a/src/extensions/geo-export/mesh-exporter.ts
+++ b/src/extensions/geo-export/mesh-exporter.ts
@@ -111,7 +111,34 @@ export abstract class MeshExporter<D extends RenderObjectExportData> implements 
         return interpolated.array;
     }
 
-    protected abstract addMeshWithColors(inpit: AddMeshInput): void;
+    protected static getInstance(input: AddMeshInput, instanceIndex: number) {
+        const { mesh, meshes } = input;
+        let vertices: Float32Array;
+        let normals: Float32Array;
+        let indices: Uint32Array | undefined;
+        let groups: Float32Array | Uint8Array;
+        let vertexCount: number;
+        let drawCount: number;
+        if (mesh !== undefined) {
+            vertices = mesh.vertices;
+            normals = mesh.normals;
+            indices = mesh.indices;
+            groups = mesh.groups;
+            vertexCount = mesh.vertexCount;
+            drawCount = mesh.drawCount;
+        } else {
+            const mesh = meshes![instanceIndex];
+            vertices = mesh.vertexBuffer.ref.value;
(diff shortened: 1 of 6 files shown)


src/extensions/geo-export/mesh-exporter.ts
```diff
@@ -111,7 +111,34 @@ export abstract class MeshExporter<D extends RenderObjectExportData> implements
...
+        } else {
+            const mesh = meshes![instanceIndex];
+            vertices = mesh.vertexBuffer.ref.value;
+            normals = mesh.normalBuffer.ref.value;
+            indices = mesh.indexBuffer.ref.value;
+            groups = mesh.groupBuffer.ref.value;
+            vertexCount = mesh.vertexCount;
+            drawCount = mesh.triangleCount * 3;
+        }
+        return { vertices, normals, indices, groups, vertexCount, drawCount };
```
dsehnal, line 138: This is a little nitpick, but directly returning the object might be better than assigning mutable values, e.g.:

```ts
return mesh 
  ? { ...mesh }
  : { vertices: ... };
```

sukolsak: Thanks! Fixed.

Issue #132, opened 2021-02-23
ASA coloring wrong after removing part of a chain


arose:
Accessible Surface Area coloring wrong after removing part of a chain. Label/tooltip value seems to be still ok. Only effects the chain that had something removed. Example below is pdb 3sn6

![3SN6](https://user-images.githubusercontent.com/272250/108805871-f3a03e80-7554-11eb-8791-a91308539892.png)
![3SN6 (1)](https://user-images.githubusercontent.com/272250/108805874-f438d500-7554-11eb-9c43-4458b3362a7f.png)

arose:
@JonStargaryen could you have a look at this? Might not be limited to ASA but since you did this, I ask you first :)

JonStargaryen:
Both the color theme as well as the labels use a similar pattern to determine the index in the array of ASA values:
https://github.com/molstar/molstar/blob/d121a11e28a4b23b8f33504ebac1d846bbf89d5b/src/mol-model-props/computed/accessible-surface-area/shrake-rupley.ts#L104
and
https://github.com/molstar/molstar/blob/d121a11e28a4b23b8f33504ebac1d846bbf89d5b/src/mol-plugin/behavior/dynamic/custom-props/computed/accessible-surface-area.ts#L87

The `unit` instance processed by the color theme is missing the deleted elements and maps to incorrect values. The one processed to create labels still contains the initial elements and delivers correct results.

First, the `unit` used to color. The `unit` used to create labels at the bottom.
![units](https://user-images.githubusercontent.com/9744615/108917183-33623700-75e4-11eb-9f90-988cf80cb381.png)

Is it expected that different `units` are processed? It throws me off that labels are correct but the coloring is messed up.

JonStargaryen:
Sorry for the late reply. c6073b8 should fix the issue.

arose:
> Sorry for the late reply. c6073b8 should fix the issue.

Thanks Sebastian. Can confirm that it fixes it for me as well.

JonStargaryen:
Turns out the serial mapping is needed and caused the issues for Assembly 3 of 5CBG in #193.

arose:
Hmm, 3sn6 is still fixed for me :)

arose:
I can't reproduce it anymore after #193. Do you still see it @JonStargaryen or have an idea what fixed it?

JonStargaryen:
[...]

PR #193, opened 2021-05-20
ANVIL fixes

JonStargaryen:
Provides a number of fixes for issues with the ANVIL implementation, which predicts the location of the membrane layer.

- new alpha-only parameter for ASA calculation which is faster and provides values closer to the NACCESS configuration used in the original ANVIL code
- fixes a typo which confused threonine and tryptophan, leading to egregiously bad predictions in some cases
- implements routine to adjust the thickness of the membrane layer
- better way to determine the radius of membrane layer
- add some status updates
- removes remnants of the deleted sphere visuals
- fix for ASA calculation/visual issues - #132 

**Question**
`Mat3.directionTransform` might fail to invert if a structure is oriented exactly along an axis (happens e.g. for 5CBG). Given that, I've changed the code to use the two plane points to construct center and normal vector and always add some 'jitter' to avoid linear dependency.
Is there a cleaner way to create membrane visuals?

Thanks for having a look!

### DIFF
--- a/src/extensions/anvil/algorithm.ts
+++ b/src/extensions/anvil/algorithm.ts
@@ -24,6 +24,7 @@ interface ANVILContext {
     minThickness: number,
     maxThickness: number,
     asaCutoff: number,
+    adjust: number,
 
     offsets: ArrayLike<number>,
     exposed: ArrayLike<boolean>,
@@ -32,11 +33,12 @@ interface ANVILContext {
 };
 
 export const ANVILParams = {
-    numberOfSpherePoints: PD.Numeric(120, { min: 35, max: 700, step: 1 }, { description: 'Number of spheres/directions to test for membrane placement. Original value is 350.' }),
+    numberOfSpherePoints: PD.Numeric(350, { min: 35, max: 700, step: 1 }, { description: 'Number of spheres/directions to test for membrane placement. Original value is 350.' }),
     stepSize: PD.Numeric(1, { min: 0.25, max: 4, step: 0.25 }, { description: 'Thickness of membrane slices that will be tested' }),
     minThickness: PD.Numeric(20, { min: 10, max: 30, step: 1}, { description: 'Minimum membrane thickness used during refinement' }),
     maxThickness: PD.Numeric(40, { min: 30, max: 50, step: 1}, { description: 'Maximum membrane thickness used during refinement' }),
-    asaCutoff: PD.Numeric(40, { min: 10, max: 100, step: 1 }, { description: 'Absolute ASA cutoff above which residues will be considered' })
+    asaCutoff: PD.Numeric(40, { min: 10, max: 100, step: 1 }, { description: 'Relative ASA cutoff above which residues will be considered' }),
+    adjust: PD.Numeric(14, { min: 0, max: 30, step: 1 }, { description: 'Minimum length of membrane-spanning regions (original values: 14 for alpha-helices and 5 for beta sheets). Set to 0 to not optimize membrane thickness.' })
 };
 export type ANVILParams = typeof ANVILParams
 export type ANVILProps = PD.Values<ANVILParams>
@@ -58,18 +60,20 @@ export function computeANVIL(structure: Structure, props: ANVILProps) {
 const centroidHelper = new CentroidHelper();
 function initialize(structure: Structure, props: ANVILProps): ANVILContext {
     const l = StructureElement.Location.create(structure);
-    const { label_atom_id, x, y, z } = StructureProperties.atom;
+    const { label_atom_id, label_comp_id, x, y, z } = StructureProperties.atom;
     const elementCount = structure.polymerResidueCount;
     centroidHelper.reset();
 
-    let offsets = new Int32Array(elementCount);
+    let offsets = new Array<number>(elementCount);
     let exposed = new Array<boolean>(elementCount);
 
     const accessibleSurfaceArea = structure && AccessibleSurfaceAreaProvider.get(structure);
     const asa = accessibleSurfaceArea.value!;
+    const asaCutoff = props.asaCutoff / 100;
 
     const vec = Vec3();
     let m = 0;
+    let e = 0, b = 0;
     for (let i = 0, il = structure.units.length; i < il; ++i) {
         const unit = structure.units[i];
         const { elements } = unit;
@@ -95,13 +99,19 @@ function initialize(structure: Structure, props: ANVILProps): ANVILContext {
 
             // keep track of offsets and exposed state to reuse
             offsets[m] = structure.serialMapping.getSerialIndex(l.unit, l.element);
-            exposed[m] = AccessibleSurfaceArea.getValue(l, asa) > props.asaCutoff;
-
+            exposed[m] = AccessibleSurfaceArea.getValue(l, asa) / MaxAsa[label_comp_id(l)] > asaCutoff;
+            if (AccessibleSurfaceArea.getValue(l, asa) / MaxAsa[label_comp_id(l)] > asaCutoff) {
+                e++;
+            } else {
+                b++;
+            }
             m++;
         }
     }
+    console.log('CAs = ' + m);
+    console.log('exposed ' + e + ' - buried ' + b);
 
-    // omit potentially empty tail1
+    // omit potentially empty tail
     offsets = offsets.slice(0, m);
     exposed = exposed.slice(0, m);
 
@@ -114,28 +124,52 @@ function initialize(structure: Structure, props: ANVILProps): ANVILContext {
         centroidHelper.radiusStep(vec);
     }
     const extent = 1.2 * Math.sqrt(centroidHelper.radiusSq);
+    console.log(`center: ${centroid}, radius: ${Math.sqrt(centroidHelper.radiusSq)}`);
 
     return {
         ...props,
-        structure: structure,
+        structure,
 
-        offsets: offsets,
-        exposed: exposed,
-        centroid: centroid,
-        extent: extent
+        offsets,
+        exposed,
+        centroid,
+        extent
     };
 }
 
 export async function calculate(runtime: RuntimeContext, structure: Structure, params: ANVILProps): Promise<MembraneOrientation> {
-    const { label_comp_id } = StructureProperties.atom;
-
     const ctx = initialize(structure, params);
-    const initialHphobHphil = HphobHphil.filtered(ctx, label_comp_id);
+    const initialHphobHphil = HphobHphil.filtered(ctx);
+    console.log(`init: ${initialHphobHphil.hphob} - ${initialHphobHphil.hphil}`);
 
-    const initialMembrane = findMembrane(ctx, generateSpherePoints(ctx, ctx.numberOfSpherePoints), initialHphobHphil, label_comp_id);
-    const alternativeMembrane = findMembrane(ctx, findProximateAxes(ctx, initialMembrane), initialHphobHphil, label_comp_id);
+    if (runtime.shouldUpdate) {
+        await runtime.update({ message: 'Placing initial membrane...' });
+    }
+    console.time('ini-membrane');
+    const initialMembrane = findMembrane(ctx, generateSpherePoints(ctx, ctx.numberOfSpherePoints), initialHphobHphil)!;
+    console.timeEnd('ini-membrane');
+    console.log(`initial: ${initialMembrane.qmax}`);
 
-    const membrane = initialMembrane.qmax! > alternativeMembrane.qmax! ? initialMembrane : alternativeMembrane;
+    if (runtime.shouldUpdate) {
+        await runtime.update({ message: 'Refining membrane placement...' });
+    }
+    console.time('ref-membrane');
+    const refinedMembrane = findMembrane(ctx, findProximateAxes(ctx, initialMembrane), initialHphobHphil)!;
+    console.timeEnd('ref-membrane');
+    console.log(`refined: ${refinedMembrane.qmax}`);
+    let membrane = initialMembrane.qmax! > refinedMembrane.qmax! ? initialMembrane : refinedMembrane;
+
+    if (ctx.adjust) {
+        if (runtime.shouldUpdate) {
+            await runtime.update({ message: 'Adjusting membrane thickness...' });
+        }
+        console.time('adj-thickness');
+        membrane = adjustThickness(ctx, membrane, initialHphobHphil);
+        console.timeEnd('adj-thickness');
+        console.log('Membrane width: ' + Vec3.distance(membrane.planePoint1, membrane.planePoint2));
+    }
+
+
 
     return {
         planePoint1: membrane.planePoint1,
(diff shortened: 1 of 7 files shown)


arose (changes requested):
Thank you @JonStargaryen, this looks like quite a nice improvement.

Can you please provide a code example or unit test for the issue with "structure is oriented exactly along an axis". I tried to reproduce it with 5CBG Assembly 1 after commenting out your jitter code but did not see any error.

When I try this for Assembly 3 of 5CBG I get "TypeError: Cannot read property 'spherePoint' of undefined at findProximateAxes"

The membrane plane for Assembly 2 of 5CBG look wrong compared to Assembly 1.

I looked at an extreme example 7CGO. The refinement step takes very long. Can this be improved or can you show more progress? For such long running tasks it usually helps give control back to the browser by showing progress. Also using StructureProperties.atom.x/y/z in tight loops is not the fastest. Probably worthwhile to get the positions manually. After >5min (with profiling enabled in Chrome) it found a reasonable placement - strictly there should be two bilayers :)

image

image

image


Much better than the old version which took >10min with most of the time after the ASA calculation as well.

image

src/mol-model-props/computed/accessible-surface-area/shrake-rupley.ts
```diff
@@ -18,7 +18,8 @@ export const ShrakeRupleyComputationParams = {
     numberOfSpherePoints: PD.Numeric(92, { min: 12, max: 360, step: 1 }, { description: 'Number of sphere points to sample per atom: 92 (original paper), 960 (BioJava), 3000 (EPPIC) - see Shrake A, Rupley JA: Environment and exposure to solvent of protein atoms. Lysozyme and insulin. J Mol Biol 1973.' }),
     probeSize: PD.Numeric(1.4, { min: 0.1, max: 4, step: 0.01 }, { description: 'Corresponds to the size of a water molecule: 1.4 (original paper), 1.5 (occassionally used)' }),
     // buriedRasaThreshold: PD.Numeric(0.16, { min: 0.0, max: 1.0 }, { description: 'below this cutoff of relative accessible surface area a residue will be considered buried - see: Rost B, Sander C: Conservation and prediction of solvent accessibility in protein families. Proteins 1994.' }),
-    nonPolymer: PD.Boolean(false, { description: 'Include non-polymer atoms as occluders.' })
+    nonPolymer: PD.Boolean(false, { description: 'Include non-polymer atoms as occluders.' }),
+    alphaOnly: PD.Boolean(false, { description: 'Compute only using alpha-carbons, if true increase probeSize accordingly (e.g., 4 A).' })
```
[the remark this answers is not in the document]
JonStargaryen: Done in 30d5b0d.

src/mol-model-props/computed/accessible-surface-area/shrake-rupley/radii.ts
```diff
@@ -50,15 +50,15 @@ export function assignRadiusForHeavyAtoms(ctx: ShrakeRupleyContext) {
                 continue;
             }
 
+            const atomId = label_atom_id(l);
             const moleculeType = getElementMoleculeType(unit, eI);
             // skip water and optionally non-polymer groups
-            if (moleculeType === MoleculeType.Water || (!ctx.nonPolymer && !isPolymer(moleculeType))) {
+            if (moleculeType === MoleculeType.Water || (!ctx.nonPolymer && !isPolymer(moleculeType)) || (ctx.alphaOnly && (atomId !== 'CA' || !MaxAsa[label_comp_id(l)]))) {
```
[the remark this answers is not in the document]
JonStargaryen: 'BB' check added in 74aa24b.

src/extensions/anvil/prop.ts
```diff
@@ -76,7 +76,8 @@ export const MembraneOrientationProvider: CustomStructureProperty.Provider<Membr
 });
 
 async function computeAnvil(ctx: CustomProperty.Context, data: Structure, props: Partial<ANVILProps>): Promise<MembraneOrientation> {
-    await AccessibleSurfaceAreaProvider.attach(ctx, data);
+    // can't get away with the default 92 points here
+    await AccessibleSurfaceAreaProvider.attach(ctx, data, { probeSize: 4.0, alphaOnly: true, numberOfSpherePoints: 184 });
```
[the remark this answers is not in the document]
JonStargaryen: 3cf67f7 fixes the coloring for trace-only ASA values. This config might also be useful for virus capsids.

JonStargaryen:
> Can you please provide a code example or unit test for the issue with "structure is oriented exactly along an axis". I tried to reproduce it with 5CBG Assembly 1 after commenting out your jitter code but did not see any error.

I've revisited this and used the code from the assembly symmetry visuals (df89351).

> When I try this for Assembly 3 of 5CBG I get "TypeError: Cannot read property 'spherePoint' of undefined at findProximateAxes"

This was caused by the serial mapping issues in the ASA calculation. Now the behavior should be correct for units that start with some offset and also stay consistent when parts of the structure are removed.

> The membrane plane for Assembly 2 of 5CBG look wrong compared to Assembly 1.

Is this still the case? Looks good for me now.

> I looked at an extreme example 7CGO. The refinement step takes very long. Can this be improved or can you show more progress? For such long running tasks it usually helps give control back to the browser by showing progress. Also using StructureProperties.atom.x/y/z in tight loops is not the fastest. Probably worthwhile to get the positions manually. After >5min (with profiling enabled in Chrome) it found a reasonable placement - strictly there should be two bilayers :)

I've added more fine-grained status updates in 4e77699. Getting coordinates manually means inlining the x/y/z methods, like in a393231?

src/mol-model-props/computed/accessible-surface-area/shrake-rupley.ts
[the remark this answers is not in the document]
arose: Thanks!

src/mol-model-props/computed/accessible-surface-area/shrake-rupley/radii.ts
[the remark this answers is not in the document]
arose: Thanks!

arose:
Thanks @JonStargaryen 

> I've added more fine-grained status updates in 4e77699. Getting coordinates manually means inlining the x/y/z methods, like in a393231?

That is a start. I think you can inline the whole setLocation and get rid of using a location at all. And you can assign unitIndices and elementIndices to variables outside the loop so there is less pointer chasing inside the loop. More generally, I think HphobHphil.filtered could be less quadratic. How about calculating to which slice every exposed residue belongs once for all slices in all directions and look up member ship later. That way you only call conformation.x/y/z once for each residue.

Also, could you try assigning the Vec3 methods locally, like in src\mol-geo\geometry\mesh\builder\ribbon.ts, for example. I have seen quite an effect with that in hot loops.

src/extensions/anvil/prop.ts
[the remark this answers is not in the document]
arose: That looks much better, thanks!

My concern is that now the ASA coloring (and values) looks very different depending on whether the user has added the Membrane Orientation visual or not. That is way I suggested the second option.

3PQR (3)
3PQR (4)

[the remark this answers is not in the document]
JonStargaryen: I see your point. With 609654b ANVIL won't attach the rather specific ASA values.

JonStargaryen:
> That is a start. I think you can inline the whole setLocation and get rid of using a location at all. And you can assign unitIndices and elementIndices to variables outside the loop so there is less pointer chasing inside the loop. More generally, I think HphobHphil.filtered could be less quadratic. How about calculating to which slice every exposed residue belongs once for all slices in all directions and look up member ship later. That way you only call conformation.x/y/z once for each residue.
> 
> Also, could you try assigning the Vec3 methods locally, like in src\mol-geo\geometry\mesh\builder\ribbon.ts, for example. I have seen quite an effect with that in hot loops.

Thanks for the array of tips! I've implemented them. Collectively, this shaved off ~25-40% computation time with the majority probably coming from the smarter HphobHphil calculation.
Even huge structures like assembly 1 of 6NK5 now perform reasonably well, not meaningful but it's a nice test case.

src/extensions/anvil/algorithm.ts
```diff
@@ -83,6 +84,7 @@ async function initialize(runtime: RuntimeContext, structure: Structure, props:
 
     let offsets = new Array<number>(elementCount);
```
arose, line 86: If you are using Array (as opposed to a typed arrays) you can .push elements without performance penalty and then you don't need to slice them later which creates a new copy. Alternatively you can use typed arrays and then get correctly sized ones without copying by using TypedArray.buffer like new Uint32Array(ta.buffer, 0, n).

src/extensions/anvil/prop.ts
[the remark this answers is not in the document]
arose: Super, thanks

src/extensions/anvil/algorithm.ts
```diff
@@ -160,82 +203,94 @@ namespace MembraneCandidate {
...
-    for (let i = 0, il = spherePoints.length; i < il; i++) {
-        const spherePoint = spherePoints[i];
-        Vec3.sub(diam, centroid, spherePoint);
-        Vec3.scale(diam, diam, 2);
-        const diamNorm = Vec3.magnitude(diam);
-        const qvartemp = [];
+    const diam = v3zero();
+    const testPoint = v3zero();
+    for (let n = 0, nl = spherePoints.length; n < nl; n++) {
+        if (runtime?.shouldUpdate && message && (n + 1) % UPDATE_INTERVAL === 0) {
```
arose, line 237: runtime is always defined

src/extensions/anvil/algorithm.ts
```diff
@@ -262,13 +490,17 @@ function qValue(currentStats: HphobHphil, initialStats: HphobHphil): number {
...
+}
+
+function _isInMembranePlane(testPoint: Vec3, normalVector: Vec3, min: number, max: number): boolean {
+    const d = -v3dot(normalVector, testPoint);
+    return d > min && d < max;
 }
 
-// generates a defined number of points on a sphere with radius = extent around the specified centroid
+/** Generates a defined number of points on a sphere with radius = extent around the specified centroid */
 function generateSpherePoints(ctx: ANVILContext, numberOfSpherePoints: number): Vec3[] {
```
arose, line 504: Does this take symmetry into account? Would a vector from south to north be the same for testing as one from north to south?

arose:
Thanks @JonStargaryen, this is getting better and better. 7CGO is now down to ~50s for me. Majority of time is still spend in .filtered. For structures like that with many residues a higher task update rate would be good.

src/extensions/anvil/algorithm.ts
JonStargaryen: Got it. Moved to `.push` and removed `.slice` in 31f4803.

src/extensions/anvil/algorithm.ts
JonStargaryen: Removed by 3849c34.

src/extensions/anvil/algorithm.ts
JonStargaryen: Somewhat. The direction is symmetric but the mirrored sphere point usually leads to slightly different results if the extent of the structure is not exactly divisible by the stepSize. Every direction is evaluated twice with a slight offset, but no efforts are made to choose a good offset (like half the stepSize).

Given that, one hemisphere could be sampled by double the number of sphere points while the other hemisphere gets ignored.

JonStargaryen:
> Thanks @JonStargaryen, this is getting better and better. 7CGO is now down to ~50s for me. Majority of time is still spend in .filtered. For structures like that with many residues a higher task update rate would be good.

I've had another look at the `filtered` method. Actually, the offsets are meaningless as they were only used to compute HphobHphil stats of each slice. Now only the final stats of each slice are stored, which avoids traversing exposed residues every time the stats of a slice are needed (54a388d).

7CGO is down to ~20 s. Now the most expensive methods are `indexOf` (b47d046) during ASA value lookup and `adjustThickness` for larger structures like 6NJM.

5eafddf adds more fine-grained updates for large structures.

src/extensions/anvil/algorithm.ts
JonStargaryen: fcac1a6 moves all sphere points to one hemisphere. I've ran some test cases and the predictions are fine.

src/mol-model-props/computed/accessible-surface-area/shrake-rupley.ts
```diff
@@ -99,13 +101,32 @@ namespace AccessibleSurfaceArea {
...
     export function getValue(location: StructureElement.Location, accessibleSurfaceArea: AccessibleSurfaceArea) {
-        const { getSerialIndex } = location.structure.root.serialMapping;
         const { area, serialResidueIndex } = accessibleSurfaceArea;
-        const rSI = serialResidueIndex[getSerialIndex(location.unit, location.element)];
+        const rSI = serialResidueIndex[indexOf(location.structure.root.serialMapping.elementIndices, location.element)];
         if (rSI === -1) return -1;
         return area[rSI];
     }
 
+    function indexOf(a: ArrayLike<number>, e: number): number {
```
dsehnal, line 110: Binary search is already implemented in [``SortedArray.indexOf``](https://github.com/molstar/molstar/blob/master/src/mol-data/int/sorted-array.ts#L23)

This seems like the code was copied from there. You can call the function directly from ``https://github.com/molstar/molstar/blob/master/src/mol-data/int/impl/sorted-array.ts#L45`` or use ``SortedArray.indexOf(a as SortedArray), e);``.

JonStargaryen: Thanks @dsehnal, fixed with e1b771b. I only found the private impl in:
https://github.com/molstar/molstar/blob/b53debcfef324066f71ba0f60a8b3f45ef6f2a31/src/perf-tests/binary-search.ts#L12

JonStargaryen:
As discussed, the recent commits revisit the ASA calculation. @josemduarte recently tested optimizations from Eisenhaber et al., 1995 (https://doi.org/10.1002/jcc.540160303) and found them to be really beneficial in the BioJava implementation. Mainly, calculated values are kept in memory and neighbors are sorted by distance to break the inner loop as fast as possible.

3J3Q was >2 minutes and is now down to ~1 minute. With the ANVIL configuration (traceOnly + 4.0 A probe size) it's at ~15 s.

src/mol-model-props/computed/accessible-surface-area/shrake-rupley.ts
```diff
@@ -102,31 +103,11 @@ namespace AccessibleSurfaceArea {
 
     export function getValue(location: StructureElement.Location, accessibleSurfaceArea: AccessibleSurfaceArea) {
         const { area, serialResidueIndex } = accessibleSurfaceArea;
-        const rSI = serialResidueIndex[indexOf(location.structure.root.serialMapping.elementIndices, location.element)];
+        const rSI = serialResidueIndex[SortedArray.indexOf(location.structure.root.serialMapping.elementIndices as unknown as SortedArray, location.element)];
```
arose, line 106: You can use SortedArray.ofSortedArray() instead of manually casting

src/extensions/anvil/algorithm.ts
arose: Very nice, thank you

arose (approved):

arose:
@JonStargaryen This is wonderful work. I am happy with this getting merged whenever you want.

src/mol-model-props/computed/accessible-surface-area/shrake-rupley.ts
JonStargaryen: Done with f930e3d.
Feel free to merge at any time and thanks again for the effective review!

PR #194, opened 2021-05-22
Fix HTMLImageElement check for SMAA

sukolsak:
`typeof` always returns a string.


arose:
Ha, good find. The check is for running on NodeJS.

sukolsak:
@arose There is another one in [mol-gl/webgl/texture.ts](https://github.com/molstar/molstar/blob/ffdcf798e0b23905d6a9e765e5e4d53ac318c99b/src/mol-gl/webgl/texture.ts#L172). (I only searched for "`=== undefined`" and missed this.)

arose:
thanks, fixed in 33de60d365c5735636d9a8abc1d5cbe7d9f369d0

PR #195, opened 2021-05-25
Update node-sass and webpack dependencies

MadCatX:
Currently, Molstar depends on an older version of node-sass, which, in turn, appears to pull in an older version of libSass. Older libSass does not compile correctly on machines with recent C++ development toolchain, rendering Molstar unbuildable on such machines.

Update of node-sass is slightly non-trivial as it requires an update of a few other modules too. Please consider updating Molstar dependencies as suggested by the PR.


dsehnal:
Thanks a lot @MadCatX for taking the time to do this, the update was long over due.

PR #197, opened 2021-05-27
Geometry export: add material to GLB

sukolsak:
Add a material to the glTF/GLB exporter based on the global renderer style, as @arose suggested.

I can't adjust the light intensity, roughness, etc. when choosing the "Custom" lighting mode in the viewer. Is this intentional?

### DIFF
--- a/src/extensions/geo-export/controls.ts
+++ b/src/extensions/geo-export/controls.ts
@@ -4,6 +4,7 @@
  * @author Sukolsak Sakshuwong <sukolsak@stanford.edu>
  */
 
+import { getStyle } from '../../mol-gl/renderer';
 import { PluginComponent } from '../../mol-plugin-state/component';
 import { PluginContext } from '../../mol-plugin/context';
 import { Task } from '../../mol-task';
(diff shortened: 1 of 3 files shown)


dsehnal:
Thanks for the PR.

> I can't adjust the light intensity, roughness, etc. when choosing the "Custom" lighting mode in the viewer. Is this intentional?

You can adjust these in the full settings panel here:

image

Perhaps we should add these to the "simplified" view too.

Do you think supporting per object settings of these parameters would be useful?

arose:
> Perhaps we should add these to the "simplified" view too.

Yeah, we should... :)

> Do you think supporting per object settings of these parameters would be useful?

Definitely and it should be no problem to add it to the exporter once I implemented per-object material support.

arose:
Thank you @sukolsak, this looks good.

sukolsak:
> Thanks for the PR.
> 
> > I can't adjust the light intensity, roughness, etc. when choosing the "Custom" lighting mode in the viewer. Is this intentional?
> 
> You can adjust these in the full settings panel here:

Thanks! I didn't know that this button existed. Yes, having those options in the "simplified" view would be helpful. Or at least point the users to the options in the "advanced" view.

> Do you think supporting per object settings of these parameters would be useful?

Yes, I think so.

PR #198, opened 2021-05-28
Geometry export: auto adjust quality of sphere/cylinder meshes

sukolsak:
Automatically adjust the quality of sphere/cylinder meshes, as @arose suggested. This should help reduce the exported file size when using spacefill with large models.


dsehnal:
Would it be possible to do the sphere/cylinder geometry instancing too?

There could be a single sphere and cylinder instance per ``ball-and-stick`` representation. The transformation is then a composition of the instance matrix and current position translation/radius scale/rotation (for bonds).

arose:
Thank you @sukolsak!

> Would it be possible to do the sphere/cylinder geometry instancing too?

Would definitely be great. Let's discuss this in a separate issue.

PR #199, opened 2021-05-29
Better handling of processed meshes

arose:
@sukolsak This adds a new property to mesh render-objects which holds some original data after applying uniformTriangleGroup to a mesh. That original data can be used to do the export and revert the change to the uniformTriangleGroup calls (bc998ab328c5a2063aa5986d1e5bb03b1d89b608). With that the file size should be the same for webgl1 and 2.

```ts
const ro = renderObjects[i];
if ('meta' in ro.values && ro.values.meta.ref.value.originalData) {
    const { indexBuffer, vertexCount, triangleCount } = ro.values.meta.ref.value.originalData;
}
```

### DIFF
--- a/src/extensions/geo-export/mesh-exporter.ts
+++ b/src/extensions/geo-export/mesh-exporter.ts
@@ -133,12 +133,22 @@ export abstract class MeshExporter<D extends RenderObjectExportData> implements 
     private async addMesh(values: MeshValues, webgl: WebGLContext, ctx: RuntimeContext) {
         const aPosition = values.aPosition.ref.value;
         const aNormal = values.aNormal.ref.value;
-        const elements = values.elements.ref.value;
         const aGroup = values.aGroup.ref.value;
-        const vertexCount = values.uVertexCount.ref.value;
-        const drawCount = values.drawCount.ref.value;
+        const originalData = Mesh.getOriginalData(values);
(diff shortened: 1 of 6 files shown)


sukolsak:
Thanks! I just tried it at https://github.com/sukolsak/molstar/commit/dd510fcd5743f1c6251bd2fb42ea21520644bb51 and can confirm that the file size is the same for WebGL1 and 2. 1RB8 is now 1.6 MB in both Chrome and Safari.

Currently, the type of `values.meta.ref.value.originalData` is `any`. Is it possible to make it `OriginalData`? Or should we export the `OriginalData` type and cast `originalData` to it?

dsehnal:
> ```ts
> 'meta' in ro.values
> ```

``'meta' in ro.values``: this evaluates to true if ``meta === undefined``:

image

arose:
> 'meta' in ro.values: this evaluates to true when meta === undefined

Typescript will tell us. RenderObjectValues only contains one meta property which is a ValueCell

arose:
> Currently, the type of values.meta.ref.value.originalData is any. Is it possible to make it OriginalData? Or should we export the OriginalData type and cast originalData to it?

Good point. I added `Mesh.getOriginalData` in a1e05387e42077f4aee0b7891f8c477f894e5418.

PR #203, opened 2021-05-30
Geometry export: use original mesh data

sukolsak:
#199 will add a new property to mesh render-objects which holds some original data after applying uniformTriangleGroup to a mesh. This PR uses that property to export meshes and also reverts bc998ab. This should make the export file size the same for WebGL1 and WebGL2. For example, 1RB8.glb is currently 4.7 MB in Safari and 1.6 MB in Chrome. With this PR, it will be 1.6 MB in both browsers.


arose:
Thank you @sukolsak

PR #212, opened 2021-06-14
Geometry export: center exported models

sukolsak:
When importing some models such as 7NCK to Blender using the geometry exporter, the models are very off the center and make it look like nothing is imported. This PR centers exported models.

This will also be useful when we implement the USDZ format for AR. Apple says that "Base of object should sit on the ground plane (y = 0)" (slide 30 in https://devstreaming-cdn.apple.com/videos/wwdc/2018/603augiuv41xoowslk8/603/603_integrating_apps_and_content_with_ar_quick_look.pdf). iOS AR doesn't automatically center the object. I'm not sure about Android.

Note: I don't factor out the translation matrix calculation because it will be a bit different for USDZ.

### DIFF
--- a/src/extensions/geo-export/controls.ts
+++ b/src/extensions/geo-export/controls.ts
@@ -5,6 +5,7 @@
  */
 
 import { getStyle } from '../../mol-gl/renderer';
+import { Box3D } from '../../mol-math/geometry';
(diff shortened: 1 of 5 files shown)


dsehnal (approved):
Thanks for the PR, looks good.

PR #215, opened 2021-06-22
[RFC] Do not assume that atom_site.auth_seq_id is an integer

MadCatX:
According to multiple resources ([IURC](http://ww1.iucr.org/iucr-top/cif/cifdic_html/2/cif_mm.dic/Iatom_site.auth_seq_id.html), [WWPDB](https://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v50.dic/Items/_atom_site.auth_seq_id.html), [Molstar code](https://github.com/molstar/molstar/blob/daa2bbd042b5541ba1e6dd23adfdc97c07ac8701/src/mol-io/reader/cif/schema/mmcif.ts#L52)) the value of `atom_site.auth_seq_id` field in mmCif is a string and no assumptions shall be made about its value. Molstar assumes that this field either directly mirrors `atom_site.label_seq_id` or at least contains a sensible integer.

This patch attempts to fix the issue. I'm submitting it as an RFC because the move from integral to string representation of `atom_cite.auth_seq_id` touches some quite low-level Molstar code and there is a pretty good chance that this change has some implications that I'm not seeing.

### DIFF
--- a/src/extensions/anvil/algorithm.ts
+++ b/src/extensions/anvil/algorithm.ts
@@ -324,7 +324,7 @@ function membraneSegments(ctx: ANVILContext, membrane: MembraneCandidate): Array
     const { units } = structure;
     const { elementIndices, unitIndices } = structure.serialMapping;
     const testPoint = v3zero();
-    const { auth_seq_id } = StructureProperties.residue;
+    const { label_seq_id } = StructureProperties.residue;
 
     const d1 = -v3dot(normalVector!, planePoint1);
     const d2 = -v3dot(normalVector!, planePoint2);
@@ -334,10 +334,10 @@ function membraneSegments(ctx: ANVILContext, membrane: MembraneCandidate): Array
     const inMembrane: { [k: string]: Set<number> } = Object.create(null);
     const outMembrane: { [k: string]: Set<number> } = Object.create(null);
     const segments: Array<{ start: number, end: number }> = [];
-    let authAsymId;
-    let lastAuthAsymId = null;
-    let authSeqId;
-    let lastAuthSeqId = units[0].model.atomicHierarchy.residues.auth_seq_id.value((units[0] as Unit.Atomic).chainIndex[0]) - 1;
+    let labelAsymId;
+    let lastLabelAsymId = null;
+    let labelSeqId;
+    let lastLabelSeqId = units[0].model.atomicHierarchy.residues.label_seq_id.value((units[0] as Unit.Atomic).chainIndex[0]) - 1;
     let startOffset = 0;
     let endOffset = 0;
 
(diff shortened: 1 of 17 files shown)


dsehnal:
Thanks for the PR and taking the time to investigate it. Here are a couple of question to get started:

- Do you have any examples of PDB files that do not have integer values in ``auth_seq_id``?
- Would an acceptable solution be to keep the default to ``int``, but add an easy way to access the optional non-int value as well? I can imagine most users would benefit from treating the seq id as an integer.
- Or add a utility function that modifies the ``auth_seq_id`` column to always contain an integer and possibly copy the original value to an extra column?
- As far as I know, the mirroring only happens for HET atoms that have undefined ``label_seq_id``.
- Even in PDB APIs, the value is treated as an integer, for example [here](https://www.ebi.ac.uk/pdbe/api/mappings/1tqn):
 
image

MadCatX:
Allow me to bore you with a bit of context. The [NtC step id](https://mmcif.wwpdb.org/dictionaries/mmcif_ndb_ntc.dic/Items/_ndb_struct_ntc_step.id.html) item is composed of a few residue identifiers, including `auth_seq_id`. (Example step from 1bna structutre: 1bna_A_DC1_DG2 (asym_id, chain_id, two label_comp_ids and auth_seq_ids))

The algorithm that creates NtC step ids treats `auth_seq_id` as string. Converting it to an integer may lead to weird issues. For instance, if the `auth_seq_id` is given as `0001` in the mmCif file, conversion to integer loses the leading zeroes and the DNATCO Molstar extenstion then creates wrong step id and fails to display NtC annotation data. That is how I found the issue in the first place.

Having some way to access raw string values of `auth_seq_id` would fix my issue of not being able to reliably generate NtC step ids in Molstar.

MadCatX:
I sometimes have to deal with artificially generated structures like this (I kept only the first three atoms for brevity):

NtC-annotated mmCif:

    loop_
    _atom_site.group_PDB
    _atom_site.id
    _atom_site.type_symbol
    _atom_site.label_atom_id
    _atom_site.label_alt_id
    _atom_site.label_comp_id
    _atom_site.label_asym_id
    _atom_site.label_entity_id
    _atom_site.label_seq_id
    _atom_site.pdbx_PDB_ins_code
    _atom_site.Cartn_x
    _atom_site.Cartn_y
    _atom_site.Cartn_z
    _atom_site.occupancy
    _atom_site.B_iso_or_equiv
    _atom_site.pdbx_formal_charge
    _atom_site.auth_seq_id
    _atom_site.auth_comp_id
    _atom_site.auth_asym_id
    _atom_site.auth_atom_id
    _atom_site.pdbx_auth_seq_id
    _atom_site.pdbx_auth_comp_id
    _atom_site.pdbx_auth_asym_id
    _atom_site.pdbx_auth_atom_name
    _atom_site.pdbx_PDB_model_num
    ATOM    1  O  "O5'"   .  A  A  1  1  ?  -2.177    0.911   1.582  1.00  0.00  ?  0001  A  A  "O5'"   0001  A  A  "O5'"   1  
    ATOM    2  C  "C5'"   .  A  A  1  1  ?  -3.209    1.589   0.876  1.00  0.00  ?  0001  A  A  "C5'"   0001  A  A  "C5'"   1  
    ATOM    3  C  "C4'"   .  A  A  1  1  ?  -4.324    0.621   0.562  1.00  0.00  ?  0001  A  A  "C4'"   0001  A  A  "C4'"   1 

created by annotating the following PDB file:

    ATOM  00001  O5' A   A0001      -2.177   0.911   1.582  1.00  0.00           O
    ATOM  00002  C5' A   A0001      -3.209   1.589   0.876  1.00  0.00           C
    ATOM  00003 H5'1 A   A0001      -2.800   2.003  -0.071  1.00  0.00           H

As far as my understanding goes these are perfectly valid mmCif and PDB files which have the str vs. int issue in `auth_seq_id`.

The workaround you suggested allows me to generate correct NtC step ids. Once I'm sure that it really fixes the problem I can submit a patch for the DNATCO extension.

Thanks!

dsehnal:
Thanks for looking into the workaround. I will give some more thought to the string fields too. Ultimatetely, I think it would be better if we followed the standard but certainly is a non-trivial change.

Also regarding https://github.com/molstar/molstar/pull/77 -- we kind of let this one die out. Would you like that one merged?

MadCatX:
Okay, the workaround will fine for me for the time being.

As for #77, can I back to you on that one? I'd like to get this merged in some way but I'd rather use my latest code than what is in #77. I'll have to take a look at my current code and do a new PR.

Thanks for your help with this.

dsehnal:
New PR sounds good and no problem. Glad I could help.

wojdyr:
>     * Do you have any examples of PDB files that do not have integer values in `auth_seq_id`?
> 

auth_seq_id was originally used for sequence ID (sequence number + insertion code), so it could be non-integer (`10A`).
Then the PDB started using this field as the sequence number, moving the insertion code to a separate field. The tag name and the documentation has never been updated, though.
IIRC I've seen such old files somewhere in the wwpdb archive, perhaps in the obsolete directory. But I'm not 100% sure.

arose:
Closing for now, we can revisit this if it turns out to be a problem for more users.

PR #217, opened 2021-06-27
Fix cylinder orientation

sukolsak:
If `Cylinder.setCylinderMat` is called for the first time with `matchDir` set to false, `tmpUp` will be a zero vector and the cylinder will have wrong orientation as shown below.
1tqn

I discovered this bug while trying to update [molrender](https://github.com/molstar/molrender). It is a bit hard to reproduce this bug in the viewer without modifying the source code, because the axes are shown by default and the cylinders in the axes are created with `matchDir` set to true.

Another way to fix this is to simply change `var tmpUp = Vec3();` to `var tmpUp = Vec3.create(0, 1, 0);` But I think this fix is better because it eliminates side effects from `setCylinderMat`.


dsehnal:
Thanks for the fix, looks good.

Issue #178, opened 2021-05-07
MOL2 format visualization


stefdoerr:
Hi I have an issue with molstar visualizing MOL2 files. 
If I pass it a water molecule it's missing a bond which is correctly defined in the file

![image](https://user-images.githubusercontent.com/7935362/117405896-ffc15880-af14-11eb-8cd2-45027ac53083.png)

```
@<TRIPOS>MOLECULE
    HOH
    3     2     0     0     0
SMALL
USER_CHARGES


@<TRIPOS>ATOM
      1 O           0.0020    0.3832    0.0000 oh         1 HOH     -0.785000
      2 H          -0.7702   -0.1875    0.0000 ho         0 HOH      0.392000
      3 H0          0.7682   -0.1957    0.0000 ho         0 HOH      0.392000
@<TRIPOS>BOND
     1    1    2 1
     2    1    3 1
```

Also as an extra question. While it handles single, double and triple bonds correct it doesn't visualize unkekulized aromatic rings. Is there any option to enable that?

![image](https://user-images.githubusercontent.com/7935362/117406385-a73e8b00-af15-11eb-91da-6c2c0d3431cb.png)
vs PyMOL
![image](https://user-images.githubusercontent.com/7935362/117406692-1320f380-af16-11eb-81b9-e649b27b549d.png)

```
@<TRIPOS>MOLECULE
    UNL
   30    32     0     0     0
SMALL
USER_CHARGES


@<TRIPOS>ATOM
      1 C           3.2867   -0.7761    0.4714 ca         1 UNL     -0.172500
      2 C0          4.6360   -0.6536    0.1004 ca         1 UNL     -0.104500
      3 C1          5.0354    0.3643   -0.7794 ca         1 UNL     -0.161000
      4 C2          4.0828    1.2570   -1.2941 ca         1 UNL     -0.104500
      5 C3          2.7334    1.1351   -0.9239 ca         1 UNL     -0.172500
      6 C4          2.3272    0.1277   -0.0305 ca         1 UNL      0.074600
      7 N           0.9486    0.0266    0.3058 ns         1 UNL     -0.469100
      8 C5          0.3788   -0.4232    1.4857 za         1 UNL      0.715300
      9 O           1.0370   -0.7522    2.4571 zb         1 UNL     -0.606100
     10 C6         -1.1051   -0.4483    1.4824 cc         1 UNL     -0.230400
     11 C7         -2.0636   -0.1265    0.4348 ca         1 UNL     -0.062800
[...]

PR #221, opened 2021-07-04
Aromatic bond display option

arose:
This adds an `aromaticBonds` option to the Ball & Stick representation defaulting to `false` (maybe set to `true` for smaller systems?). Aromaticity of bonds is taken from input data and is calculated calculated for rings (flat and containing only certain types of atoms). Aromaticity is stored separately from bond order (as done in mmcif as well), so you can view the kekulized form, if available, by  setting `aromaticBonds` to false.

Addresses #1 and #178. Ping @stefdoerr.

image

image


src/mol-util/array.ts
```diff
@@ -105,6 +105,21 @@ export function arraySetRemove<T>(xs: T[], x: T) {
...
+    return false;
+}
+
+export function arrayIntersectionSize<T>(xs: T[], ys: T[]) {
+    let count = 0;
+    for (let i = 0, il = xs.length; i < il; ++i) {
+        if (ys.includes(xs[i])) count += 1;
+    }
+    return count;
+}
```
dsehnal, line 121: I am a little bit worried about the ``O(n^2)`` complexity here. It should be fine for the rings as they are small, but would be careful about leaving the method here as it is. Maybe add a comment at least saying it's for small arrays only?

src/mol-repr/structure/visual/util/bond.ts
```diff
@@ -20,6 +20,7 @@ export const BondParams = {
     includeTypes: PD.MultiSelect(ObjectKeys(BondType.Names), PD.objectToOptions(BondType.Names)),
     excludeTypes: PD.MultiSelect([] as BondType.Names[], PD.objectToOptions(BondType.Names)),
     ignoreHydrogens: PD.Boolean(false),
+    aromaticBonds: PD.Boolean(false, { description: 'Display aromatic bonds with dashes' }),
```
dsehnal, line 23: Why false by default?

dsehnal (approved):
Looks good, nothing seems to pop out.

src/mol-repr/structure/visual/util/bond.ts
arose: 1) Don't want to change the current behavior. 2) Not sure what people want to see more, the kekulized form or the aromatics.

src/mol-util/array.ts
arose: Good point, will add a note

PR #222, opened 2021-07-04
Backbone representation

arose:
image


src/mol-math/linear-algebra/3d/vec3.ts
```diff
@@ -474,7 +474,9 @@ namespace Vec3 {
 
     /** Computes the angle between 2 vectors, reports in radians. */
     export function angle(a: Vec3, b: Vec3) {
-        const theta = dot(a, b) / Math.sqrt(squaredMagnitude(a) * squaredMagnitude(b));
+        const denominator = Math.sqrt(squaredMagnitude(a) * squaredMagnitude(b));
+        if (denominator === 0) return Math.PI / 2;
```
dsehnal, line 478: Would ``denominator <= epsilon`` produce more stable results?

dsehnal (commented):
Nice and definitely useful thanks to the ability to use impostors to render the backbone.

Would it also make sense to support spline interpolation so that it would look like cartoons without secondary structures? Or, alternatively, add "ignore secondary structure/backbone only" option for cartoon representation as well?

src/mol-math/linear-algebra/3d/vec3.ts
arose: maybe, the current way is like three.js is doing it

arose:
> 
> 
> Nice and definitely useful thanks to the ability to use impostors to render the backbone.
> 
> Would it also make sense to support spline interpolation so that it would look like cartoons without secondary structures? Or, alternatively, add "ignore secondary structure/backbone only" option for cartoon representation as well?

Isn't that what the putty representation with a uniform size theme does?

image

dsehnal:
> 
> 
> > Would it also make sense to support spline interpolation so that it would look like cartoons without secondary structures? Or, alternatively, add "ignore secondary structure/backbone only" option for cartoon representation as well?
> 
> Isn't that what the putty representation with a uniform size theme does?
> 
> image

Right, I was thinking I was missing something.

src/mol-math/linear-algebra/3d/vec3.ts
dsehnal: Ok I guess we could leave it as it is.

PR #224, opened 2021-07-10
USDZ export

sukolsak:
This PR adds support for USDZ export. [USD/USDZ](https://graphics.pixar.com/usd/docs/index.html) is a 3D file format developed by Pixar. It's [heavily](https://developer.apple.com/augmented-reality/quick-look/) [used](https://developer.apple.com/augmented-reality/tools/) on Apple's platforms. Users can view USDZ files natively on iOS and project them into the real world through the built-in augmented reality viewer. For macOS, they can be viewed through Xcode.

Notes:
* **AR**: On iOS, it's possible to generate a USDZ blob link that [takes the user directly to the AR viewer](https://webkit.org/blog/8421/viewing-augmented-reality-assets-in-safari-for-ios/) instead of downloading it. Android also has a [similar feature](https://developers.google.com/ar/develop/java/scene-viewer) for GLTF, but unfortunately, the GLTF file must be served from the server, [not dynamically generated](https://github.com/google/model-viewer/issues/617). Let me know if you are interested in this feature, and I can implement it. It will work only on iOS until Android supports GLTF blob links. You can see a [demo](https://stanford.edu/~sukolsak/files/molstar_ar.mov) here.
* **Vertex colors**: Vertex coloring in USD seems to be supported only on macOS, but [not on iOS](https://github.com/kcoley/gltf2usd/issues/108#issuecomment-440171377). Apple [recommends](https://download.developer.apple.com/Applications/Reality_Converter/Reality_Converter_Beta_Readme.pdf) that developers "bake" vertex colors to textures when converting from GLTF to USDZ. I'm not sure how to do that with "volume" color types. Since the output files will probably be viewed mainly on iOS, I implement colors using materials. To avoid creating too many materials when using smooth coloring, I use color quantization like in OBJ export.
* **ASCII/Binary**: USD has two formats: ASCII (.usda) and binary (.usdc). USDZ is a zero compression zip archive of .usda or .usdc. There is currently no published spec for the binary format. Three.js is [using the ASCII format](https://github.com/mrdoob/three.js/blob/dev/examples/jsm/exporters/USDZExporter.js). Blender [supports both formats](https://github.com/blender/blender/tree/master/source/blender/io/usd) through [Pixar's USD library](https://github.com/PixarAnimationStudios/USD), which is cumbersome to use. There is an [open-source project](https://github.com/robmcrosby/BlenderUSDZ) that generates the binary directly without using Pixar's library. It seems that the author ported the relevant C++ code from Pixar's library to Python. I use the ASCII format in this PR. The export file size is larger than the binary format, but I think the code is easier to understand and maintain.
* **Instancing**: USD supports [instancing](https://graphics.pixar.com/usd/docs/api/_usd__page__scenegraph_instancing.html). But it seems to have problems with `GeomSubset`s, which I use to bind meshes to multiple materials. The instanced objects seem to lose colors. I don't know how to fix this yet. So I'm currently not using this feature.


arose:
Thanks @sukolsak! The code looks good. I would like to test the usdz output but too. I tried this Blender extension (https://github.com/robmcrosby/BlenderUSDZ) but nothing gets imported. Not sure why. Haven't tried it with other usdz files yet. Also should have access to a mac in the next days to test.

sukolsak:
Thanks @arose. That Blender extension can only import the binary format (i.e., .usdc aka "crate"). When extracting .usdz, it only looks for .usdc files. See https://github.com/robmcrosby/BlenderUSDZ/blob/master/io_scene_usdz/import_usdz.py#L29 This PR uses the ASCII format (i.e., .usda). The extension doesn't work with USDZ exported from [Three.js Editor](https://threejs.org/editor/) either because it also uses the ASCII format.

arose:
Hi @sukolsak, I could try it via BrowserStack on an iPad and built-in Preview App on MacOS can view them as well. Works very nicely. Thanks!

> On iOS, it's possible to generate a USDZ blob link [...]. Let me know if you are interested in this feature, and I can implement it.

That would be great. You have that already on https://stanford.edu/~sukolsak/molstar/ with some feature detection, right?

> Vertex colors

I don't think there is a general way we can "bake" vertex colors to a texture because we would need a way to create uv coordinates for molecular and gaussian surfaces. Using the same approach as for obj is fine.

> ASCII/Binary

Sounds good.

> Instancing

Ok, if you ever figure out how to fix it, feel free to add it :)

sukolsak:
@arose Thanks for trying it. I have implemented the USDZ AR feature. I just wasn't sure how you felt about iOS-only features. Here is what it looks like on an iPhone, with a new AR icon.
molstar

arose:
Nice, thanks @sukolsak

PR #242, opened 2021-08-02
Add overpaint support to geometry exporters

### LINKED ISSUE #241: Add overpaint support to geometry exporters
It would be nice to add overpaint support to geometry exporters. The overpaint data similar to the transparency data already used in the exporters.

sukolsak:
Resolves #241


dsehnal (approved):
Looks good, thanks!

PR #248, opened 2021-08-09
Do not cache LevelTexturesFramebuffers as they may become invalid

MadCatX:
Texture framebuffers use in histogram pyramid calculation are cached in a global variable that exists throughout the entire lifetime of the application. These cached objects may become invalid under some circumstances. One example could be a JS application that creates and destroys a Molstar plugin on demand. Framebuffer objects created by the plugin will not be valid when the plugin's instance gets destroyed and recreated at a later time. Density maps will then fail to display as a result and the only way out is to reload the entire page in the browser.

This could possibly be improved by flushing the cache only when it's actually necessary but I couldn't find a place in the code where doing so wouldn't feel really hacky.

### DIFF
--- a/src/mol-gl/compute/histogram-pyramid/reduction.ts
+++ b/src/mol-gl/compute/histogram-pyramid/reduction.ts
@@ -64,20 +64,15 @@ function createHistopyramidReductionRenderable(ctx: WebGLContext, inputLevel: Te
 }
 
 type TextureFramebuffer = { texture: Texture, framebuffer: Framebuffer }
-const LevelTexturesFramebuffers: TextureFramebuffer[] = [];
 function getLevelTextureFramebuffer(ctx: WebGLContext, level: number) {
-    let textureFramebuffer = LevelTexturesFramebuffers[level];
     const size = Math.pow(2, level);
-    if (textureFramebuffer === undefined) {
-        const texture = ctx.isWebGL2
-            ? getTexture(`level${level}`, ctx, 'image-int32', 'alpha', 'int', 'nearest')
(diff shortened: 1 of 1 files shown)


dsehnal (commented):
This is a nice find, thanks!

Can you please add 

```ts
function tryGetFramebuffer(name: string, webgl: WebGLContext): Framebuffer | undefined {
    const _name = `${HistogramPyramidName}-${name}`;
    return webgl.namedFramebuffers[_name];
}

...
const name = `level${level}`;
let textureFramebuffer = tryGetFramebuffer(name, webgl);
if (!textureFramebuffer) { 
    ...
}
...
```

before we merge this?

MadCatX:
> This is a nice find, thanks!
> 
> Can you please add
> 
> ```ts
> function tryGetFramebuffer(name: string, webgl: WebGLContext): Framebuffer | undefined {
>     const _name = `${HistogramPyramidName}-${name}`;
>     return webgl.namedFramebuffers[_name];
> }
> 
> ...
> const name = `level${level}`;
> let textureFramebuffer = tryGetFramebuffer(name, webgl);
> if (!textureFramebuffer) { 
>     ...
> }
> ...
> ```
> 
> before we merge this?

Sure, but I'm afraid I don't understand the intent behind a `tryGetFramebuffer` function. `getFramebuffer` already either creates a new framebuffer or returns an existing one if it's available.

dsehnal:
To skip the ``attachFramebuffer`` call.

MadCatX:
> To skip the `attachFramebuffer` call.

Oh, right. Is the code in the second commit what you had in mind?

dsehnal:
Yeah that's good, thanks!

PR #252, opened 2021-08-11
binary model loading support and latest mycoplasma model.

corredD:


src/extensions/cellpack/representation.ts
```diff
@@ -0,0 +1,84 @@
...
+
+import { ShapeRepresentation } from '../../mol-repr/shape/representation';
+import { Shape } from '../../mol-model/shape';
+import { ColorNames } from '../../mol-util/color/names';
+import { RuntimeContext } from '../../mol-task';
+import { ParamDefinition as PD } from '../../mol-util/param-definition';
+import { Mesh } from '../../mol-geo/geometry/mesh/mesh';
+import { MeshBuilder } from '../../mol-geo/geometry/mesh/mesh-builder';
+import { Polyhedron, DefaultPolyhedronProps } from '../../mol-geo/primitive/polyhedron';
+import { Icosahedron } from '../../mol-geo/primitive/Icosahedron';
```
arose, line 15: Icosahedron should be lower-case

arose:
Always so cool to see these large models :)

CELLPACK_MEMBRANE

arose:
I have an idea on how to speed up the marking. Currently it takes 60s to highlight the membrane on my machine, which is a bit long :)

arose:
I can look at the PR some more later this week. Let me know if you are still making changes.

src/extensions/cellpack/state.ts
```diff
@@ -26,23 +30,165 @@ const ParseCellPack = PluginStateTransform.BuiltIn({
...
+                let ctr_pos = new Float32Array();
+                // let ctr_norm = new Float32Array();
+                let ctr_info = new Float32Array();
+                let curve_ids = new Float32Array();
+                // let ctrl_pts= new Float32Array();
+                // let ctrl_normal= new Float32Array();
+                // let ctrl_info= new Float32Array();//#curve_id, curve_type, angle, uLength
+                let offset = 12;
+                if (ninst !== 0){
+                    pos = new Float32Array(model_data.buffer, offset, ninst * 4);offset += ninst * 4 * 4;
```
dsehnal, line 123: What endianess are your data? There are ``isLittleEndian()``/``IsNativeEndianLittle`` and ``flipByteOrder`` functions in Mol* that you can use to make sure the data are loaded correctly.

dsehnal (commented):
Very cool!

src/extensions/cellpack/state.ts
dsehnal: The other way around (you want to convert into the native endian). Something like this (I hope :)):

```ts
if (!IsNativeEndianLittle) {
    // only flip the byte order on the current view
    const buffer = flipByteOrder(new Uint8Array(data.buffer, offset, 4 * size), 4);
    return new Float32Array(buffer);
} else {
    return new Float32Array(data.buffer, offset, size);
}
```

corredD: great thanks, I made the change.

arose (changes requested):
Very nice @corredD. Thanks for up-streaming your work on this! I left a few comments inline.

Might be worth to create separate Structure objects for the large fibers so they can easily be switched on/off and maybe in the future optionally replaced by a simpler geometry like for the membrane.

There is a double slash in some URLs which leads to a lot of extra redirects. Probably worth getting rid of when loading so many files.

image

src/extensions/cellpack/representation.ts
```diff
@@ -0,0 +1,84 @@
...
+export type UnitcellProps = PD.Values<MBParams>
+
+function getMBMesh(data: MembraneSphereData, props: UnitcellProps, mesh?: Mesh) {
+    const state = MeshBuilder.createState(256, 128, mesh);
+    const radius = props.radius;
+    let p = DefaultPolyhedronProps;
+    p.detail = 3;
+    p.radius = radius;
+    const { vertices, indices } = Icosahedron();
+    const asphere = Polyhedron(vertices, indices, p);
```
arose, line 54: The usual way is to get a unit sphere with `Sphere()` and then scale it when adding it.

src/extensions/cellpack/model.ts
```diff
@@ -477,30 +506,57 @@ async function loadMembrane(plugin: PluginContext, name: string, state: State, p
...
-        representation: params.preset.representation,
+    const props = {
+        type: {
+            name: 'assembly' as const,
+            params: { id: '1' }
+        }
     };
-
-    await CellpackMembranePreset.apply(membrane, membraneParams, plugin);
+    if (params.source.name === 'id' && params.source.params !== 'MycoplasmaGenitalium.json'){
```
arose, line 515: Not too happy that there is a specific filename buried here. The info what kind of membrane this is should be in the packings data.

src/extensions/cellpack/model.ts
```diff
@@ -477,30 +506,57 @@ async function loadMembrane(plugin: PluginContext, name: string, state: State, p
...
     let cellPackJson: StateBuilder.To<PSO.Format.Json, StateTransformer<PSO.Data.String, PSO.Format.Json>>;
+    let modelFile: Asset.File|null = params.results;
     if (params.source.name === 'id') {
         const url = Asset.getUrlAsset(plugin.managers.asset, getCellPackModelUrl(params.source.params, params.baseUrl));
+        // console.log("getting "+params.source.params+" "+url.url);
         cellPackJson = state.build().toRoot()
             .apply(StateTransforms.Data.Download, { url, isBinary: false, label: params.source.params }, { state: { isGhost: true } });
+
+        if (params.source.params === 'MycoplasmaGenitalium.json'){
+            const m_url = Asset.getUrlAsset(plugin.managers.asset, `${params.baseUrl}/results/results_149_curated_serialized.bin`);
```
arose, line 551: How about putting the URL to the .bin file into the .json file and resolve it in `StructureFromCellpack` instead of passing it as a parameter?

The .bin file should retrieved using the `plugin.managers.asset`
`const asset = await plugin.runTask(plugin.managers.asset.resolve(Asset.getUrlAsset(assetManager, url), 'binary'));`. The resulting `asset` needs to be added to the `assets` variable in `StructureFromCellpack.apply`. That way the bin file will be part of the .molx session file.

arose (changes requested):

arose:
> I have an idea on how to speed up the marking. Currently it takes 60s to highlight the membrane on my machine, which is a bit long :)

Sped it up by inline the loci size calculation code, see 2482ef92af306979102d734119c760fbbb8339d0. It is about 4-5 times faster for me than before in Chrome. Performance analysis shows much more even time-spent-in breakdown (`diff` is related to the loci size calc; `mark` I already optimized as much as I could before; `texImage2D` is a built-in).

image

src/extensions/cellpack/model.ts
corredD: It was temporary, the new push should resolve this issue.

src/extensions/cellpack/model.ts
corredD: I moved the code in the model.ts file. However I am not sure how to add the asset to the asset manager 
`        if (options.resultfile) {
            const asset = await plugin.runTask(plugin.managers.asset.resolve(Asset.getUrlAsset(plugin.managers.asset, `${params.baseUrl}/results/${options.resultfile}`), 'binary', true));
            modelFile = Asset.File(new File([asset.data], 'model.bin'));
        }`

arose: Getting the asset using `plugin.managers.asset.resolve` already adds it. That part is good and it will be part of the .molx session. What is missing is the ensuring that the asset gets disposed of when the transform is deleted from the state tree, see https://github.com/corredD/molstar-proto/blob/393fc99ed2f717f844bb738b269a63f2910c6870/src/extensions/cellpack/state.ts#L132-L140.

They way you are doing the bin packing now does not yet work with sessions. Currently you are doing all the bin packing loading as part of an the LoadCellPackModel Action which means the bin packing data is not added to the state and thus not saved in the session. You can try to add the bin packing parsing as part of the ParseCellPack Transform.

src/extensions/cellpack/model.ts
```diff
@@ -707,6 +625,20 @@ async function loadPackings(plugin: PluginContext, runtime: RuntimeContext, stat
                         await loadMembrane(plugin, packings[i].name, state, params);
                     }
                 }
+            } else if (params.membrane === 'spheres'){
+                if (packings[i].mb) {
+                    let nSpheres =  packings[i].mb!.positions!.length / 3;
+                    console.log('ok mb ', nSpheres);
+                    for (let j = 0;j < nSpheres;j++) {
+                        await state.build()
```
dsehnal, line 633: - How many spheres would be here? It would be much more efficient to have a ``CreateSpheres`` transform which would take a list of vectors (Just using ``PD.Value<Vec3[]>``).
- Also, when applying multiple transforms, it is much more efficient to do:

```ts
const update  = state.build().toRoot();
for (let j = 0;j < nSpheres;j++) {
  update.apply(...);
}
await update.commit();
```

src/extensions/cellpack/state.ts
```diff
@@ -95,9 +100,101 @@ const ParseCellPack = PluginStateTransform.BuiltIn({
                     }
                 }
             }
+            const { options } = cell;
+            if (!params.modelFile && options) {
+                if (options.resultfile) {
+                    const asset = await plugin.runTask(plugin.managers.asset.resolve(Asset.getUrlAsset(plugin.managers.asset, `${params.baseUrl}/results/${options.resultfile}`), 'binary', true));
```
dsehnal, line 106: You want to be able to reuse this asset in multiple transforms from the parent ``Cell``?

src/extensions/cellpack/data.ts
```diff
@@ -35,6 +44,15 @@ export interface Recipe {
 export interface Compartment {
     surface?: Packing
     interior?: Packing
+    geom?: string
+    geom_type?: string
+    mb?: Membrane
+}
+
+export interface Membrane{
+    positions?: number[];
```
arose, line 53: why are positions and radii optional?

src/extensions/cellpack/representation.ts
```diff
@@ -0,0 +1,86 @@
...
+
+interface MembraneSphereData {
+    radius: number
+    center: Vec3
+}
+
+
+const MembraneSphereParams = {
+    ...Mesh.Params,
+    cellColor: PD.Color(ColorNames.orange),
```
arose, line 30: membraneColor?

src/extensions/cellpack/representation.ts
```diff
@@ -0,0 +1,86 @@
...
+    cellScale: PD.Numeric(2, { min: 0.1, max: 5, step: 0.1 }),
```
arose, line 31: cellScale is probably not needed

src/extensions/cellpack/representation.ts
```diff
@@ -0,0 +1,86 @@
...
+    radius: PD.Numeric(2, { min: 0.1, max: 5, step: 0.1 }),
```
arose, line 32: you could give two radii or a thickness since this will always be to represent a lipid bilayer, right?

src/extensions/cellpack/representation.ts
```diff
@@ -0,0 +1,86 @@
...
+    return m;
+}
+
+function getMBShape(ctx: RuntimeContext, data: MembraneSphereData, props: UnitcellProps, shape?: Shape<Mesh>) {
+    const geo = getMBMesh(data, props, shape && shape.geometry);
+    const label = 'mb';
+    return Shape.create(label, data, geo, () => props.cellColor, () => 1, () => label);
+}
+
+//
```
arose, line 72: please remove unused code

src/extensions/cellpack/representation.ts
```diff
@@ -0,0 +1,86 @@
...
+export const MBParams = {
+    ...MembraneSphereParams
+};
+export type MBParams = typeof MBParams
+export type UnitcellProps = PD.Values<MBParams>
+
+function getMBMesh(data: MembraneSphereData, props: UnitcellProps, mesh?: Mesh) {
+    const state = MeshBuilder.createState(256, 128, mesh);
+    const radius = props.radius;
+    // const p = DefaultPolyhedronProps;
```
arose, line 52: please remove unused code

src/extensions/cellpack/state.ts
```diff
@@ -148,3 +294,27 @@ const StructureFromAssemblies = PluginStateTransform.BuiltIn({
         b?.data.customPropertyDescriptors.dispose();
     }
 });
+
+const CreateTransformer = StateTransformer.builderFactory('cellPACK');
+export const CreateSphere = CreateTransformer({
```
arose, line 299: Please adjust all the names and labels to include 'membrane' so it is clear what representation this creates, thanks.

src/extensions/cellpack/util.ts
```diff
@@ -74,4 +76,121 @@ export function getStructureMean(structure: Structure) {
...
+    if (negative) {
+        // need to set the most significant 8 bits to 1's since a js
+        // number is 32 bits but our mantissa is only 24.
+        mantissa |= 255 << 24;
+    }
+
+    return mantissa * Math.pow(10, exponent);
+}
+
+/*
```
arose, line 112: remove?

src/extensions/cellpack/util.ts
```diff
@@ -37,7 +38,8 @@ async function downloadPDB(plugin: PluginContext, url: string, id: string, asset
 }
 
 export async function getFromPdb(plugin: PluginContext, pdbId: string, assetManager: AssetManager) {
-    const { cif, asset } = await downloadCif(plugin, `https://models.rcsb.org/${pdbId.toUpperCase()}.bcif`, true, assetManager);
+    // ${pdbId.toUpperCase()}
```
arose, line 41: please remove

arose (changes requested):

src/extensions/cellpack/data.ts
corredD: positions and radii are one way to define a compartment. Its not always the case, and to be more accurate a mesh or a volume is actually used for the compartment.

src/extensions/cellpack/representation.ts
corredD: thats right I will look into it.

corredD:
molstar-image(2)
The xray shading is really nice ! How can I get it by default ?
`
const url = Asset.getUrlAsset(plugin.managers.asset, `${params.baseUrl}/geometries/${name}`);
b = b.apply(StateTransforms.Data.Download, { url, isBinary: false, label: name }, { state: { isGhost: true } });
await b.apply(StateTransforms.Data.ParsePly, undefined, { state: { isGhost: true } })
            .apply(StateTransforms.Model.ShapeFromPly)
            .apply(StateTransforms.Representation.ShapeRepresentation3D, { xrayShaded: true, doubleSided: true, color: ColorNames.orange })
            .commit({ revertOnError: true });
`
the above code only work for the doubleSided property.

arose:
Hmm, not sure what happened there for you. Maybe it wasn't compiled yet? It looks right except for the coloring. The below works for me.

```ts
await b.apply(StateTransforms.Data.ParsePly, undefined, { state: { isGhost: true } })
    .apply(StateTransforms.Model.ShapeFromPly)
    .apply(StateTransforms.Representation.ShapeRepresentation3D, { xrayShaded: true, doubleSided: true, coloring: { name: 'uniform', params: { color: ColorNames.orange } } })
    .commit({ revertOnError: true });
```

corredD:
Thanks! working nicely 
molstar-image(3)

PR #257, opened 2021-08-20
ANVIL: improve prediction for 3pqr

JonStargaryen:
I've had a look at the dubious ANVIL prediction for 3pqr that Alex reported. Seems like lowering the number of sphere points to 140 was a bit too greedy.

175, half of the original value, results in a better prediction in this case while not affecting performance greatly. Calculations for extreme cases like 7cgo or 6nk5 still finish similarly fast.

Left: 140 points, right: 175
3pqr


dsehnal:
Thanks for the update.

PR #258, opened 2021-08-22
Add optional marking pass

arose:
This adds an optional marking pass that outlines visible and hidden parts of highlighted/selected groups. I suggest we leave it disabled by default for now as it has quite the performance cost (needs to render the scene an additional time) and revisit enabling it for the next major release. There are also a few parameters to tweak the appearance. I tried to do the marking pass at half resolution but it does not look good, not crisp.

image

image


arose:
TODO: handle fog by fading out marking outline

src/mol-geo/geometry/marker-data.ts
```diff
@@ -11,18 +11,35 @@ import { TextureImage, createTextureImage } from '../../mol-gl/renderable/util';
...
+    markerStatus: ValueCell<number>
+}
+
+export function getMarkersAverage(array: Uint8Array, count: number): number {
+    if (count === 0) return 0;
+    let sum = 0;
+    for (let i = 0; i < count; ++i) {
+        if (array[i]) sum += 1;
+    }
+    return sum / count;
```
dsehnal, line 24: We are only ever interested in empty / full / has-non-zero element.

Maybe this code would work better on average? (I haven't ran it so there might be some error, but the overall idea should be sound).

```ts
export function getMarkersState(array: Uint8Array, count: number): 0 | 0.5 | 1 {
    // 0 empty
    // 0.5 partially full
    // 1 full

    if (count === 0) return 0;

    let zero = +!array[0];
    for (let i = 1; i < count; ++i) {
        const v = array[i] && 1;
        // (v && zero) || (!v && !zero) .. I hope :)
        if (!(v ^ zero)) return 0.5;
        if (v) zero = 0;
    }
    return zero ? 0 : 1;
}
```

src/mol-canvas3d/passes/draw.ts
```diff
@@ -338,15 +359,15 @@ export class DrawPass {
...
 
-    render(renderer: Renderer, camera: Camera | StereoCamera, scene: Scene, helper: Helper, toDrawingBuffer: boolean, transparentBackground: boolean, postprocessingProps: PostprocessingProps) {
+    render(renderer: Renderer, camera: Camera | StereoCamera, scene: Scene, helper: Helper, toDrawingBuffer: boolean, transparentBackground: boolean, postprocessingProps: PostprocessingProps, markingProps: MarkingProps) {
         renderer.setTransparentBackground(transparentBackground);
         renderer.setDrawingBufferSize(this.colorTarget.getWidth(), this.colorTarget.getHeight());
 
         if (StereoCamera.is(camera)) {
-            this._render(renderer, camera.left, scene, helper, toDrawingBuffer, transparentBackground, postprocessingProps);
-            this._render(renderer, camera.right, scene, helper, toDrawingBuffer, transparentBackground, postprocessingProps);
+            this._render(renderer, camera.left, scene, helper, toDrawingBuffer, transparentBackground, postprocessingProps, markingProps);
```
dsehnal, line 367: Maybe it's time to turn the params of this into an object?

dsehnal (commented):
Very nice!

src/mol-geo/geometry/marker-data.ts
arose: Thanks, nice idea. I added some other code to handle cases where early return would be helpful. With that is seems to be faster to have a branch-less loop to calculate this.

src/mol-canvas3d/passes/draw.ts
arose: at some point...

src/mol-data/int/impl/sorted-array.ts
```diff
@@ -61,6 +62,7 @@ export function areEqual(a: Nums, b: Nums) {
     if (a === b) return true;
     const aSize = a.length;
     if (aSize !== b.length || a[0] !== b[0] || a[aSize - 1] !== b[aSize - 1]) return false;
+    if (isRange(a)) return true;
     for (let i = 0; i < aSize; i++) {
```
dsehnal, line 66: Tiny update since 1st and last elements are being compared above:

```ts
aSize--;
for (let i = 1; i < aSize; i++)
```

src/mol-model/structure/structure/element/loci.ts
```diff
@@ -470,7 +470,10 @@ export namespace Loci {
     }
 
     function getUnitIndices(elements: SortedArray<ElementIndex>, indices: SortedArray<ElementIndex>) {
-        return OrderedSet.ofSortedArray(SortedArray.indicesOf<ElementIndex, UnitIndex>(elements, indices));
+        if (SortedArray.areEqual(elements, indices) && SortedArray.isRange(elements)) {
```
dsehnal, line 473: Make the ``SortedArray.isRange(elements)`` first condition since it's cheaper to evaluate?

src/mol-geo/geometry/marker-data.ts
```diff
@@ -19,7 +21,7 @@ export function getMarkersAverage(array: Uint8Array, count: number): number {
     if (count === 0) return 0;
     let sum = 0;
     for (let i = 0; i < count; ++i) {
-        if (array[i]) sum += 1;
+        sum += +!!array[i];
```
dsehnal, line 24: Intuitively, ``sum += array[i] && 1`` might be slightly more efficient.

dsehnal (approved):
Nice improvements!

PR #266, opened 2021-09-09
ANVIL: add TMDET amino acid classification and reference

JonStargaryen:
- the ANVIL prediction depends on the set of amino acids that are considered 'membrane-favoring', this PR adds an alternative classification from TMDET, the TMDET definition shows better performance for porins & most beta-barrel structures and might be useful to some users as a fallback for flawed predictions
- adds a reference to TMDET which, in general, shows striking similarities to ANVIL, it's only fair to give them some credit as well
- fixes prediction for [6y1z](https://www.rcsb.org/3d-view/6Y1Z/1?preset=membrane) which contains amino acids that are not part of an entity

Thanks for having a look!


arose:
Thanks @JonStargaryen, looks good to me.

PR #270, opened 2021-09-19
Additional measurement controls

arose:
This adds 'orientation' and 'plane' measurement controls to the UI.


dsehnal (commented):
Code looks good.

PR #271, opened 2021-09-19
Picking improvements

arose:
- Fix pickScale not taken into account in line/point shader
- Add pixel-scale & pick-scale GET params to Viewer app
- Fix selecting bonds not adding their atoms in selection manager
- Add ``preferAtoms`` option to SelectLoci/HighlightLoci behaviors
- Make the implicit atoms of bond visuals pickable
    - Add ``preferAtomPixelPadding`` to Canvas3dInteractionHelper
- Add points visual to Line representation

Also looks around the initial picked pixel in a spiral of radius 5 until a hit is found.

https://user-images.githubusercontent.com/272250/134830371-c1106aec-3de1-42ae-b096-69b318f5c4d9.mp4

--

An alternative would be to push responsibility to representation objects by passing the pick position to the Representation.getLoci method.

### DIFF
--- a/src/mol-canvas3d/helper/interaction-events.ts
+++ b/src/mol-canvas3d/helper/interaction-events.ts
@@ -27,7 +27,7 @@ const tmpNorm = Vec3();
 
 export const Canvas3dInteractionHelperParams = {
     maxFps: PD.Numeric(30, { min: 10, max: 60, step: 10 }),
-    preferAtomPixelPadding: PD.Numeric(5, { min: 0, max: 20, step: 1 }, { description: 'Number of extra pixels at which to prefer atoms over bonds.' }),
+    preferAtomPixelPadding: PD.Numeric(3, { min: 0, max: 20, step: 1 }, { description: 'Number of extra pixels at which to prefer atoms over bonds.' }),
 };
 export type Canvas3dInteractionHelperParams = typeof Canvas3dInteractionHelperParams
 export type Canvas3dInteractionHelperProps = PD.Values<Canvas3dInteractionHelperParams>
--- a/src/mol-canvas3d/helper/interaction-events.ts
+++ b/src/mol-canvas3d/helper/interaction-events.ts
@@ -12,8 +12,7 @@ import { RxEventHelper } from '../../mol-util/rx-event-helper';
 import { Vec2, Vec3 } from '../../mol-math/linear-algebra';
 import { Camera } from '../camera';
 import { ParamDefinition as PD } from '../../mol-util/param-definition';
-import { Bond, StructureElement } from '../../mol-model/structure';
-import { OrderedSet } from '../../mol-data/int';
+import { Bond } from '../../mol-model/structure';
 
 type Canvas3D = import('../canvas3d').Canvas3D
 type HoverEvent = import('../canvas3d').Canvas3D.HoverEvent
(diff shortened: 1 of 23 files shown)


src/mol-canvas3d/passes/pick.ts
```diff
@@ -138,6 +142,9 @@ export class PickHelper {
 
             this.setupBuffers();
         }
+
+        // TODO: make spiral size configurable
+        this.spiral = spiral2d(Math.ceil(this.pickScale * 5));
```
dsehnal, line 147: I know the spiral helps for lines, but for bigger objects, e.g. residues in cartoon, it feels a bit weird when the residue lights up when you are 5 pixels from it.

src/mol-canvas3d/helper/interaction-events.ts
```diff
@@ -144,11 +163,34 @@ export class Canvas3dInteractionHelper {
...
+            const pixelSize = this.camera.getPixelSize(tmpPos);
+            let radius = repr.theme.size.size(loci.bonds[0]) * (repr.props.sizeFactor ?? 1);
+            if (repr.props.lineSizeAttenuation === false) {
+                // divide by two to get radius
+                radius *= pixelSize / 2;
+            }
+            radius += this.props.preferAtomPixelPadding * pixelSize;
+            if (Vec3.distance(tmpPos, tmpPosA) < radius) {
+                return { repr, loci: Bond.toFirstStructureElementLoci(loci) };
+            }
```
dsehnal, line 182: I know the bonds are duplicated in the Loci in each direction, but how does this how does this work with just checking the 1st atom of the 1st bond?

dsehnal (commented):
The 5 pixel radius solution for picking bonds doesn't feel very good, mostly because it behaves differently based on how far the bonds is from the camera.

Would it be possible to have a solution based on percentage distance from the bond? E.g., something like this:

1. Project both ends ``A`` and ``B`` of the bond to screen-space.
2. Project the screenspace pick point ``P`` to this vector to get ``t`` in ``P = A + B t``.
3. If ``t < delta``, pick ``A``, if ``t > 1 - delta`` pick ``B``, pick whole bond otherwise.

arose:
> The 5 pixel radius solution for picking bonds doesn't feel very good, mostly because it behaves differently based on how far the bonds is from the camera.

Actually that is on purpose. The idea being that you need help hitting a target especially when it is far away and bonds are much easier to hit than atoms. Also, 5 pixels is just a tiny amount of extra padding anyway. The main picking area comes from the size of the bond visual. If you are not working with small lines/points that extra padding can be set lower as well. Maybe 3 is a better default?

src/mol-canvas3d/helper/interaction-events.ts
arose: Convention, the bond visuals emit bond loci where the first atom belongs to the half of the bond it represents.

src/mol-canvas3d/passes/pick.ts
arose: Hmm, did not feel odd to me when playing around with it. Pixels are so small nowadays. Anyway, I will make it configurable and am happy to make the default smaller than 5. Also consider that currently the pick buffers are rendered at quarter resolution. So there are already a few pixels +/- off target. Maybe `Math.round(this.pickScale * pickPadding)` with `defaultPickPadding = 3`?

src/mol-canvas3d/helper/interaction-events.ts
dsehnal: I see, this is half-bonds, not bonds.

src/mol-canvas3d/passes/pick.ts
dsehnal: Maybe, could be better

dsehnal:
If we are doing pixel based thresholds for the atoms, I think it's ok whether it's 3 or 5 pixels. Maybe it's just unusual change in general as it's been different for so long :)

PR #273, opened 2021-10-03
add basic unit tests for graphics shaders

arose:
- compile shaders using `gl` package
- the gl package is optional
- skip test if gl package not available
- run gh action with xvfb


arose:
Hey @dsehnal, I think this is ready. Any chance you can have a look? Thanks.

dsehnal (approved):
Code looks fine, tests run on windows and are skipped on Mac. So seems to be working as expected.

PR #274, opened 2021-10-06
Update selection history when Loci is removed from selection

MadCatX:
When a Loci is removed from selection, additionsHistory does not seem to get updated properly. This causes a bit of bad user experience with the Measurements functionality because the user cannot change their mind about what to measure unless they deselect everything and start over.

(This looks like way too easy a fix, what am I missing here?)

### DIFF
--- a/src/mol-plugin-state/manager/structure/selection.ts
+++ b/src/mol-plugin-state/manager/structure/selection.ts
@@ -113,7 +113,10 @@ export class StructureSelectionManager extends StatefulPluginComponent<Structure
 
         const sel = entry.selection;
         entry.selection = StructureElement.Loci.subtract(entry.selection, loci);
-        // this.addHistory(loci);
+
+        const historyEntry = this.additionsHistory.find(he => Loci.areEqual(he.loci, loci));
+        if (historyEntry)
+            this.modifyHistory(historyEntry, 'remove');
         this.referenceLoci = loci;
         this.events.loci.remove.next(loci);
         return !StructureElement.Loci.areEqual(sel, entry.selection);


dsehnal:
The measurements are order dependent, that's why it only registers additions and forces the user to "start again" (or at least that was my logic when implementing it). However, I agree that making a mistake on the "current" edit could be somewhat annoying. How about only having the option to "toggle" the latest addition to the selection history?

MadCatX:
I was hoping I'd get a bit more constructive feedback from my users to discuss this further; what I got could be summarized like this:

- The visual mismatch between what looks like selected and what is actually measured is confusing. Other common tools avoid this by not allowing the user deselect things. That makes it obvious that there is no way back if a mistake is made.
- Having a sorted (and reorderable) list of locis selected for measurement is good enough navigate sort out situations where the order of locis is different to what the user wants.
- Because it is nonsensical to just disable deselections, removing locis from measurement list and accepting the fact that the list may sometimes reorder in an "unwanted" way feels less confusing that the current behavior.

I'm sorry for this meta reply with no actionable information, feel free to close this.

I thought about adding a small visual cue (numbers from 1 to 4 as needed) to show the measurement order directly in the 3D view. Is this something you'd consider for merging?

dsehnal:
> I thought about adding a small visual cue (numbers from 1 to 4 as needed) to show the measurement order directly in the 3D view. Is this something you'd consider for merging?

Yeah, adding a label seems like a good idea, but only when the "add measurement" panel is open. What do you think?

MadCatX:
> Yeah, adding a label seems like a good idea, but only when the "add measurement" panel is open. What do you think?

This is exactly what I have in mind. I'll see if I can find some time to do this.

PR #283, opened 2021-11-08
Improve measurements user experience

MadCatX:
As a followup to #274 I propose the following tweaks to behavior of the measurements feature:

* Show the order in which the selected locis will by used for measurements directly in the 3D structure view. The order is displayed only then the Measurements control is in the 'add' action mode.

measurements_order_shown_s

measurements_order_hidden_s

* Use the entire element to highlight a loci from additionsHistory. Currently, only the first button with highlights the loci. When a user tries to move or delete an entry from history, they cannot see which loci they are working with in the view.

### DIFF
--- a/src/mol-plugin-state/manager/structure/measurement.ts
+++ b/src/mol-plugin-state/manager/structure/measurement.ts
@@ -17,10 +17,12 @@ import { ParamDefinition as PD } from '../../../mol-util/param-definition';
 import { MeasurementRepresentationCommonTextParams, LociLabelTextParams } from '../../../mol-repr/shape/loci/common';
 import { LineParams } from '../../../mol-repr/structure/representation/line';
 import { Expression } from '../../../mol-script/language/expression';
+import { Color } from '../../../mol-util/color';
 
 export { StructureMeasurementManager };
 
 export const MeasurementGroupTag = 'measurement-group';
+export const MeasurementOrderLabelTag = 'measurement-order-label';
 
 export type StructureMeasurementCell = StateObjectCell<PluginStateObject.Shape.Representation3D, StateTransform<StateTransformer<PluginStateObject.Molecule.Structure.Selections, PluginStateObject.Shape.Representation3D, any>>>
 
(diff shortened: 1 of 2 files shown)


src/mol-plugin-ui/structure/measurements.tsx
```diff
@@ -63,8 +63,15 @@ export class MeasurementList extends PurePluginUIComponent {
     render() {
         const measurements = this.plugin.managers.structure.measurement.state;
 
+        const labels = [];
+        for (const cell of measurements.labels) {
+            const tags = (cell.obj as any)['tags'] as string[];
+            if (!tags || !tags.includes(MeasurementOrderLabelTag))
+                labels.push(cell);
+        }
```
dsehnal, line 71: I think this filtering should happen when ``structure.measurement.state`` in the manager and not here.

src/mol-plugin-ui/structure/measurements.tsx
```diff
@@ -80,13 +87,32 @@ export class MeasurementControls extends PurePluginUIComponent<{}, { isBusy: boo
...
 
         this.subscribe(this.plugin.behaviors.state.isBusy, v => {
             this.setState({ isBusy: v });
         });
     }
 
+    componentDidUpdate(prevProps: {}, prevState: { isBusy: boolean, action?: 'add' | 'options' }) {
+        if (this.state.action !== prevState.action)
+            this.updateOrderLabels();
+    }
```
dsehnal, line 101: I think this will keep showing the labels if you hide the whole panel by clicking on the header. Could you please handle that case too?

dsehnal (commented):
Looking good overall, could you notes. 

Would you also please consider not including the package-lock file update in this PR? This seems to be caused by NPM6 vs 7 and I am not sure we should do this update here.

src/mol-plugin-state/manager/structure/measurement.ts
```diff
@@ -281,6 +283,39 @@ class StructureMeasurementManager extends StatefulPluginComponent<StructureMeasu
...
+                .apply(StateTransforms.Model.MultiStructureSelectionFromExpression, {
+                    selections: [
+                        { key: 'a', ref: cell.transform.ref, expression: StructureElement.Loci.toExpression(loci) },
+                    ],
+                    isTransitive: true,
+                    label: 'Order'
+                }, { dependsOn, tags: MeasurementOrderLabelTag })
+                .apply(StateTransforms.Representation.StructureSelectionsLabel3D, {
+                    textColor: Color.fromRgb(255, 0, 0),
+                    textSize: 0.33,
```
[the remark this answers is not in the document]
MadCatX: I suppose it does.  The following labels look okay to me.

(Atom vs. residue)
measurements_atom_vs_residue_s

(Three selections of various size)
three_large_selections_s

src/mol-plugin-state/manager/structure/measurement.ts
```diff
@@ -281,6 +283,39 @@ class StructureMeasurementManager extends StatefulPluginComponent<StructureMeasu
...
+            update
+                .apply(StateTransforms.Model.MultiStructureSelectionFromExpression, {
+                    selections: [
+                        { key: 'a', ref: cell.transform.ref, expression: StructureElement.Loci.toExpression(loci) },
+                    ],
+                    isTransitive: true,
+                    label: 'Order'
+                }, { dependsOn, tags: MeasurementOrderLabelTag })
+                .apply(StateTransforms.Representation.StructureSelectionsLabel3D, {
+                    textColor: Color.fromRgb(255, 0, 0),
```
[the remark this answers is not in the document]
MadCatX: Okay, that does look a lot better.

src/mol-plugin-ui/structure/measurements.tsx
MadCatX: True, fixed.

src/mol-plugin-ui/structure/measurements.tsx
MadCatX: Thanks, I didn't think of that. Fixed.

MadCatX:
Thanks a lot for reviewing this so quickly! All of your concerns should be addressed. I'm sorry about the accidental commit if the package-lock, it should be gone now.

dsehnal (approved):
Nice addition, thanks! Looking good and if any issues arise with it, we will fix it as we go.

PR #284, opened 2021-11-08
Move jest types to devDependencies

russellp17:
Moves the @types/jest package from regular the dependencies list to devDependencies. This solves an issue where projects using molstar as a library can have conflicting type definitions between jest and other testing frameworks.


dsehnal:
Thanks for the tweak!

PR #285, opened 2021-11-09
Tweak measurement order labels

MadCatX:
Naturally, I discovered a few minor issues with the newly added measurement labels code. Sorry... :)

* Border width of 0.5 is too wide and may cause visual artifacts. It is reduced to 0.3.
* Default offsetZ value of 2 is most likely okay for ordinary labels but it seems too high for the order labels. If the user zooms in very closely, order labels get clipped off even though the structure itself is still clearly visible. Value of 0.75 seems to improve this a lot without causing the labels to sink into the structure.
* There was a minor glitch in the text rendering code where the border width was initially set at half of the intended value. With the fix in place it might make sense to check if the default border width (0.2?) is not too thick now.

### DIFF
--- a/src/mol-geo/geometry/text/text.ts
+++ b/src/mol-geo/geometry/text/text.ts
@@ -245,7 +245,7 @@ export namespace Text {
             ...BaseGeometry.createValues(props, counts),
             uSizeFactor: ValueCell.create(props.sizeFactor),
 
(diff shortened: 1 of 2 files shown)


dsehnal:
LGTM, thanks.

dsehnal:
And one note for future PRs: please try to remember to update the changelog too (I often forget to update it too).
@MadCatX

MadCatX:
Sure. I'll keep that in mind for the next time. Thanks a lot!

PR #288, opened 2021-11-23
Added the ability to handle v3000 sd files

### LINKED ISSUE #286: V3000 SDF Support
Hello,

Would it be possible to support a V3000 SDF format when loading SDF into the viewer?

Thanks for your time,
Jason

jpattle:
Added a set of utility functions for parsing atoms and bonds from v3000 sd files.
Updated the existing sdf parser to determine the version and run the v3000 sd file parser functions instead of the default v2000 ones.
Added tests to verify parsing functionality for example v3000 ctab

resolves #286

### DIFF
--- a/src/mol-io/reader/sdf/parser.ts
+++ b/src/mol-io/reader/sdf/parser.ts
@@ -11,8 +11,9 @@ import { Task } from '../../../mol-task';
 import { ReaderResult as Result } from '../result';
 import { Tokenizer, TokenBuilder } from '../common/text/tokenizer';
 import { TokenColumnProvider as TokenColumn } from '../common/text/column/token';
+import { handleAtomsV3, handleBondsV3, handleCountsV3, isV3 } from './parser-v3-util';
 
-/** http://c4.cabrillo.edu/404/ctfile.pdf - page 41 */
+/** http://c4.cabrillo.edu/404/ctfile.pdf - page 41 & 79 */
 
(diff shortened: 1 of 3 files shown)


src/mol-io/reader/sdf/parser.ts
```diff
@@ -71,9 +72,17 @@ function handleMolFile(tokenizer: Tokenizer) {
...
 
-    const atomCount = +counts.substr(0, 3), bondCount = +counts.substr(3, 3);
+    let atomCount = NaN;
+    let bondCount = NaN;
+
+    if (molIsV3) {
+        ({ atomCount, bondCount } = handleCountsV3(tokenizer));
+    } else {
+        atomCount = +countsAndVersion.substr(0, 3), bondCount = +countsAndVersion.substr(3, 3);
+    }
```
dsehnal, line 85: This parts feels a bit weird now. Please add ``handleCountsV2`` and do

```ts
const { atomCount, bondCount } = molIsV3 
    ? handleCountsV3(tokenizer)
    : handleCountsV2(countsAndVersion);
```

dsehnal (commented):
This is great, thanks! Just one small comment and we can merge.

src/mol-io/reader/sdf/parser.ts
jpattle: Good point, your suggestion is much cleaner. I have pushed an update.

PR #290, opened 2021-11-25
Smooth Overpaint and Transparency

arose:
image

image


src/extensions/geo-export/mesh-exporter.ts
```diff
@@ -184,11 +210,12 @@ export abstract class MeshExporter<D extends RenderObjectExportData> implements
         }
     }
 
-    protected static getColor(values: BaseValues, groups: Float32Array | Uint8Array, vertexCount: number, instanceIndex: number, isGeoTexture: boolean, interpolatedColors: Uint8Array | undefined, vertexIndex: number): Color {
+    protected static getColor(values: BaseValues, groups: Float32Array | Uint8Array, vertexCount: number, instanceIndex: number, isGeoTexture: boolean, interpolatedColors: Uint8Array | undefined, interpolatedOverpaint: Uint8Array | undefined, vertexIndex: number): Color {
```
dsehnal, line 213: Should we make it a convention in the code that a function that takes more than say 4 args should take an object parameter?

src/mol-geo/geometry/base.ts
```diff
@@ -33,6 +34,40 @@ export const VisualQualityOptions = PD.arrayToOptions(VisualQualityNames);
...
+    }),
+};
+export type ColorSmoothingParams = typeof ColorSmoothingParams
+
+export function hasColorSmoothingProp(props: PD.Values<any>): props is PD.Values<ColorSmoothingParams> {
+    return !!props.smoothColors;
+}
+
+export function getColorSmoothingProps(smoothColors: PD.Values<ColorSmoothingParams>['smoothColors'], preferSmoothing?: boolean, resolution?: number) {
+    if ((smoothColors.name === 'on' || (smoothColors.name === 'auto' && preferSmoothing)) && resolution && resolution < 3) {
```
dsehnal, line 54: What's the purpose of the ``resolution < 3`` check? Maybe make this a ``maxResolution`` option in ``auto`` and ``on``?

dsehnal (commented):
Very nice!

src/extensions/geo-export/mesh-exporter.ts
arose: :) I will consolidate those args

src/mol-geo/geometry/base.ts
arose: The smoothing uses vertex positions as sample points. At higher resolutions (of e.g. coarse gaussian surfaces) there are only a few vertices per color. In that case smoothing isn't helpful as the linear interpolation in the shader is enough. It is a bit of a magic number but until we see a problem with it I think we can keep it.

dsehnal: Okay, makes sense.

PR #300, opened 2021-12-05
Fix USDZ and OBJ export

sukolsak:
The current code throws an error when exporting the default 1TQN as USDZ or OBJ. ("Cannot read properties of undefined (reading 'vertexCount')".) This is because it tries to read the `vertexCount` property of `input.mesh`, which is only defined for GraphicsRenderObjects that share the same vertex and normal arrays for every instance, such as Mesh or TextureMesh. For other kinds of GraphicsRenderObjects, `input.meshes` is defined. The proper way to read `vertexCount` of an instance is to use `MeshExporter.getInstance(input, instanceIndex).vertexCount`, which takes care of `input.mesh` and `input.meshes`.


arose:
Thank you for the fix and the explanation, @sukolsak!

PR #305, opened 2021-12-10
Make cleanup script work with older vesions of Node.js

MadCatX:
The original cleanup script required at least Node v14.14 . The updated version should work with any version of Node that can be expected to build Molstar.


dsehnal (approved):
LGTM, thanks.

PR #309, opened 2021-12-12
ModelArchive / AlphaFold support

arose:
- Add ``PluginConfig.Structure.DefaultRepresentationPreset``
- Add ModelArchive schema extensions (e.g., AlphaFold uses it for the pLDDT score)
- Add ModelArchive option in DownloadStructure action
- Improve support for loading AlphaFold structures
    - Automatic coloring by pLDDT
    - AlphaFold DB option in DownloadStructure action
    - ``afdb`` GET parameter for Viewer app
    - ``Viewer.loadAlphaFoldDb`` method
- Add QualityAssessment extension (using data from ma_qa_metric_local mmcif category)
    - pLDDT & qmean score: coloring, repr presets, molql symbol, loci labels (including avg for multi-residue selections)
    - pLDDT: selection query


arose:
Hi @mandarsd and @JonStargaryen, this is my take on generalizing what you have done in PDBe/RCSB molstar for AlphaFold support. Any suggestion on making this generally useful are most welcome. Best Alex

arose:
FYI @brindakv this includes initial support for the qmean local metric as well, i.e. the https://github.com/ihmwg/MA-dictionary/blob/master/examples/monomer_example.cif example is colored (note that score is only given for three residues in the file).

src/apps/viewer/index.ts
```diff
@@ -409,4 +432,32 @@ export interface LoadTrajectoryParams {
     | { kind: 'coordinates-data', data: string | number[] | ArrayBuffer | Uint8Array, format: BuildInStructureFormat },
     coordinatesLabel?: string,
     preset?: keyof PresetTrajectoryHierarchy
-}
\ No newline at end of file
+}
+
+export const ViewerAutoPreset = StructureRepresentationPresetProvider({
```
dsehnal, line 437: I think this should be part of the extension and not just the viewer.

src/extensions/model-archive/quality-assessment/behavior.ts
```diff
@@ -0,0 +1,212 @@
...
+        if (summary.length) {
+            return summary.join('</br>');
+        }
+    }
+}
+
+//
+
+const confidentPLDDT = StructureSelectionQuery('Confident pLDDT (> 70)', MS.struct.modifier.union([
+    MS.struct.modifier.wholeResidues([
```
dsehnal, line 154: How about something like this (+modify the behaviour that adds this):

```ts
const selectionQueries = 
  [
     ['Confident pLDDT (> 70)', 'gr', 70], 
     ['Low Confidence pLDDT (< 50)', 'lt', 50],
     ...
   ].map(([label, rel, value] => 
      ... 
      (MS.core.rel as any)[rel]([QualityAssessment.symbols.pLDDT.symbol(), value]) 
      ... 
   )
```

src/mol-plugin-state/manager/structure/component.ts
```diff
@@ -82,7 +82,7 @@ class StructureComponentManager extends StatefulPluginComponent<StructureCompone
             if (r.cell.transform.transformer !== StructureRepresentation3D) continue;
 
             const params = r.cell.transform.params as StateTransformer.Params<StructureRepresentation3D>;
-            if (!!params.type.params.ignoreHydrogens !== ignoreHydrogens || params.type.params.quality !== quality || params.type.params.material !== material) {
+            if (!!params.type.params.ignoreHydrogens !== ignoreHydrogens || params.type.params.quality !== quality || !shallowEqual(params.type.params.material, material)) {
```
dsehnal, line 85: Nice find.

src/mol-plugin-state/builder/structure/hierarchy-preset.ts
```diff
@@ -112,7 +114,8 @@ const allModels = TrajectoryHierarchyPresetProvider({
             structures.push(structure);
 
             const quality = structure.obj ? getStructureQuality(structure.obj.data, { elementCountFactor: tr.frameCount }) : 'medium';
-            await builder.representation.applyPreset(structureProperties, params.representationPreset || 'auto', { theme: { globalName: 'model-index' }, quality });
+            const representationPreset = params.representationPreset || plugin.config.get(PluginConfig.Structure.DefaultRepresentationPreset) || PresetStructureRepresentations.auto.id;
```
dsehnal, line 117: `` || PresetStructureRepresentations.auto.id`` is redundant because ``PluginConfig.Structure.DefaultRepresentationPreset`` will return the default value (``auto``) if not set.

src/mol-plugin-state/actions/structure.ts
```diff
@@ -121,6 +129,11 @@ const DownloadStructure = StateAction.build({
             asTrajectory = !!src.params.options.asTrajectory;
             format = 'pdb';
             break;
+        case 'alphafolddb':
+            downloadParams = getDownloadParams(src.params.id, id => `https://alphafold.ebi.ac.uk/files/AF-${id.toUpperCase()}-F1-model_v1.cif`, id => `AlphaFold DB: ${id}`, false);
```
dsehnal, line 133: There are now ``_v2`` files too (and I suppose more will follow). E.g. ``https://alphafold.ebi.ac.uk/entry/Q9UTE2`` => ``https://alphafold.ebi.ac.uk/files/AF-Q9UTE2-F1-model_v2.cif``. 

The URL can be obtained from https://www.alphafold.ebi.ac.uk/api/prediction/Q9UTE2 and they even provide BCIF files!

dsehnal (commented):
Cool... great to see this directly in Mol*!

src/apps/viewer/index.ts
```diff
@@ -251,6 +258,22 @@ export class Viewer {
         }));
     }
 
+    loadAfdb(afdb: string) {
```
dsehnal, line 261: Maybe be explicit here: ``loadAlphaFoldDb``?

src/extensions/model-archive/quality-assessment/behavior.ts
arose: We could but I find it would spam the UI  with lots of entries. I think we should rather add a molql prompt (with pymol, vmd syntax support)

src/apps/viewer/index.ts
arose: Hmm, I added two presets (pLDDT and qmean) to the extension. I thought of this one as specific to the viewer where we can integrate code from multiple extensions.

src/mol-plugin-state/manager/structure/component.ts
arose: :)

src/mol-plugin-state/builder/structure/hierarchy-preset.ts
arose: Right, but this is not type checked at the moment as `PluginConfigItem<T>.defaultValue` can be `undefined`. Should we change that and revisit all the usage of `plugin.config.get`?

src/mol-plugin-state/actions/structure.ts
arose: cool, did not knew that api

done, but the bcif files are broken for me, at least molstar throws an error parsing them, so using cif for now

src/apps/viewer/index.ts
arose: sounds better, done

src/mol-plugin-state/actions/structure.ts
dsehnal: Hmm, should probably investigate what's going on there about the bcif files.

src/mol-plugin-state/builder/structure/hierarchy-preset.ts
dsehnal: Can just add ``!`` since you know that the config variable has a default value set. Tho the extra || is fine too I guess.

src/mol-plugin-state/actions/structure.ts
dsehnal: They will be fixing the BCIF support when people get back from holidays.

src/mol-plugin-ui/index.ts
```diff
@@ -13,16 +13,16 @@ import { DefaultPluginUISpec, PluginUISpec } from './spec';
 
 export function createPlugin(target: HTMLElement, spec?: PluginUISpec): PluginUIContext {
     const ctx = new PluginUIContext(spec || DefaultPluginUISpec());
-    ctx.init();
-    ReactDOM.render(React.createElement(Plugin, { plugin: ctx }), target);
+    ctx.init().then(() => {
+        ReactDOM.render(React.createElement(Plugin, { plugin: ctx }), target);
+    });
```
dsehnal, line 18: This breaks the ``lighting`` example and I suppose it will break some other code as well. I suggest we remove the "sync" ``createPlugin`` in 3.0 altogether.

Issue #313, opened 2021-12-14
Add formal charge label to representation


ptourlas:
Greetings!

Is there a quick and painless way to add a label displaying the formal charge (`M  CHG  1   8  -1`) to the nitrogen at position 8
(` 0.2451   -0.3881   -0.5673 N   5  5  0  0  0  2  0  0  0  0  0  0`)

```
SMILES-CFFFSOON-SOOCFFF
     RDKit          3D

 15 14  0  0  0  0  0  0  0  0999 V2000
    2.4644   -0.6833    0.6783 C   0  0  0  0  0  0  0  0  0  0  0  0
    3.1589   -0.1223   -0.3545 F   0  0  0  0  0  0  0  0  0  0  0  0
    2.4354   -2.0263    0.4264 F   0  0  0  0  0  0  0  0  0  0  0  0
    3.2334   -0.5214    1.7934 F   0  0  0  0  0  0  0  0  0  0  0  0
    0.7862    0.0206    0.8608 S   0  0  0  0  0  6  0  0  0  0  0  0
    0.9929    1.4434    1.1263 O   0  0  0  0  0  0  0  0  0  0  0  0
    0.1705   -0.7131    1.9690 O   0  0  0  0  0  0  0  0  0  0  0  0
    0.2451   -0.3881   -0.5673 N   5  5  0  0  0  2  0  0  0  0  0  0 <--------
   -1.0100    0.4366   -1.0736 S   0  0  0  0  0  6  0  0  0  0  0  0
   -1.6478   -0.2086   -2.2311 O   0  0  0  0  0  0  0  0  0  0  0  0
   -0.7964    1.8705   -1.2847 O   0  0  0  0  0  0  0  0  0  0  0  0
   -2.2304    0.2627    0.2953 C   0  0  0  0  0  0  0  0  0  0  0  0
   -2.3428   -1.0280    0.7238 F   0  0  0  0  0  0  0  0  0  0  0  0
   -1.9714    1.0394    1.3799 F   0  0  0  0  0  0  0  0  0  0  0  0
   -3.4880    0.6179   -0.1175 F   0  0  0  0  0  0  0  0  0  0  0  0
  1  2  1  0
  1  3  1  0
  1  4  1  0
  1  5  1  0
  5  6  2  0
  5  7  2  0
  5  8  1  0
  8  9  1  0
  9 10  2  0
  9 11  2  0
  9 12  1  0
 12 13  1  0
 12 14  1  0
 12 15  1  0
[...]

PR #314, opened 2021-12-14
Feature/formal charge labels

ptourlas:
@dsehnal this is the (very) first step to implementing the feature. My question is: given that we find the line which contains the atom id and the charge, where should we pass it next?

I understand that when
`export async function getMolModels(mol: MolFile, format: ModelFormat<any> | undefined, ctx: RuntimeContext) {
`
is called, the necessary info should already be in the `mol: MolFile`, right?

### DIFF
--- a/src/mol-io/reader/sdf/parser.ts
+++ b/src/mol-io/reader/sdf/parser.ts
@@ -29,12 +29,18 @@ export interface SdfFile {
 
 
 const delimiter = '$$$$';
+const formalChargePrefix = 'M  CHG';
+
 function handleDataItems(tokenizer: Tokenizer): { dataHeader: Column<string>, data: Column<string> } {
     const dataHeader = TokenBuilder.create(tokenizer.data, 32);
     const data = TokenBuilder.create(tokenizer.data, 32);
 
     while (tokenizer.position < tokenizer.length) {
         const line = Tokenizer.readLine(tokenizer);
+        if (line.startsWith(formalChargePrefix)) {
+            console.log('charge found');
+            console.log(line);
+        }
         if (line.startsWith(delimiter)) break;
         if (!line) continue;
 
--- a/src/mol-model-formats/structure/mol.ts
+++ b/src/mol-model-formats/structure/mol.ts
@@ -45,6 +45,7 @@ export async function getMolModels(mol: MolFile, format: ModelFormat<any> | unde
         type_symbol,
 
         pdbx_PDB_model_num: Column.ofConst(1, atoms.count, Column.Schema.int),
+        pdbx_formal_charge: Column.range(-1, 1)
     }, atoms.count);
 
     const entityBuilder = new EntityBuilder();


stefdoerr:
Since chemical file formats are my favorite subject to complain about I'll chime in. There are two places to store formal charge in an SDF file. 
See page 9. It can be stored either in the Atom block or the Properties block.
https://www.daylight.com/meetings/mug05/Kappler/ctfile.pdf

In the atom block the second number after the element can be converted to formal charge. The mapping for this field is:
`chargemap = {"7": -3, "6": -2, "5": -1, "0": 0, "3": 1, "2": 2, "1": 3, "4": 0}`
Thus the 3 is a positive 1 charge on the nitrogen, which is also written in the CHG field.

Different software reads either (or neither) of them to make our life more fun.
As the docs say:
> Generic] Wider range of values in M CHG and M RAD lines below. Retained for compatibility with older Ctabs, M CHG and M RAD lines take precedence

So you will have to read both but overwrite values if M CHG exists

dsehnal:
> @dsehnal this is the (very) first step to implementing the feature. My question is: given that we find the line which contains the atom id and the charge, where should we pass it next?
>
> I understand that when
> `export async function getMolModels(mol: MolFile, format: ModelFormat<any> | undefined, ctx: RuntimeContext) { `
> is called, the necessary info should already be in the `mol: MolFile`, right?

I don't know if the format charge is part of "mol" or "sdf" file spec, but depending on this, add the data to either [SDFFileCompound](https://github.com/molstar/molstar/blob/a736fe7989de73bdce6f28374e808ded144f76fc/src/mol-io/reader/sdf/parser.ts#L18) or [MolFile](https://github.com/molstar/molstar/blob/a736fe7989de73bdce6f28374e808ded144f76fc/src/mol-io/reader/mol/parser.ts#L14). Then it will be "automatically" passed down. If it's in the "SDF file", you might have to modify the signature of ``getMolModels`` and pass the charge data [here](https://github.com/molstar/molstar/blob/a736fe7989de73bdce6f28374e808ded144f76fc/src/mol-model-formats/structure/sdf.ts#L28). 

Also please add yourself to the headers of files you modify (e.g. [here](https://github.com/molstar/molstar/blob/a736fe7989de73bdce6f28374e808ded144f76fc/src/mol-model-formats/structure/sdf.ts#L3)) and as a contributor to ``package.json``.

----

Thanks for the help with this @stefdoerr.

src/mol-io/reader/mol/parser.ts
```diff
@@ -28,6 +40,10 @@ export interface MolFile {
         readonly atomIdxB: Column<number>,
         readonly order: Column<number>
     }
+    readonly formalCharges: {
+        readonly atomIdx: Column<number>;
+        readonly charge: Column<number>;
+    }
 }
 
 export function handleAtoms(tokenizer: Tokenizer, count: number): MolFile['atoms'] {
```
ptourlas, line 49: In `const x = TokenBuilder.create(tokenizer.data, count * 2);`
@dsehnal Why allocate space that is twice the size of the atom count?

dsehnal: Start and end index for each token. E.g. it doesn't do "substr" during the parsing, but just passes down the original string and indices into that string. This allows for faster parsing of numbers down the line for example.

ptourlas:
Greetings!
I have made some new additions and some input would be useful at this point.
The mol file parser now gets the formal charges from both the atom and the properties block.  As @arose suggested:
> 
> You can parse both in `mol-io/reader/mol/parser.ts` and only when you create a `Model` in `mol-model-formats/structure/mol.ts` to decide which to use. Specifically when you add the atom_site.pdbx_formal_charge column.

* For now, I just added the charges from the atom block this way: 
`pdbx_formal_charge: Column.asArrayColumn(atoms.formal_charge)` (see [here](https://github.com/ptourlas/molstar/commit/5c5f8aa7411376e0c1bb1da98372f8de14ebc124#diff-476b99c207510b70c545aa05106b4e5ea67c0ab2dc0d3f28536669ca05643e3d))
Is `Column.asArrayColumn()` the right one for the job?`
* The idea is to replace values from the atom block with values found at the `M CHG * * *` properties. But can we modify a column already built?
* Given that the columns contain merely the indices that point to the values and not the values themselves, another problem arises: The charges in the `M CHG * * *` properties contain the charge, e.g. `M  CHG  1   2  -1` means that atom no2 has a -1 charge (so we should go to the column and replace it). But the charges in the atom block calculated based on the mapping
```
const chargemap = {
    '7': -3,
    '6': -2,
    '5': -1,
    '0': 0,
    '3': 1,
    '2': 2,
    '1': 3,
    '4': 0,
};
```
How should the mapping be used in order to get the actual charge? Perhaps I should create an array that is independent from the data and contains the actual charges?

arose:
You can use Column.ofLambda to 'change' column values
```
indexA: Column.ofLambda({
    value: (row: number) => otherCol.value(row), // do the mapping here
    rowCount: otherCol.rowCount,
    schema: otherCol.schema,
}),
```

dsehnal:
Either use the map as @arose  suggested or just create a new array and populate it with priority based on where the data comes from, and calling ``Column.asArrayColumn()`` on that:

```ts
const charges = new Float64Array(atomCount);
for (const [k, v] of Object.entries(chargeMap)) {
  charges[+k] = v;
} 
// TODO: fill charges not present in M CHG  map from "atom block"
column = Column.asArrayColumn(charges);
```

ptourlas:
The `getMolModels()` function is finalized and I have (hopefully) covered the edge cases as well. This means that the remaining step is getting the charges to be shown in the screen.
I have seen https://github.com/molstar/molstar/issues/313#issuecomment-993757332 about displaying the labels, but I would like some extra hints to integrate this in the UI as well.

arose:
> I have seen #313 (comment) about displaying the labels, but I would like some extra hints to integrate this in the UI as well.

For integrating that into the molstar library I would go with #330

ptourlas:
> For integrating that into the molstar library I would go with #330

@arose I'm afraid I have no clue on how to work on this :sweat_smile:

dsehnal:
> > For integrating that into the molstar library I would go with #330
> 
> @arose I'm afraid I have no clue on how to work on this 😅

We are going to implement this ourselves and you will just be able to use that.

src/mol-model-formats/structure/mol.ts
```diff
@@ -18,13 +19,40 @@ import { IndexPairBonds } from './property/bonds/index-pair';
...
+                    block is treated as zero.
+                */
+                return 0;
+            }
+            const idx = atoms.formal_charge.value(row).toString();
+            return formalChargeMapper(idx);
+        },
+        rowCount: atoms.formal_charge.rowCount,
+        schema: atoms.formal_charge.schema,
+    });
```
dsehnal, line 54: Please change this to something like this:

```ts
const computedFormalCharges = new Int32Array(mol.atoms.count);

for (let i = 0; i < mol.atoms.count; i++) {
    computedFormalCharges[i] = atoms.format_charge.value(i);
}

for (let i = 0; i < formalCharges.atomIdx.rowCount; i++) {
    computedFormalCharges[+formalCharges.atomIdx.value(i)] = +formalCharges.charge.value(i)
}

const pdbx_formal_charge = Column.ofIntArray(computedFormalCharges);
```

src/mol-io/reader/mol/parser.ts
```diff
@@ -20,21 +22,92 @@ export interface MolFile {
...
+    mmm         (ATOM-ATOM MAPPING NUMBER, 60-63)
+    nnn         (INVERSION/RETENTION FLAG, 63-66)
+    eee         (EXACT CHANGE FLAG, 66-69)
+*/
+
+/**
+ * @param key - The value found at the atom block.
+ * @returns The actual formal charge based on the mapping.
+ */
+export function formalChargeMapper(key: string) {
```
dsehnal, line 71: Why does this take string param when all the values are "numeric"?

dsehnal (commented):
Made some suggestions to the code.

- Could you also please add tests for the new charge related features to https://github.com/molstar/molstar/blob/master/src/mol-io/reader/_spec/mol.spec.ts ?

Once this is addressed I think we can merge this PR and solve displaying the format charge information in a separate PR.

src/mol-io/reader/mol/parser.ts
```diff
@@ -84,6 +160,79 @@ export function handleBonds(tokenizer: Tokenizer, count: number): MolFile['bonds
...
+        Tokenizer.trim(tokenizer, s + 3, s + 6);
+        const propertyType = Tokenizer.getTokenString(tokenizer);
+
+        if (propertyType === 'END') break;
+        Tokenizer.eatLine(tokenizer);
+
+        switch (propertyType) {
+            case 'CHG':
+                formalCharges = handleFormalCharges(tokenizer, s);
+                break;
```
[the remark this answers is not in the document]
ptourlas: It's not always one charge per line. Each `M CHG (...)` line can hold up to 8 charges.
The first argument defines the number of charges in the line:
 `M  CHG  N   idx1   chg1   idx2   chg2 (...) idxN   chgN`
If more than 8 charges are present in the molecule, then a new `M CHG (...)` line is added.
([reference](http://www.nonlinear.com/progenesis/sdf-studio/v0.9/faq/sdf-file-format-guidance.aspx))

src/mol-io/reader/mol/parser.ts
ptourlas: Unnecessary indeed. Addressed by https://github.com/molstar/molstar/pull/314/commits/7bc0e9db7c160d6c3990c5f32a270ef4d8c6b11e.

src/mol-io/reader/mol/parser.ts
```diff
@@ -20,21 +22,92 @@ export interface MolFile {
...
+    let chg = 0;
+    switch (key) {
+        case '7':
+            chg = -3;
+            break;
+        case '6':
+            chg = -2;
+            break;
+        case '5':
+            chg = -1;
```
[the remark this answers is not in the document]
ptourlas: Addressed by https://github.com/molstar/molstar/pull/314/commits/7bc0e9db7c160d6c3990c5f32a270ef4d8c6b11e.

src/mol-io/reader/mol/parser.ts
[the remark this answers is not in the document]
dsehnal: Yes, but what he's saying is that you can have either 8 charges per single ``M CHG`` or 8 seprate ``M CHG`` or 2x4, etc. You code only handles a single record.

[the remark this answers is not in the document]
ptourlas: Right. I'm fixing this.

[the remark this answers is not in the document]
ptourlas: Addressed by https://github.com/molstar/molstar/pull/314/commits/dd415bf80216f9efff7f8dcdf46b4772d66ac401.

src/mol-io/reader/mol/parser.ts
```diff
@@ -84,6 +160,79 @@ export function handleBonds(tokenizer: Tokenizer, count: number): MolFile['bonds
...
+/** Call an appropriate handler based on the property type.
+ * (For now it only calls the formal charge handler, additional handlers can
+ * be added for other properties.)
+ */
+export function handlePropertiesBlock(tokenizer: Tokenizer): MolFile['formalCharges'] {
+
+    let formalCharges = null;
+
+    let i = 0;
+    while (i < 50) { // Added a "big" value to avoid any infinite loops by accident.
```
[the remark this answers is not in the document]
ptourlas: Addressed by https://github.com/molstar/molstar/pull/314/commits/23c2dcdfd48e865cf8b88b563dc19aaf3d53f2e3.

src/mol-io/reader/mol/parser.ts
```diff
@@ -84,6 +160,79 @@ export function handleBonds(tokenizer: Tokenizer, count: number): MolFile['bonds
...
+        TokenBuilder.addUnchecked(atomIdx, tokenizer.tokenStart, tokenizer.tokenEnd);
+        Tokenizer.trim(tokenizer, lineStart + offset + 4, lineStart + offset + 8);
+        TokenBuilder.addUnchecked(charge, tokenizer.tokenStart, tokenizer.tokenEnd);
+    }
+
+    /*
+        Once the line is read, move to the next one.
+        Otherwise the cursor will be one position behind on the next line.
+    */
+    Tokenizer.eatLine(tokenizer);
```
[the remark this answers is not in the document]
ptourlas: The truth is those tokens gave me a hard time. :sweat_smile: 
Since it's functional, can we postpone this optimization for a later PR?

src/mol-model-formats/structure/mol.ts
ptourlas: Addressed by https://github.com/molstar/molstar/pull/314/commits/1b431b1d207b5b189fe9d221b7f103b959ab4a68.

src/mol-io/reader/mol/parser.ts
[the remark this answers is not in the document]
dsehnal: Ok, tho to make sure about the functionality, please add the tests.

src/mol-io/reader/sdf/parser.ts
```diff
@@ -95,10 +97,11 @@ function handleMolFile(tokenizer: Tokenizer) {
 
     const atoms = molIsV3 ? handleAtomsV3(tokenizer, atomCount) : handleAtoms(tokenizer, atomCount);
     const bonds = molIsV3 ? handleBondsV3(tokenizer, bondCount) : handleBonds(tokenizer, bondCount);
+    const formalCharges = handleFormalCharges(tokenizer,1)
```
dsehnal, line 100: Please format the documents (e.g. missing comma and ``;`` here) (``npm run lint-fix`` should do the trick).

src/mol-io/reader/mol/parser.ts
```diff
@@ -84,6 +142,77 @@ export function handleBonds(tokenizer: Tokenizer, count: number): MolFile['bonds
...
+        Tokenizer.trim(tokenizer, lineStart + offset, lineStart + offset + 4);
+        const _atomIdx = Tokenizer.getTokenString(tokenizer);
+        formalCharges.atomIdx.push(+_atomIdx);
+        Tokenizer.trim(tokenizer, lineStart + offset + 4, lineStart + offset + 8);
+        const _charge = Tokenizer.getTokenString(tokenizer);
+        formalCharges.charge.push(+_charge);
+    }
+    /* Once the line is read, move to the next one. */
+    Tokenizer.eatLine(tokenizer);
+    return formalCharges;
```
dsehnal, line 178: No longer need to return value and can remove the ``:any`` in function declaration.

src/mol-model-formats/structure/mol.ts
```diff
@@ -18,13 +19,24 @@ import { IndexPairBonds } from './property/bonds/index-pair';
...
 
     const MOL = Column.ofConst('MOL', mol.atoms.count, Column.Schema.str);
     const A = Column.ofConst('A', mol.atoms.count, Column.Schema.str);
     const type_symbol = Column.asArrayColumn(atoms.type_symbol);
     const seq_id = Column.ofConst(1, atoms.count, Column.Schema.int);
 
+    const computedFormalCharges = new Int32Array(mol.atoms.count);
+    if (formalCharges.atomIdx.rowCount > 0) {
+        for (let i = 0; i < formalCharges.atomIdx.rowCount; i++) {
+            computedFormalCharges[formalCharges.atomIdx.value(i) - 1] = formalCharges.charge.value(i);
```
dsehnal, line 32: Could the formal charge assignment be composed of both ``M CHG`` and "inline" values from ``atoms.formal_charge``?

src/mol-io/reader/mol/parser.ts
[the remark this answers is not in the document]
ptourlas: No worries, they will be part of this PR.

src/mol-model-formats/structure/mol.ts
ptourlas: No, the specification dictates that if there's a `M CHG` line then charges from the atom block are treated as zeros.
Since the constructor already sets all the array elements to zero, I skipped the part of doing it explicitly.

src/mol-io/reader/mol/parser.ts
ptourlas: Oops! Missed those, fixed by https://github.com/molstar/molstar/pull/314/commits/bcec1d96379de5d32e8ecd34211ec2d8885d9281.

ptourlas:
@dsehnal I believe everything is ready for the most part. One last remark is that I did not find any tests for the [`getMolModels`](https://github.com/molstar/molstar/blob/9bd60f8e8e65c18872d27ac82b003a7c438e7726/src/mol-model-formats/structure/mol.ts#L20) function. Based on its signature, which also requires a model format and a runtime context, I guess that it would be quite trickier to test?

src/mol-io/reader/_spec/mol.spec.ts
```diff
@@ -49,6 +49,33 @@ const MolString = `2244
...
+
+  4  3  0  0  0  0            999 V2000
+    0.0000    0.0000    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
+    0.8660    0.5000    0.0000 N   0  0  0  0  0  0  0  0  0  0  0  0
+   -0.8660    0.5000    0.0000 S   0  0  0  0  0  0  0  0  0  0  0  0
+    0.0000   -1.0000    0.0000 P   0  0  0  0  0  0  0  0  0  0  0  0
+  1  4  2  0  0  0  0
+  3  1  1  0  0  0  0
+  2  1  1  0  0  0  0
+M  CHG  3   2  -1   3   1   4   1
```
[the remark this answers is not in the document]
ptourlas: Addressed by https://github.com/molstar/molstar/pull/314/commits/1bd4d841a1db147ad757be46a9964beb89c834d8.

ptourlas:
> Looks good.
> 
> If you want to write a test for `getMolModels`, you can pass `RuntimeContext.Synchronous` + `MolFormat.create(molFile)` to it.

I shall add it to my backlog, it's not a high priority at the moment.

ptourlas:
@dsehnal @arose I believe we are ready for takeoff with this PR. :rocket:

dsehnal (approved):
LGTM, thanks for the contribution.

PR #316, opened 2021-12-16
update State Snapshots canApply when snapshots are added

JonStargaryen:
Under Export Animation, when exporting State Snapshots, the canApply values won't get updated when additional state snapshots are added (or removed). Will keep reporting 'At least 2 states required.' and requires e.g. switching temporarily to another animation type.

state-snapshots


dsehnal (approved):
Thanks for the update.

PR #320, opened 2021-12-19
Shader compilation

arose:
Reduce number of created programs/shaders
- Support specifying variants when creating graphics render-items
- Change double-side shader param from define to uniform
- Remove dMarkerType shader define (use uMarker as needed)
- Support to ignore defines depending on the shader variant
- Combine pickObject/pickInstance/pickGroup shader variants into one
- Combine markingDepth/markingMask shader variants into one
- Correctly set shader define flags for overpaint, transparency, substance, clipping


src/mol-canvas3d/passes/postprocessing.ts
```diff
@@ -226,7 +226,7 @@ function getPostprocessingRenderable(ctx: WebGLContext, colorTexture: Texture, d
 
         uMaxPossibleViewZDiff: ValueCell.create(0.5),
 
-        dOcclusionEnable: ValueCell.create(false),
+        dOcclusionEnable: ValueCell.create(true),
```
dsehnal, line 229: Why this change?

dsehnal (commented):
Code looks good and it things seem to work after some basic testing.

src/mol-canvas3d/passes/postprocessing.ts
arose: That is just to bring it in line with the default params of the pass so the shader does not get re-created right after initialization. Btter, of course, would be to support passing the params but did not get to that.

PR #321, opened 2021-12-19
mol-plugin-ui/createPluginUI

dsehnal:
- There have been two ways to create a plugin instance and both were a little bit broken. 
- Taking the opportunity of 3.0 to improve on the situation by introducing ``createPluginUI`` and removing the previous options.


arose (approved):
lgtm

PR #323, opened 2021-12-23
Direct-volume refactor

arose:
- remove isosurface mode (use gpu mc)
- move coloring to theme


dsehnal (approved):
LGTM, nothing really pops out in the code and worked when I tested it.

PR #325, opened 2021-12-27
Integrate pymol/vmd transpilers from MolQL project

ptourlas:
Just adding the source code here, will soon work on the depedencies that are changed.


ptourlas:
Please note that I added the comment `/* FAULTY IMPORTS */` to the imports that were available at the MolQL repo but not here. How should we handle these?

arose:
> Please note that I added the comment /* FAULTY IMPORTS */ to the imports that were available at the MolQL repo but not here. How should we handle these?

- `../types.ts`, `../helper.ts`, ``../transpiler.ts`` you can add to `/mol-script/transpilers/`
- `Expression` you can import from `/mol-script/language/expression.ts`

Could you please also add the relevant (pymol, vmd) spec tests from `molql/src/transpilers/spec`? We use jest in molstar so they need a bit of adjustment (_spec folder name, no need to import jasmine, ...), see `src/mol-util/_spec/zip.spec.ts` for an example.

ptourlas:
- [x] `../types.ts`, `../helper.ts`, `../transpiler.ts` you can add to `/mol-script/transpilers/`
- [x] `Expression` you can import from `/mol-script/language/expression.ts`
> 
> 
> Could you please also add the relevant (pymol, vmd) spec tests from `molql/src/transpilers/spec`? We use jest in molstar so they need a bit of adjustment (_spec folder name, no need to import jasmine, ...), see `src/mol-util/_spec/zip.spec.ts` for an example.

Added the spec tests and the relevant `utils.ts` file. However, there is a `compile()` function which is imported and requires substitution as well. With what though :monocle_face: ?

ptourlas:
Additionally, @arose @dsehnal what is required to make a functional example of this feature?
Even something very rough and unpolished (like providing a string to a function which is attached to the `window` object and have the selection expression returned). I just want to see it in action to wrap my head around it, even if not finalized

dsehnal:
Use the [StructureSelectionFromScript](https://github.com/molstar/molstar/blob/28678e2f8019e3aced03e5d6eb3d9c39cd1f2738/src/mol-plugin-state/transforms/model.ts#L708) transform and modify [Script.toExpression](https://github.com/molstar/molstar/blob/28678e2f8019e3aced03e5d6eb3d9c39cd1f2738/src/mol-script/script.ts#L33) to be able to parse the languages you add.

You can do this from code, but eventually an UI control(s) should be added too to support the additional languages.

arose:
> Additionally, @arose @dsehnal what is required to make a functional example of this feature?

Writing specs with the functions David noted above would be good. 

If you want to quickly add something to the UI, maybe do it like the "Atom/Residue Identifier List" https://github.com/molstar/molstar/blob/master/src/mol-plugin-ui/structure/selection.tsx#L252-L259

image

image

ptourlas:
@arose @dsehnal Greetings!
Is there any update on this PR?

arose:
@ptourlas There are quite a few errors since the MonadicParser in molstar is not exactly the same as the one in molql. I can have a look but not sure when yet.

arose:
> However, there is a compile() function which is imported and requires substitution as well. With what though 🧐 ?

`import { compile } from '../../runtime/query/compiler';`

ptourlas:
> @ptourlas There are quite a few errors since the MonadicParser in molstar is not exactly the same as the one in molql. I can have a look but not sure when yet.

@arose thank you for your response, if there is anything I can do to help with this, I'd be glad to.

yakomaxa:
Greetings,

 I have been interested in this PR and recently managed to enable PyMOL language in Molstar using MolQL as you have discussed in this PR and this issue at MolQL repo.(https://github.com/MolQL/molql/issues/4).  

As my implementation disrupted (replaced) mol-script functionality and no spec test has been done, I don't think I should open PR. However, the compilation succeeded, and the PyMOL scripting seems to work fine in the scripting window. 

Since my implementation is mostly based on your previous contributions, I'd like to notify you that it was successful and thank you. 

Though the code is not clean enough, as I'm a beginner in typescript, I implemented some missing functions as you discussed in this PR.  Please check the repo if you got interested. 

https://github.com/yakomaxa/molstar_my

Sincerely.

arose:
@yakomaxa great that you are interested and are making progress on this. It would be very welcome to integrate it into the main molstar repository. You are right that we would not want to replace mol-script functionallity but that can be worked around by adding a language selector to the relevant parameters. There can also be a global parameter for a default. Please feel free to start a PR as that simplifies discussing any changes.

yakomaxa:
@arose Thank you for your response. Now most of molql transpiler behaviors are reproduced, I will open a PR to main repository after cleaning the codes.

arose:
closing in favor of https://github.com/molstar/molstar/pull/514

PR #326, opened 2021-12-29
Marker improvements

arose:
- various marker improvements
- enable marking pass by default (a bit expensive for cellpack models but would like to test this more widely)
- improve temporal multi-sample handling

### DIFF
--- a/src/mol-canvas3d/canvas3d.ts
+++ b/src/mol-canvas3d/canvas3d.ts
@@ -360,12 +360,12 @@ namespace Canvas3D {
                 helper.handle.scene.update(void 0, true);
                 helper.camera.scene.update(void 0, true);
                 const prevPickDirty = pickHelper.dirty;
-                draw(true);
+                draw(true, true);
                 pickHelper.dirty = prevPickDirty; // marking does not change picking buffers
             }
         }
 
-        function render(force: boolean) {
+        function render(force: boolean, allowMulti: boolean) {
             if (webgl.isContextLost) return false;
 
             let resized = false;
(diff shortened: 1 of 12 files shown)


src/mol-plugin/behavior/dynamic/representation.ts
```diff
@@ -186,8 +186,12 @@ export const SelectLoci = PluginBehavior.create({
                         Structure.areEquivalent(structure, oldStructure) &&
                         Structure.areHierarchiesEqual(structure, oldStructure)) return;
 
-                    const reprs = this.ctx.state.data.select(StateSelection.Generators.ofType(SO.Molecule.Structure.Representation3D, ref));
-                    for (const repr of reprs) this.applySelectMark(repr.transform.ref, true);
+                    const children = this.ctx.state.data.select(StateSelection.children(ref));
```
arose, line 189: @dsehnal  add .childrenOfType helper?

src/mol-canvas3d/canvas3d.ts
```diff
@@ -416,9 +421,9 @@ namespace Canvas3D {
         let currentTime = 0;
         let drawPaused = false;
 
-        function draw(force?: boolean) {
+        function draw(force?: boolean, allowMulti?: boolean) {
```
dsehnal, line 424: Since we can break some APIs with 3.0, let's change this to ``options?: { force?: boolean, allowMulti?: boolean }``?

src/mol-canvas3d/passes/multi-sample.ts
```diff
@@ -114,7 +122,8 @@ export class MultiSamplePass {
         }
     }
 
-    private renderMultiSample(renderer: Renderer, camera: Camera | StereoCamera, scene: Scene, helper: Helper, toDrawingBuffer: boolean, transparentBackground: boolean, props: Props) {
+    private renderMultiSample(ctx: RenderContext, toDrawingBuffer: boolean, props: Props) {
```
dsehnal, line 125: Phew... :)

src/mol-plugin/behavior/dynamic/representation.ts
dsehnal: ``StateSelection.children(ref).ofType(SO.Molecule.Structure.Representation3D)`` should do exactly that.

PR #333, opened 2021-12-31
Use size theme in molecular/gaussian surface & label representations

arose:
Allows, for example, to create a Gaussian surface based on B-factors.

image


dsehnal (approved):
LGTM

PR #335, opened 2022-01-08
move Viewer class to separate file

dsehnal:
Allows to import ``Viewer`` in 3rd party apps without having to deal with 

```ts
import './embedded.html';
import './favicon.ico';
import './index.html';
require('mol-plugin-ui/skin/light.scss');
```


arose (approved):
good idea, thanks, code looks fine

PR #336, opened 2022-01-08
IndexPairBonds improvements

arose:
- Improve bond assignment from ``IndexPairBonds``
    - Add ``id`` for mapping to source data
    - Fix assignment of bonds with unphysical length

### DIFF
--- a/src/mol-model-formats/structure/property/bonds/index-pair.ts
+++ b/src/mol-model-formats/structure/property/bonds/index-pair.ts
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2019-2021 Mol* contributors, licensed under MIT, See LICENSE file for more info.
+ * Copyright (c) 2019-2022 Mol* contributors, licensed under MIT, See LICENSE file for more info.
  *
  * @author Alexander Rose <alexander.rose@weirdbyte.de>
  */
@@ -10,9 +10,9 @@ import { Column } from '../../../../mol-data/db';
 import { FormatPropertyProvider } from '../../common/property';
 import { BondType } from '../../../../mol-model/structure/model/types';
 import { ElementIndex } from '../../../../mol-model/structure';
-import { DefaultBondMaxRadius } from '../../../../mol-model/structure/structure/unit/bonds/common';
 
 export type IndexPairsProps = {
+    readonly id: ArrayLike<number>
     readonly order: ArrayLike<number>
     readonly distance: ArrayLike<number>
     readonly flag: ArrayLike<BondType.Flag>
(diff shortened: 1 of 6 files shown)


src/mol-model-formats/structure/property/bonds/index-pair.ts
```diff
@@ -45,27 +47,42 @@ export namespace IndexPairBonds {
     export type Data = {
         pairs: {
             indexA: Column<number>,
-            indexB: Column<number>
+            indexB: Column<number>,
+            id?: Column<number>,
```
dsehnal, line 51: Should this be called ``key`` instead? I think it would better reflect what it is for.

src/mol-model/structure/structure/unit/bonds/inter-compute.ts
```diff
@@ -82,11 +82,22 @@ function findPairBonds(unitA: Unit.Atomic, unitB: Unit.Atomic, props: BondComput
...
                 const dist = getDistance(unitA, aI, unitB, bI);
-                if ((d !== -1 && equalEps(dist, d, 0.5)) || dist < maxDistance) {
+                const pairingThreshold = getPairingThreshold(
+                    aeI, beI, getElementThreshold(aeI), getElementThreshold(beI)
+                );
+
+                if ((d !== -1 && equalEps(dist, d, 0.5)) ||
+                    (maxDistance !== -1 && dist < maxDistance) ||
+                    dist < pairingThreshold
+                ) {
```
dsehnal, line 100: The logic here seems a bit fishy. I think it should be something like this instead:

```ts
if (isHydrogen(aeI) && isHydrogen(beI)) {
  // TODO: check if molecule has only 2 bonds and both are hydrogen to take care of edge case?
  continue;
}

if (maxDistance < 0) {
  // pairingThreshold based check
} else if (dist < maxDistance) {
  builder.add(_aI, _bI, { order: order[i], flag: flag[i] });
}
```

src/mol-model-formats/structure/property/bonds/index-pair.ts
```diff
@@ -45,27 +47,42 @@ export namespace IndexPairBonds {
...
             distance?: Column<number>,
             flag?: Column<BondType.Flag>,
         },
         count: number
     }
 
-    export const DefaultProps = { maxDistance: DefaultBondMaxRadius };
+    export const DefaultProps = {
+        /**
+         * If -1, test using element-based threshold, otherwise distance in Angstrom.
```
dsehnal, line 66: If *negative*, test using element-based threshold, otherwise distance in Angstrom.

?

src/mol-model/structure/structure/unit/bonds/intra-compute.ts
```diff
@@ -72,11 +73,21 @@ function findIndexPairBonds(unit: Unit.Atomic) {
 
             const _bI = SortedArray.indexOf(unit.elements, bI) as StructureElement.UnitIndex;
             if (_bI < 0) continue;
-            if (isHa && type_symbol.value(bI) === 'H') continue;
+
+            const beI = getElementIdx(type_symbol.value(bI));
+            // only element-based test when maxDistance !== -1
+            if (maxDistance !== -1 && isHa && isHydrogen(beI)) continue;
```
dsehnal, line 79: Same as for inter-bonds.

src/mol-model-formats/structure/property/bonds/index-pair.ts
arose: ok, changed it to key

src/mol-model/structure/structure/unit/bonds/inter-compute.ts
arose: yes, thanks, gave it an overhaul

src/mol-model-formats/structure/property/bonds/index-pair.ts
arose: changed it

src/mol-model/structure/structure/unit/bonds/intra-compute.ts
arose: done

Issue #183, opened 2021-05-09
Camera rock animation


arose:
Similar to spin rotating around the z axis but only a set amount of degrees clockwise before switching to counter-clockwise. Use easing when switching direction.

PR #337, opened 2022-01-09
Add rock animation

arose:
The to rock animation in trackball controls can be used while exploring a structure interactively. The rocking stops when using the mouse and starts again after. Good for looking at binding site or ligands where a continuous spin is less helpful.


src/mol-canvas3d/controls/trackball.ts
```diff
@@ -434,11 +452,38 @@ namespace TrackballControls {
...
+            Vec2.add(_rotCurr, _rotPrev, _spinSpeed);
+        }
+
+        let _rockAngleSum = 0;
+        let _rockDirection = 1;
+        const _rockSpeed = Vec2.create(0.005, 0);
+        function rock(deltaT: number) {
+            if (p.animate.name !== 'rock' || p.animate.params.speed === 0 || _isInteracting) return;
+
+            // TODO get rid of the 3.3 factor
```
arose, line 468: @dsehnal do you have a better idea for formulating this? I couldn't figure out how smoothly change direction at a specific angle, hence the magic factor.

Relatedly, could you look into how to add a ``PluginStateAnimation`` for this? To make rocking videos.

arose:
#183

src/mol-canvas3d/controls/trackball.ts
```diff
@@ -40,8 +41,16 @@ export const TrackballControlsParams = {
     zoomSpeed: PD.Numeric(7.0, { min: 1, max: 15, step: 1 }),
     panSpeed: PD.Numeric(1.0, { min: 0.1, max: 5, step: 0.1 }),
 
-    spin: PD.Boolean(false, { description: 'Spin the 3D scene around the x-axis in view space' }),
-    spinSpeed: PD.Numeric(1, { min: -20, max: 20, step: 1 }),
+    animate: PD.MappedStatic('off', {
```
dsehnal, line 44: Use ``cycle`` styles for this?

src/mol-canvas3d/controls/trackball.ts
dsehnal: The magic factor: maybe instead of ``_rockAngleSum`` animate a ``t \in [0, 1]``  parameter and make the ``currentAngle \in [-angle, +angle]`` a function of that?

-----

### Animations

Expose ``animation`` properties of the trackball. Something along the lines of:

```ts
{
  animations: {
    rock: { reset: () => ..., tick: t => ... },
    spin: ...
  }
}
```

Then in the ``PluginStateAnimation.setup`` call ``reset``` and ``tick`` in ``PluginStateAnimation.apply``. Can also remember the old state in ``setup`` and restore it in ``teardown``.

src/mol-canvas3d/controls/trackball.ts
arose: nah, hinders discoverability and is an odd ux pattern here, imho

src/mol-canvas3d/controls/trackball.ts
arose: > The magic factor: maybe instead of _rockAngleSum animate a t \in [0, 1] parameter and make the currentAngle \in [-angle, +angle] a function of that?

yeah, exactly what I could not figure out 😅

arose:
@dsehnal couldn't figure out a closed formula 🤷‍♂️, so I pre-calculate an array with angles for the rocking animation and sample from it.

arose:
@dsehnal thanks for the `sin()` tip, much cleaner and similar enough to `smoothstep()`. Does the rock state animation loops smoothly for you when exported as a video?

src/mol-canvas3d/controls/trackball.ts
```diff
@@ -434,11 +451,36 @@ namespace TrackballControls {
...
-            if (!_isInteracting) Vec2.add(_rotCurr, _rotPrev, _spinSpeed);
+            Vec2.add(_rotCurr, _rotPrev, _spinSpeed);
+        }
+
+        let _rockPhase = 0;
+        const _rockSpeed = Vec2.create(0.005, 0);
+        function rock(deltaT: number) {
+            if (p.animate.name !== 'rock' || p.animate.params.speed === 0 || _isInteracting) return;
+
+            const dt = deltaT / 4000; // normalize to 4s duration
```
dsehnal, line 466: Are you sure about this 4s normalization? I would rather adjust the speed param itself and make it "per 1s".

dsehnal (approved):
No visible stutter to me:

https://user-images.githubusercontent.com/21145776/149750867-41a9be47-8c55-4a7f-9d20-adc7ae7d706a.mp4

src/mol-canvas3d/controls/trackball.ts
arose: 👍

PR #341, opened 2022-01-11
Pass animation info to state animations

### LINKED ISSUE #339: Stuttering when looping exported camera spin video
> The animation appears to include frames at both 0 and 360 degrees, making it stutter at every full rotation.

reported on twitter: https://twitter.com/mortenkj/status/1480237573641486337

dsehnal:
- Fix camera stutter for "camera spin"

Closes #339

### DIFF
--- a/src/mol-plugin-state/animation/built-in/camera-spin.ts
+++ b/src/mol-plugin-state/animation/built-in/camera-spin.ts
@@ -39,13 +39,9 @@ export const AnimateCameraSpin = PluginStateAnimation.create({
             return { kind: 'finished' };
         }
 
-        const phase = clamp(t.current / ctx.params.durationInMs, 0, 1);
-
-        if (phase >= 0.99999) {
-            ctx.plugin.canvas3d?.requestCameraReset({ snapshot, durationMs: 0 });
-            return { kind: 'finished' };
-        }
-
+        const phase = t.animation
+            ? t.animation?.currentFrame / (t.animation.frameCount + 1)
+            : clamp(t.current / ctx.params.durationInMs, 0, 1);
         const angle = 2 * Math.PI * phase * ctx.params.speed * (ctx.params.direction === 'ccw' ? -1 : 1);
 
         Vec3.sub(_dir, snapshot.position, snapshot.target);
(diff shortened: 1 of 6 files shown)


src/mol-plugin-state/animation/built-in/camera-spin.ts
```diff
@@ -39,13 +39,9 @@ export const AnimateCameraSpin = PluginStateAnimation.create({
...
 
-        const phase = clamp(t.current / ctx.params.durationInMs, 0, 1);
-
-        if (phase >= 0.99999) {
-            ctx.plugin.canvas3d?.requestCameraReset({ snapshot, durationMs: 0 });
-            return { kind: 'finished' };
-        }
-
+        const phase = t.animation
+            ? t.animation?.currentFrame / (t.animation.frameCount + 1)
```
arose, line 43: I don't quite understand the `+1` here

arose (commented):
Thanks! I see less stutter. There is still something but that might from my video player???

```
if (t.current === 0) {
    return { kind: 'next', state: animState };
}
```
For my understanding, this is skip the first frame because it is already rendered?

src/mol-plugin-state/animation/built-in/camera-spin.ts
dsehnal: > Thanks! I see less stutter. There is still something but that might from my video player???

It looked fine in my VLC.


https://user-images.githubusercontent.com/21145776/149094312-b5b06878-f666-4440-a050-a00794f64986.mp4



> 
> ```
> if (t.current === 0) {
>     return { kind: 'next', state: animState };
> }
> ```
> For my understanding, this is skip the first frame because it is already rendered?

Yes.

dsehnal: > I don't quite understand the `+1` here

That is to skip the last frame because 0 and 360 are the same angles.

arose: ok, thanks

arose: > It looked fine in my VLC.

Ok, must be a fluke on my machine, telling me I need as faster GPU :)

It works fine when I look at it through the browser but in the VLC I see subtle stutter.

PR #349, opened 2022-01-22
add custom theme colors

arose:
Add custom theme colors for

- element-symbol
- molecule-type
- residue-name
- secondary-structure

Using MappedStatic param with "default" and "custom" to avoid always adding all the colors to the state.


dsehnal (approved):
Very cool. Code looks good and worked when I tested it locally.

PR #353, opened 2022-01-27
emit drag event whenever started within viewport

arose:
What do you think? Would it cause any issues? @dsehnal


dsehnal (approved):
Should be fine, if something goes horribly wrong could always reverse it.

PR #365, opened 2022-02-04
Volume Streaming Error Message

JonStargaryen:
Folks on our end are interested in a more user-friendly error message when volume streaming can't be enabled. We would like a distinction between unexpected/parsing errors and the case where a structure just doesn't provide density data, indicated by a 404.

This would bring the error messages in alignment with the assembly symmetry component.


dsehnal (approved):
Thanks, that's a nice improvement!

PR #366, opened 2022-02-05
cleaner default representation style

arose:
I am proposing some tweaks to the cartoon and focus representation defaults

- Cartoon: use ``nucleotide-ring`` instead of ``nucleotide-block``
- Focus: use ``xrayShaded`` instead of opacity; adjust target size; don't show non-covalent interactions twice

New
image

Old
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,11 @@ Note that since we don't clearly distinguish between a public and private interf
 
 ## [Unreleased]
 
-- Fix ``xrayShader`` & ``ignoreLight`` params not working at the same time
+- Fix ``xrayShaded`` & ``ignoreLight`` params not working at the same time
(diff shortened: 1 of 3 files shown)


arose:
ping @dsehnal @JonStargaryen @mandarsd any objections?

dsehnal:
Looks much better!

JonStargaryen:
Yeah, nice change!

dhrblav:
Nice changes indeed, however I still find the focus highlight hard to see and I suggest a brighter representation.

arose:
> Nice changes indeed, however I still find the focus highlight hard to see and I suggest a brighter representation.

Thanks for the suggestions, that's is a good idea (used that myself before...). Unfortunately it is currently not possible to make all color themes brighter , and the focus representation needs to work for all themes, at least in the generic viewer.

Issue #363, opened 2022-02-01
Canvas3d settings overwritten on loading pdb


jpattle:
Hi,

I have an instance of the molstar plugin that has an initial `PluginUISpec` that turns occlusion off and outline on in the canvas3d post-processing settings:

```
await createPluginUI(
    htmlElement,
    {
        ...DefaultPluginUISpec(),
        canvas3d: {
            occlusion: { name: 'off', params: {} },
            outline: { name: 'on', params: { scale: 1, threshold: 0.33, color: Color(0) } } 
        }
    }
)
```

However, whenever I load a pdb into the viewer these settings are overwritten to the default (with occlusion on, and outline off). This happens whether I load the pdb by the 'Download Structure' action, the 'Download File' action, or the 'Open Files' action.

I also noticed that this happens in the official molstar viewer found on [your site](https://molstar.org/viewer/).

https://user-images.githubusercontent.com/84507154/152007231-8eb5c78f-114e-4ff0-82f2-415b4c6cd40c.mov

How do I configure the plugin such that my initial canvas3d settings are the preserved for whenever a pdb is loaded in?

Thank you!

dsehnal:
This seems to happen by design (not sure why @arose added it).

For now you will have to reset the params manually after the structure is loaded with ``plugin.canvas3d.setProps``.

arose:
> This seems to happen by design (not sure why @arose added it).

I added it to support component/representation presets that can change post-processing settings. Not my best decision...

I wanted a one-click solution to get preset styles like Goodsell's Illustrative. Doing this within the component/representation presets is not good because makes them much less reusable and has the side effect you are seeing. I propose to 1) remove changing the post-processing params in the component/representation presets and 2) add a new tool panel called "Wizards" on the right as a home for these one-click presets that may make sweeping changes to parameters all over the place.

jpattle:
Ah I see, thank you for the explanation. For now my workaround will be exactly as David suggested; change the params manually after the structure is loaded. My experience is that these settings are reset every time a structure is loaded into a zero state; so if I load a structure, update the settings, unload the structure and then load a new structure I will need to update the settings once again.

Thank you for your help,

Jason

PR #367, opened 2022-02-05
Quick styles

arose:
I propose this as a solution to #363 so we can have reusable representation presets and one-click styles.

image

Reset: Applies default representation preset. Set outline and occlusion effects to defaults.
image

Illustrative: Applies illustrative representation preset. Enables outline and occlusion effects. Enables ignore-light parameter.
image

Stylized: Applies no representation preset. Enables outline and occlusion effects. Enables ignore-light representation parameter.
image


arose:
ping @jpattle @dsehnal

dsehnal:
Looks cool. Maybe instead of "reset" use "default"?

PR #368, opened 2022-02-06
iterate over `structure.unitSymmetryGroups`, fixes #364

### LINKED ISSUE #364: Exported Structures Render without Secondary Structure if Transformed
Alex mentioned in #334 that transformed chains get exported without secondary structure information. A problematic file is attached. As mentioned before, it's rendered with SSE after the `database_2` category is omitted.

No `struct_conf` content is exported due to the check at:
https://github.com/molstar/molstar/blob/b53debcfef324066f71ba0f60a8b3f45ef6f2a31/src/mol-model/structure/export/categories/secondary-structure.ts#L83

Can this check be more lenient and include records if the original unit would have been included and was only moved as part of a superposition?

[1U19.zip](https://github.com/molstar/molstar/files/7998366/1U19.zip)

JonStargaryen:
This ensures that SSE information is exported for transformed structures.
`const u = unit.units[0];` is used in other places, I assume this is also the way to go here.

### DIFF
--- a/src/mol-model/structure/export/categories/secondary-structure.ts
+++ b/src/mol-model/structure/export/categories/secondary-structure.ts
@@ -78,12 +78,12 @@ function findElements<T extends SecondaryStructure.Element>(ctx: CifExportContex
     const ssElements: SSElement<any>[] = [];
 
     const structure = ctx.structures[0];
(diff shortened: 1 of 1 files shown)


arose:
thanks @JonStargaryen

PR #369, opened 2022-02-07
Better support for atom_site.pdbx_sifts_xref

dsehnal:
- Rename "best database mapping" to "SIFTS Mapping"
- Add schema and export support for ``atom_site.pdbx_sifts_xref_*`` fields
- Add schema export support for ``atom_site.pdbx_label_index`` field

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ Note that since we don't clearly distinguish between a public and private interf
 
 ## [Unreleased]
 
+- Rename "best database mapping" to "SIFTS Mapping"
+- Add schema and export support for ``atom_site.pdbx_sifts_xref_*`` fields
+- Add schema export support for ``atom_site.pdbx_label_index`` field
+
 ## [v3.1.0] - 2022-02-06
 
 - Fix ``xrayShaded`` & ``ignoreLight`` params not working at the same time
--- a/data/cif-field-names/mmcif-field-names.csv
+++ b/data/cif-field-names/mmcif-field-names.csv
@@ -24,6 +24,10 @@ atom_site.auth_asym_id
 atom_site.auth_seq_id
 atom_site.pdbx_PDB_model_num
 atom_site.ihm_model_id
+atom_site.pdbx_sifts_xref_db_name
+atom_site.pdbx_sifts_xref_db_acc
+atom_site.pdbx_sifts_xref_db_num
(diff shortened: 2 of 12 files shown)


arose (commented):
Nice feature.

Please change the docs of the `alignSequences` param to `For "Chain"-based 3D superposition, perform a sequence alignment and use the aligned residue pairs as a guide.` to indicate it only affects the "Chain" option.

I suggest to add a new param `traceOnly` that is `true` by default (structures would need to be extremely similar to use `false` if I am not mistaken, would like to see an example it's useful) and would affect "Chain" and "Uniprot".

There is a console.log statement in src\mol-model\structure\structure\util\superposition-db-mapping.ts

Will it work when structures have different number of atoms due to alternate locations, including micro-heterogeneity?

Would be nice to have a legend for the SIFTS mapping color theme.

Some message if structures can't be aligned would be nice. Currently for 4hhb and 1tqn I get "Superposed 2 structures with avg. RMSD NaN Å."

Some more info in the button tooltip would be good. As far as I understand it only works 2 (or more) structures have at least one chain with the same sifts mapping. Is the main purpose to align a (large) number of structures for the same uniprot sequence?

How does it behave when there is more than one chain with sifts mapping?

Since there is already a uniprot id given in mmcif files it could be used - even if it is outdated compared to sifts - as a fallback to the sifts mapping: decide on a pair of chains to automatically align based on _struct_ref, then do sequence alignment guided 3d superposition. @JonStargaryen

dsehnal:
Thanks for the comments @arose.

> I suggest to add a new param `traceOnly` that is `true` by default (structures would need to be extremely similar to use `false` if I am not mistaken, would like to see an example it's useful) and would affect "Chain" and "Uniprot".

Agree, but let's do this in a separate PR. @JonStargaryen can you help with this?

> Will it work when structures have different number of atoms due to alternate locations, including micro-heterogeneity?

It should, tho it will introduce a small error in the RMSD/alignment.

> Some more info in the button tooltip would be good. As far as I understand it only works 2 (or more) structures have at least one chain with the same sifts mapping. Is the main purpose to align a (large) number of structures for the same uniprot sequence?

Yes, use "common SIFTS mapping" to superpose structures. I will improve the tooltip.

> How does it behave when there is more than one chain with sifts mapping?

Should work the same, will use the "intersection" of the mappings for the superposition.

Issue #371, opened 2022-02-08
Add TraceOnly option for Structure Superposition


dsehnal:
Add a new param `traceOnly` to [``StructureSuperpositionParams``](https://github.com/molstar/molstar/blob/master/src/mol-plugin-ui/structure/superposition.tsx#L50) that is `true` by default and would affect "Chain" and "Uniprot" superposition options.

@JonStargaryen would you be able to help with this?

JonStargaryen:
I've opened a PR: #373

dsehnal:
Closed by #373, thanks @JonStargaryen

PR #373, opened 2022-02-10
Add TraceOnly option for Structure Superposition

JonStargaryen:
This addresses #371.

Is it intentional that the camera only resets for the UniProt alignment but not for the ones based on chains/atoms?

### DIFF
--- a/src/mol-model/structure/structure/util/superposition-sifts-mapping.ts
+++ b/src/mol-model/structure/structure/util/superposition-sifts-mapping.ts
@@ -98,7 +98,6 @@ function getPositionTables(index: IndexEntry[], pivot: number, other: number, N:
 }
 
 function traceAtom(unit: Unit.Atomic, eI: ElementIndex): boolean {
-    // TODO could check based on traceElementIndex too
     const l = unit.model.atomicHierarchy.atoms.label_atom_id.value(eI);
     return l === 'CA' || l === 'BB' || l === `C4'`;
 }
--- a/src/mol-model/structure/structure/util/superposition-sifts-mapping.ts
+++ b/src/mol-model/structure/structure/util/superposition-sifts-mapping.ts
@@ -64,25 +64,29 @@ function getPositionTables(index: IndexEntry[], pivot: number, other: number, N:
 
     let o = 0;
     for (const { pivots } of index) {
-        const a = pivots[pivot];
-        const b = pivots[other];
+        let a = pivots[pivot];
+        let b = pivots[other];
         if (!a || !b) continue;
+
+        if (traceOnly) {
+            a = findTraceAtom(...a);
+            b = findTraceAtom(...b);
+            if (!a || !b) continue;
+        }
 
         const l = Math.min(a[2] - a[1], b[2] - b[1]);
 
         // TODO: check if residue types match?
         for (let i = 0; i < l; i++) {
-            const eI = (a[1] + i) as ElementIndex;
-            const eJ = (b[1] + i) as ElementIndex;
-            if (traceOnly && (!traceAtom(a[0], eI) || !traceAtom(b[0], eJ))) continue;
-
+            let eI = (a[1] + i) as ElementIndex;
             xs.x[o] = a[0].conformation.x(eI);
             xs.y[o] = a[0].conformation.y(eI);
             xs.z[o] = a[0].conformation.z(eI);
 
-            ys.x[o] = b[0].conformation.x(eJ);
-            ys.y[o] = b[0].conformation.y(eJ);
-            ys.z[o] = b[0].conformation.z(eJ);
+            eI = (b[1] + i) as ElementIndex;
+            ys.x[o] = b[0].conformation.x(eI);
+            ys.y[o] = b[0].conformation.y(eI);
+            ys.z[o] = b[0].conformation.z(eI);
             o++;
         }
     }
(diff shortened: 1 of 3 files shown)


src/mol-model/structure/structure/util/superposition-sifts-mapping.ts
```diff
@@ -69,25 +69,39 @@ function getPositionTables(index: IndexEntry[], pivot: number, other: number, N:
 
         const l = Math.min(a[2] - a[1], b[2] - b[1]);
 
-        // TODO: allow to use just backbone atoms?
         // TODO: check if residue types match?
         for (let i = 0; i < l; i++) {
-            let eI = (a[1] + i) as ElementIndex;
+            const eI = (a[1] + i) as ElementIndex;
+            const eJ = (b[1] + i) as ElementIndex;
+            if (traceOnly && (!traceAtom(a[0], eI) || !traceAtom(b[0], eJ))) continue;
```
dsehnal, line 76: This will only work if the atoms are in the same order, which might not be the case. I know the previous code didn't handle that either, but I can complain about that here :D

Need to have something like:

```ts
if (traceOnly) {
  const  trace = findTraceAtoms(a, a[1], a[2]);
  for (let j = b[1]; j < b[2]; j++) {
    if (!trace.has(...)) continue;
    ...
  }
}
```

dsehnal (commented):
Please add a message to changelog too.

>Is it intentional that the camera only resets for the UniProt alignment but not for the ones based on chains/atoms?

Not intentional, but I think it makes sense, so add that for chains too?

----

And thanks for the PR.

src/mol-model/structure/structure/util/superposition-sifts-mapping.ts
```diff
@@ -23,7 +23,7 @@ export interface AlignmentResult {
     failedPairs: [number, number][]
```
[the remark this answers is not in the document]
JonStargaryen: Done in f1d8f0e.

src/mol-model/structure/structure/util/superposition-sifts-mapping.ts
```diff
@@ -23,7 +23,7 @@ export interface AlignmentResult {
...
 }
 
-export function alignAndSuperposeWithSIFTSMapping(structures: Structure[]): AlignmentResult {
+export function alignAndSuperposeWithSIFTSMapping(structures: Structure[], traceOnly: boolean = true): AlignmentResult {
```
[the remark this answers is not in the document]
JonStargaryen: Sounds good, will try to do that going forward. Changed with 7329fa5.

src/mol-model/structure/structure/util/superposition-sifts-mapping.ts
JonStargaryen: Reverted back to some of the original code and added an optional step that narrows the range down to the backbone atom if traceOnly (b21a78a). Seems to work on a manipulated file.

The `BB` check might be relevant when dealing with a SIFTS mapping, correct?

JonStargaryen:
> Please add a message to changelog too.

Done with 1d127f2.

> Not intentional, but I think it makes sense, so add that for chains too?

I think so too. 9841c77 will reset after all alignment operations.

src/mol-model/structure/structure/util/superposition-sifts-mapping.ts
```diff
@@ -85,9 +91,26 @@ function getPositionTables(index: IndexEntry[], pivot: number, other: number, N:
...
+        ys.x = ys.x.slice(0, o);
+        ys.y = ys.y.slice(0, o);
+        ys.z = ys.z.slice(0, o);
+    }
     return [xs, ys];
 }
 
+function findTraceAtom(unit: Unit.Atomic, start: ElementIndex, end: ElementIndex): [Unit.Atomic, ElementIndex, ElementIndex] | undefined {
+    for (let i = start; i < end; i++) {
+        const l = unit.model.atomicHierarchy.atoms.label_atom_id.value(i);
```
dsehnal, line 107: A small nitpick, but maybe move ``unit.model.atomicHierarchy.atoms.label_atom_id`` out of the loop.

src/mol-model/structure/structure/util/superposition-sifts-mapping.ts
```diff
@@ -57,19 +58,24 @@ export function alignAndSuperposeWithSIFTSMapping(structures: Structure[]): Alig
...
-        const b = pivots[other];
+        let a = pivots[pivot];
+        let b = pivots[other];
         if (!a || !b) continue;
 
+        if (traceOnly) {
+            a = findTraceAtom(...a);
+            b = findTraceAtom(...b);
+            if (!a || !b) continue;
+        }
```
dsehnal, line 75: Sorry to be so annoying about this, but I think you should move the trace atom identification to the ``buildIndex`` function. E.g., just make ``start`` index of the trace atom https://github.com/molstar/molstar/blob/9841c773cb2e0dfe716611ad1052f5216903235a/src/mol-model/structure/structure/util/superposition-sifts-mapping.ts#L186 and make ``end = start + 1``. Then you won't need the ``slice`` at the end here and won't have to do ``...a`` in the function call.

JonStargaryen: Good call, and no worries :) I'll get back to this next week.

dsehnal (approved):
LGTM, thanks!

src/mol-model/structure/structure/util/superposition-sifts-mapping.ts
JonStargaryen: Implemented in [2a00248](https://github.com/molstar/molstar/pull/373/commits/2a002488124d5383eb01a36174ff96019692fae9).

Looks to me as if the chain/residue iterators are still needed, or is it possible to avoid that when in trace-only mode?

dsehnal: It's fine to keep them.

PR #378, opened 2022-02-20
fix visuals for aromatic/delocalized bonds

arose:
- Fix visual for bonds between two aromatic rings
- Fix visual for delocalized bonds (parsed from mmcif and mol2)


src/mol-model/structure/structure/unit/resonance.ts
```diff
@@ -0,0 +1,65 @@
...
+export type UnitResonance = {
+    /**
+     * Lookup for triplets of atoms in delocalized bonds.
+     *
+     * Does not include triplets that are part of aromatic rings.
+     */
+    delocalizedTriplets: {
+        /** Return 3rd element in triplet or undefined if `a` and `b` are not part of a triplet */
+        getThirdElement: (a: StructureElement.UnitIndex, b: StructureElement.UnitIndex) => StructureElement.UnitIndex | undefined
+    }
```
dsehnal, line 21: Wouldn't it make sense to add field ``triplets: SortedArray<StructureElement.UnitIndex>[]`` as well?

arose: yeah, I will add it and a lookup

PR #383, opened 2022-02-23
Resolve marking in main render loop

### LINKED ISSUE #382: Performance Impact of Temporal Multi-Sample
The sequence panel is quite laggy after the default multi-sample mode was changed to temporal. This gets intensified if there are additional subscribers to these events.

https://recordit.co/ZA0LVYkZBc

Is it possible to address this or should one just disable multi-sampling when running into this issue? Thanks!

dsehnal:
- Apply making in the main render loop instead of calling the draw function eagerly each time marking changes.
- Use ``throttleTime`` instead of ``debounceTime`` in the sequence viewer (felt better overall).

Should fix #382.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ Note that since we don't clearly distinguish between a public and private interf
 - Fix parsing contour-level from emdb v3 header files
 - Fix invalid CSS (#376)
 - Fix "texture not renderable" & "texture not bound" warnings (#319)
+- Resolve marking in main rernder loop to improve overall performance
+- Use ``throttleTime`` instead of ``debounceTime`` in sequence viewer for better responsiveness
 
 ## [v3.2.0] - 2022-02-17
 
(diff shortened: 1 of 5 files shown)


JonStargaryen (approved):
Many thanks, way more responsive now :)

LGTM, only spotted a typo (rernder) in the changelog.

arose (commented):
I like it, quite better than before. There is a regression. When just moving the mouse (no drag, no camera movement) it should do the multi-sampling in one pass (the forceOn flag), otherwise you get this annoying flicker.

dsehnal:
> I like it, quite better than before. There is a regression. When just moving the mouse (no drag, no camera movement) it should do the multi-sampling in one pass (the forceOn flag), otherwise you get this annoying flicker.

@arose did [486d12b](https://github.com/molstar/molstar/pull/383/commits/486d12b6ac48bafe741fc8dad464dd1acad9146a) fix this?

arose:
Yeah, that looks good! Thanks

PR #385, opened 2022-02-26
Coordinate trajectory related improvements

arose:
- Change line geometry default ``scaleFactor`` to 2 (3 is too big after fixing line rendering)
- Trajectory animation performance improvements
    - Reuse ``Model.CoarseGrained`` for coordinate trajectories
    - Avoid calculating ``InterUnitBonds`` when ``Structure.parent`` ones are empty
    - Reuse unit boundary if sphere has not changed too much
    - Don't show 'inter-bond' and 'element-cross' visuals in line representations of polymerAndLigand preset
- Fix additional mononucleotides detected as polymer components
- Fix and improve ``canRemap`` handling in ``IntraUnitBonds``


src/mol-model/structure/structure/structure.ts
```diff
@@ -232,7 +232,12 @@ class Structure {
 
     get interUnitBonds() {
         if (this.state.interUnitBonds) return this.state.interUnitBonds;
-        this.state.interUnitBonds = computeInterUnitBonds(this, { ignoreWater: !this.dynamicBonds });
+        if (this.parent && this.state.dynamicBonds === this.parent.state.dynamicBonds && this.parent.state.interUnitBonds?.edgeCount === 0) {
+            // no need to compute InterUnitBonds if parent's ones are empty
+            this.state.interUnitBonds = new InterUnitBonds(new Map());
+        } else {
```
dsehnal, line 238: This seems to be missing the condition ``this.parent.state.interUnitBonds !== void 0``, which is not the same as ``this.parent.state.interUnitBonds?.edgeCount === 0`` (e.g. it is possible to trigger the computation on the child structure without doing the same for the parent.

src/mol-model/structure/structure/unit.ts
```diff
@@ -220,9 +223,32 @@ namespace Unit {
...
+                        for (let i = 0, il = elements.length; i < il; i++) {
+                            pos(elements[i], v);
+                            v3add(center, center, v);
+                        }
+                        Vec3.scale(center, center, 1 / elements.length);
+
+                        // only invalidate boundary if sphere has changed too much
+                        if (Vec3.distance(center, b.sphere.center) / b.sphere.radius >= 1.0) {
+                            props.boundary = undefined;
+                        }
```
dsehnal, line 249: This seems almost as expensive as actually recomputing the boundary, no?

src/mol-model/structure/structure/structure.ts
arose: indeed, thanks

src/mol-model/structure/structure/unit.ts
arose: No, much cheaper, for the boundary we project each position onto 14 or 98 directions and go twice over them.

PR #386, opened 2022-02-27
Improve AO performance

arose:
- Reuse occlusion for secondary passes during multi-sampling
- Check if marking passes are needed before doing them
- Add ``scaleFactor`` parameter to adjust resolution of occlusion calculation

Not 100% sure about the default ``scaleFacor`` as there are some "halo" artifacts but the performance gain is substantial, ~1.8x for 1TQN with molecular surface. => Set default to 1, i.e. no down sampling. Doesn't look good enough in general, imho, if the scene has fine details. Leave it in as an option, though, still useful in many cases.


dsehnal (approved):
Code lgtm and I *think* it feels snappier.

PR #393, opened 2022-03-06
Zenodo import

arose:
image


src/mol-model-formats/structure/util.ts
```diff
@@ -47,30 +47,31 @@ export function guessElementSymbolTokens(tokens: Tokens, str: string, start: num
...
+    if (l === 0) return atomId; // empty
+    if (l === 1) return atomId; // one char
 
     if (l === 2) { // two chars
-        if (str === 'NA' || str === 'CL' || str === 'FE' || str === 'SI' ||
-            str === 'BR' || str === 'AS'
-        ) return str;
+        if (atomId === 'NA' || atomId === 'CL' || atomId === 'FE' || atomId === 'SI' ||
+            atomId === 'BR' || atomId === 'AS'
+        ) return atomId;
```
dsehnal, line 61: Since we are revisiting this:

```ts
const _guessSymbolTwoCharNames = new Set(['NA', 'CL', 'FE', 'SI', 'BR', 'AS']);
//... 
  if (_guessSymbolTwoCharNames.has(atomId)) return atomId;
```

src/mol-model-formats/structure/util.ts
```diff
@@ -47,30 +47,31 @@ export function guessElementSymbolTokens(tokens: Tokens, str: string, start: num
...
     }
 
-    if (l === 3) { // three chars
-        if (str === 'SOD') return 'NA';
-        if (str === 'POT') return 'K';
-        if (str === 'CES') return 'CS';
-        if (str === 'CAL') return 'CA';
-        if (str === 'CLA') return 'CL';
+    if (l === 3 && compId === atomId) { // three chars
```
dsehnal, line 64: Where does the ``compId === atomId`` condition come from? Seems a bit weird to me...

src/mol-model-formats/structure/util.ts
```diff
@@ -47,30 +47,31 @@ export function guessElementSymbolTokens(tokens: Tokens, str: string, start: num
...
+        if (atomId === 'SOD') return 'NA';
+        if (atomId === 'POT') return 'K';
+        if (atomId === 'CES') return 'CS';
+        if (atomId === 'CAL') return 'CA';
+        if (atomId === 'CLA') return 'CL';
     }
 
-    const c = str[0];
+    const c = atomId[0];
     if (c === 'C' || c === 'H' || c === 'N' || c === 'O' || c === 'P' || c === 'S') return c;
```
dsehnal, line 73:
```ts
const _guessSymbolFirstCharNames = new Set(['C', 'H', 'N', 'O', 'P', 'S']);
//... 
  if (_guessSymbolFirstCharNames.has(atomId[0])) return atomId;
```

dsehnal (approved):
That's a lot of code, good job on going through this... let's fix the bugs as the appear :)

For 4.0, I think we should make our goal a tree-shakeable (library) build.

src/mol-model-formats/structure/util.ts
arose: Yeah, that warrants a comment. It is to correctly assign elements for ions from files with Charmm FF names. For those ions compId and atomId are the same.

PR #394, opened 2022-03-06
Fix inverted hydrophobicity color model

hadim:
I think the hydrophobicity colormap is inverted.

This PR fixes the assigned colours but you could eventually prefer fixing the label "Hydrophobic" and "Hydrophilic" instead. I think the current fix is more intuitive (green for hydrophilic and red for hydrophobic) but let me know what you prefer.

**Before**
before

**After**
after

### DIFF
--- a/src/mol-theme/color/hydrophobicity.ts
+++ b/src/mol-theme/color/hydrophobicity.ts
@@ -58,7 +58,7 @@ export function HydrophobicityColorTheme(ctx: ThemeDataContext, props: PD.Values
 
     const scale = ColorScale.create({
         listOrName: props.list.colors,
-        domain: [max, min],
+        domain: [min, max],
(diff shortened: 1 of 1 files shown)


hadim:
In the screenshots looks at the colour of the two residues: GLU125 and TRP126

hadim:
Looking at this https://www.cgl.ucsf.edu/chimera/docs/ContributedSoftware/defineattrib/wwHydrophobicity.txt, it seems like the values has been all multiplied by -1. Maybe the best fix is to fix the values?

arose:
Thank you @hadim, the labels where wrong. I fixed it in f0e33e1e4ef1467fe6f158ebdedbd4b1db8d2b0f. Hydrophilic should be red and hydrophobic green. We mainly use for membrane proteins where we want to see the green part inside the membrane.

PR #396, opened 2022-03-15
DNATCO extension: Fix off-by-one error in element lookup

MadCatX:
This fixes some Confal pyramids failing to display in some peculiar structures such as [https://blackbox.ibt.biocev.org/devel_molstar/RCSB/cif_dnatco_updated/1jbs_v32C35A23.cif.gz](https://blackbox.ibt.biocev.org/devel_molstar/RCSB/cif_dnatco_updated/1jbs_v32C35A23.cif.gz).

(FYI, I have some more changes to the DNATCO extension lined up but this time I'll try to break them down to smaller MRs so things don't stall like last time)

### DIFF
--- a/src/extensions/dnatco/confal-pyramids/util.ts
+++ b/src/extensions/dnatco/confal-pyramids/util.ts
@@ -7,7 +7,7 @@
 
 import { ConfalPyramidsProvider } from './property';
 import { ConfalPyramidsTypes as CPT } from './types';
-import { OrderedSet, Segmentation } from '../../../mol-data/int';
+import { Segmentation } from '../../../mol-data/int';
 import { Vec3 } from '../../../mol-math/linear-algebra';
 import { ChainIndex, ElementIndex, ResidueIndex, Structure, StructureElement, StructureProperties, Unit } from '../../../mol-model/structure';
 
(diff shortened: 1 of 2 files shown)


src/extensions/dnatco/confal-pyramids/util.ts
```diff
@@ -123,18 +123,23 @@ export namespace ConfalPyramidsUtil {
     export class UnitWalker extends Utility {
         private getAtomIndices(names: string[], residue: Residue): ElementIndex[] {
             let rI = residue.start;
-            const rILast = residue.end - 1;
+            const rILast = residue.end;
             const indices: ElementIndex[] = [];
 
-            for (; rI !== rILast; rI++) {
+            for (; rI <= rILast; rI++) {
```
dsehnal, line 129: I think this should be just ``< rILast`` since the ranges are always ``[start, end)`` in Mol* (Unless you made it ``[start, end]`` in your implementation).

src/extensions/dnatco/confal-pyramids/util.ts
```diff
@@ -123,18 +123,23 @@ export namespace ConfalPyramidsUtil {
...
                 const eI = this.unit.elements[rI];
                 const loc = StructureElement.Location.create(this.structure, this.unit, eI);
```
dsehnal, line 131: Quite common to have mutable locations in the code to save some memory:

```ts
loc = StructureElement.Location.create(this.structure, this.unit, 0);
...
loc.element = ...;
const atom_id = tructureProperties.atom.label_atom_id(loc);
```

src/extensions/dnatco/confal-pyramids/util.ts
MadCatX: Oops, sorry, looks like I didn't fix the stop condition correctly after changing L126.

MadCatX:
I took an actual look at what this patch tries to do and I guess that the original fix was wrong to begin with. I replaced some of the rather convoluted ElementIndex lookup code with something more sensible. This fixes the problem and hopefully doesn't do anything weird with the indices.

Note that I also changed the way the NtC data is read from mmCif. The original code always read data only for the current model whereas now it reads everything right away. Without the change pyramids don't get displayed for anything but the first model. I'm quite sure the original code worked before, odd.

I'm sorry for the unnecessary confusion.

src/extensions/dnatco/confal-pyramids/util.ts
```diff
@@ -122,19 +115,22 @@ export namespace ConfalPyramidsUtil {
...
-                if (names.includes(thisName)) indices.push(eI);
+                if (names.includes(thisName)) indices.push(loc.element);
             }
 
-            if (indices.length === 0)
-                throw new Error(`Element ${name} not found on residue ${residue.index}`);
+            if (indices.length === 0) {
+                let namesStr = '';
+                for (const n of names)
+                    namesStr += `${n} `;
```
dsehnal, line 130:
```suggestion
                const namesStr = names.join(' ');
```

dsehnal (approved):
Looks consistent now. Added a small tip about joining strings.

src/extensions/dnatco/confal-pyramids/util.ts
MadCatX: Thanks :) I'll streamline this as part of some updates I'm working on.

PR #399, opened 2022-03-18
Make Confal pyramids colors changeable

MadCatX:
Let's use proper Molstar facilities for confal pyramids color theme


arose:
@MadCatX could you use a `default` and `custom` approach for the changeable colors like here

https://github.com/molstar/molstar/blob/6e5c20f442af01a85aff27e2bad956280532630d/src/mol-theme/color/element-symbol.ts#L42-L45

That way the colors are only stored in the state when the `custom` option is used.

MadCatX:
> @MadCatX could you use a `default` and `custom` approach for the changeable colors like here
> 
> https://github.com/molstar/molstar/blob/6e5c20f442af01a85aff27e2bad956280532630d/src/mol-theme/color/element-symbol.ts#L42-L45
> 
> That way the colors are only stored in the state when the `custom` option is used.

Nice, that's very sensible. Fixed in the updated code.

src/extensions/dnatco/confal-pyramids/color.ts
```diff
@@ -11,149 +11,248 @@ import { Location } from '../../../mol-model/location';
...
-    ['ZZ2S', ['Z', 'SYN']],
-    ['ZZS1', ['SYN', 'Z']],
-    ['ZZS2', ['SYN', 'Z']],
-]);
+const PyramidsColors = ColorMap({
+    NANT_Upr: DefaultClassColors.N,
+    NANT_Lwr: DefaultClassColors.N,
+    AA00_Upr: DefaultClassColors.A,
+    AA00_Lwr: DefaultClassColors.A,
+    AA02_Upr: DefaultClassColors.A,
```
dsehnal, line 40: To reduce the amount of code here, you can do:

```ts
{
  N: ['NANT_Upr', 'NANT_Lwr', ...],
  A: [...]
}
```

and transforms that to the object accepted by the ``ColorMap``.

MadCatX: Thanks for the suggestion. I tried it out to see what it'd look like but I'm not sure how I feel about it. The original object - while quite verbose - always kept the upper and lower part of the pyramid together whereas this compact mapping spreads them across multiple arrays. My other concern is that the original code created a logically laid out legend table that followed the published NtC ordering, this compact code doesn't.

Orignal (upper and lower parts are always together):
mapping_original

Compact ("random-ish" order from the user's perspective):
mapping_compact

dsehnal: Fair enough, thanks for trying it. I guess it's ok to reverse it back then.

dsehnal (approved):
LGTM, thanks for the PR.

PR #402, opened 2022-03-26
fix entity type handling, make case insensitive

arose:


src/examples/proteopedia-wrapper/index.ts
```diff
@@ -386,7 +386,10 @@ class MolStarProteopediaWrapper {
             if (!options?.hideLabels) {
                 // Labels
                 const waters = MS.struct.generator.atomGroups({
-                    'entity-test': MS.core.rel.eq([MS.struct.atomProperty.macromolecular.entityType(), 'water']),
+                    'entity-test': MS.core.str.match([
+                        MS.re('water', 'i'),
```
dsehnal, line 390: ``^water$`` for this to be equivalent?

Or should we add ``lower-case`` function instead?

src/extensions/cellpack/model.ts
```diff
@@ -128,7 +128,10 @@ async function getStructure(plugin: PluginContext, model: Model, source: Ingredi
     } else {
         query = MS.struct.modifier.union([
             MS.struct.generator.atomGroups({
-                'entity-test': MS.core.rel.eq([MS.ammp('entityType'), 'polymer'])
+                'entity-test': MS.core.str.match([
+                    MS.re('^polymer', 'i'),
```
dsehnal, line 132: ``^polymer&`` ?

dsehnal (commented):
Most of the regex tests will now do substring test, which I don't think we should allow even though it now "works".

I wonder if instead we should all a transform directly to the schema instead of applying the lowercase by hand?

arose:
> I wonder if instead we should all a transform directly to the schema instead of applying the lowercase by hand?

Yeah, why not. How about #405

dsehnal:
@arose 
Well, the case-insensitivity is supported by the CIF schema itself, so how about this to the ``Column.Schema.Str`` type, e.g.:

```ts
export type Str = { '@type': 'str', T: string, lowerCase?: boolean } & Base<'str'>
```

and update the column definitions accordingly? 

That way we can use it in other places where this is required.

arose:
> Well, the case-insensitivity is supported by the CIF schema itself, so how about this to the Column.Schema.Str type, e.g.:

Thanks, I like the idea but don't have time look into that and change the schema generator. Do you? Is #405 OK with you as a stop-gap measure?

arose:
superseded by #414

PR #403, opened 2022-03-26
improve pdb parsing

### LINKED ISSUE #156: Non-unique (per residue) atom names break ligand bonding structure in pdb files
Hey again,

I am loading in a PDB file that includes a protein and ligand. However, I see that the ligand is somehow bonded in a way that is not defined in the PDB file. After testing what is happening, I found that the ligand is not using unique atom names and somewhere in Molstar, it is creating new bonds because the ligand atom names are `N`, `C`, and `O`.

Is there a way to not have Molstar try to detect backbone atoms as it is still a valid PDB file (both VMD and NGL read the protein and ligand fine without creating bad bonds)?

Here's the ligand when the atom names are non-unique

Screen Shot 2021-03-30 at 4 34 07 PM

And when the atom names are incremental (`N1`, `C2`, `C3`, `O4`, ...)

Screen Shot 2021-03-30 at 4 33 48 PM

I'm attaching the two files that create these representations below (I used .txt because of GitHub, but they are .pdb)

[bad_ligand.txt](https://github.com/molstar/molstar/files/6229823/bad_ligand.txt)
[good_ligand.txt](https://github.com/molstar/molstar/files/6229824/good_ligand.txt)

Let me know if anything wasn't clear.
Thanks

arose:
- handle non unique atom and chain names
- fixes #156

### DIFF
--- a/src/mol-model-formats/structure/pdb/atom-site.ts
+++ b/src/mol-model-formats/structure/pdb/atom-site.ts
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2019-2021 mol* contributors, licensed under MIT, See LICENSE file for more info.
+ * Copyright (c) 2019-2022 mol* contributors, licensed under MIT, See LICENSE file for more info.
  *
  * @author David Sehnal <david.sehnal@gmail.com>
  * @author Alexander Rose <alexander.rose@weirdbyte.de>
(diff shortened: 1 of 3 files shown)


src/mol-model-formats/structure/pdb/atom-site.ts
```diff
@@ -40,26 +40,96 @@ export function getAtomSiteTemplate(data: string, count: number) {
...
+        labelAsymIds[i] = currLabelAsymId;
+
+        if (atomIdCounts.has(atomId)) {
+            const atomIdCount = atomIdCounts.get(atomId)! + 1;
+            atomIdCounts.set(atomId, atomIdCount);
+            atomId = `${atomId}_${atomIdCount}`;
+        } else {
+            atomIdCounts.set(atomId, 0);
+        }
+        labelAtomIds[i] = atomId;
```
dsehnal, line 109: |Id|Asym|
|--|-----|
|X|A|
|X|A|
|X|B|
|X|A|

Will end up being parsed as 

|Id|Asym|
|--|-----|
|X|A|
|X_1|A|
|X|B|
|X|A_1|

Should this instead be 

|Id|Asym|
|--|-----|
|X|A|
|X_1|A|
|X|B|
|X_2|A|

?

arose: I want [id, seqId, asymId]

X, 1, A
X, 1, A
X, 2, A
X, 1, B
X, 1, A

to become

X, 1, A
X_1, 1, A
X, 2, A
X, 1, B
X, 1, A_1

This assumes that chains are in order, that is, there are no interleaved parts of two chains. Chains in PDB files require a TER record to mark their end so interleaved chains should not exist. Added some code to assume interleaved chains when no TER record is present.

Atom ids only need to be unique within a residue.

The repeated use of a chain name I want to handle comes from files with more than 24 chains (e.g., from MD sims).

dsehnal: ok

arose:
Thanks for having a look @dsehnal!

PR #409, opened 2022-03-29
Fix unit mapping in bondedAtomicPairs MolScript query

JonStargaryen:
The Disulfide Bridges selection includes wrong residues.

disulfide


arose:
Nice find, thanks @JonStargaryen

PR #377, opened 2022-02-19
parse contour-level from emdb v3 header files

JonStargaryen:
There was a breaking change wrt the format of EMDB XML header files (https://www.wwpdb.org/news/news?year=2021#610ad31fef055f03d1f222bb). This causes all loading of EM maps on rcsb.org & molstar.org to fail.

PR #411, opened 2022-03-30
fix volume streaming for entries with multiple contour lists

JonStargaryen:
Fixes an issue introduced by me in https://github.com/molstar/molstar/pull/377: It's imperative to select the `map` tag before looking for the contour list.

Happens e.g. for 6Z0S.

### DIFF
--- a/src/mol-plugin/behavior/dynamic/volume-streaming/util.ts
+++ b/src/mol-plugin/behavior/dynamic/volume-streaming/util.ts
@@ -69,7 +69,8 @@ export async function getContourLevel(provider: 'emdb' | 'pdbe', plugin: PluginC
 export async function getContourLevelEmdb(plugin: PluginContext, taskCtx: RuntimeContext, emdbId: string) {
     const emdbHeaderServer = plugin.config.get(PluginConfig.VolumeStreaming.EmdbHeaderServer);
     const header = await plugin.fetch({ url: `${emdbHeaderServer}/${emdbId.toUpperCase()}/header/${emdbId.toLowerCase()}.xml`, type: 'xml' }).runInContext(taskCtx);
-    const contours = header.getElementsByTagName('contour');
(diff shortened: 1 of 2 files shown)


src/mol-plugin/behavior/dynamic/volume-streaming/util.ts
```diff
@@ -69,7 +69,8 @@ export async function getContourLevel(provider: 'emdb' | 'pdbe', plugin: PluginC
 export async function getContourLevelEmdb(plugin: PluginContext, taskCtx: RuntimeContext, emdbId: string) {
     const emdbHeaderServer = plugin.config.get(PluginConfig.VolumeStreaming.EmdbHeaderServer);
     const header = await plugin.fetch({ url: `${emdbHeaderServer}/${emdbId.toUpperCase()}/header/${emdbId.toLowerCase()}.xml`, type: 'xml' }).runInContext(taskCtx);
-    const contours = header.getElementsByTagName('contour');
+    const map = header.getElementsByTagName('map')[0];
+    const contours = map.getElementsByTagName('contour');
```
dsehnal, line 73: So this wasn't working at all before?

JonStargaryen: Only happens for header files which reference multiple primary contours (like https://ftp.wwpdb.org/pub/emdb/structures/EMD-11022/header/emd-11022.xml which has additional primary items as part of the `interpretation` tag).

Think that's quite rare and took us some time to run into the issue.

dsehnal: So doesn't this break the existing use case? Or the format was that when there is just 1 map there is a global contour property, but the map array is always present?

dsehnal: Ok, sounds good, makes sense. Thanks.

PR #421, opened 2022-04-09
better color theme typing

arose:
Quick attempt at better color theme typing. Can break things in some cases.


src/mol-repr/volume/slice.ts
```diff
@@ -32,7 +32,7 @@ export async function createImage(ctx: VisualContext, volume: Volume, theme: The
     const isoVal = Volume.IsoValue.toAbsolute(isoValue, volume.grid.stats).absoluteValue;
 
     // TODO more color themes
-    const color = theme.color.color(NullLocation, false);
+    const color = 'color' in theme.color ? theme.color.color(NullLocation, false) : Color(0xffffff);
```
dsehnal, line 35:
```suggestion
    const color = theme.color.color?.(NullLocation, false) ?? Color(0xffffff);
```

src/mol-theme/color.ts
```diff
@@ -64,6 +62,26 @@ interface ColorTheme<P extends PD.Params> {
...
+} & ColorThemeShared<P>
+
+type ColorThemeDirect<P extends PD.Params> = {
+    readonly granularity: ColorTypeDirect
+} & ColorThemeShared<P>
+
+type ColorTheme<P extends PD.Params, G extends ColorType = any> =
+    G extends ColorTypeLocation ? ColorThemeLocation<P> :
+        G extends ColorTypeGrid ? ColorThemeGrid<P> :
+            G extends ColorTypeDirect ? ColorThemeDirect<P> : never
```
dsehnal, line 83:
```suggestion
            G extends ColorTypeDirect ? ColorThemeDirect<P> : ColorTheme<any>
```

And this should prevent any breaking?

src/mol-repr/volume/slice.ts
arose: Doesn't work, need to type narrow first

src/mol-theme/color.ts
arose: I like to keep `never` so we are forced to handle any new granularity types

arose:
need to think some more about it

arose:
Set `ColorTypeLocation` as the default `ColorType` in `ColorTheme`, so you only need to set the `ColorType` when using `ColorTypeDirect` or `ColorTypeGrid`, which is probably hardly used. 

For example `ColorTheme<VolumeValueColorThemeParams, ColorTypeDirect>`

PR #423, opened 2022-04-13
Added an option to disable the drag overlay

jpattle:
Added an option to the components spec to allow disabling of the drag-and-drop overlay for Molstar that loads files into the viewer. By default the drag overlay is enabled, just as it was before.


dsehnal (approved):
LGTM, thanks for the PR.

PR #427, opened 2022-04-23
Transparent object outline

arose:
image


src/mol-gl/scene.ts
```diff
@@ -247,14 +249,35 @@ namespace Scene {
...
+                if (primitives.length === 0) return 0;
+                let count = 0;
+                let opacityAverage = 0;
+                for (let i = 0, il = primitives.length; i < il; ++i) {
+                    const p = primitives[i];
+                    if (!p.state.visible) continue;
+
+                    // TODO: simplify, handle in renderable.state???
+                    // uAlpha is updated in "render" so we need to recompute it here
+                    const alpha = clamp(p.values.alpha.ref.value * p.state.alphaFactor, 0, 1);
```
dsehnal, line 276: You mean assign the alpha value as part of ``renderable.state``?

dsehnal (approved):
Visuals look really cool!

Code looks fine as well (or at least nothing really "pops out"). Do the tests cover compiling all shader combinations of will we need to the rendering by hand for all visuals?

src/mol-gl/scene.ts
arose: yes, but I haven't though it through

arose:
> Do the tests cover compiling all shader combinations of will we need to the rendering by hand for all visuals?

The tests only cover the default params. Would we worthwhile to loop over common combinations.

PR #434, opened 2022-05-04
proposed fix for headless-gl issues

JonStargaryen:
Headless-gl requires a `[0]` suffix when getting the location of array uniforms (https://github.com/stackgl/headless-gl/issues/170). Without neither `uLightDirection` nor `uLightColor` get set.

See https://github.com/molstar/molrender/issues/3#issuecomment-1117615459.

### DIFF
--- a/src/mol-gl/webgl/uniform.ts
+++ b/src/mol-gl/webgl/uniform.ts
@@ -43,11 +43,7 @@ export function getUniformType(gl: GLRenderingContext, kind: UniformKind) {
 }
 
 export function isArrayUniform(kind: UniformKind) {
-    switch (kind) {
-        case 'b[]': case 'f[]': case 'i[]': case 'v2[]': case 'v3[]': case 'v4[]': case 'm3[]': case 'm4[]': return true;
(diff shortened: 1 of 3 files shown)


src/mol-gl/webgl/uniform.ts
```diff
@@ -42,6 +42,14 @@ export function getUniformType(gl: GLRenderingContext, kind: UniformKind) {
     }
 }
 
+export function isArrayUniform(kind: UniformKind) {
```
arose, line 45: I think just `return kind.endsWith('[]');` is fine

arose (commented):
Nice find, thank you @JonStargaryen.

arose:
Also, please add a line to the changelog

src/mol-gl/webgl/uniform.ts
JonStargaryen: Sure, simplified with [cbc941f](https://github.com/molstar/molstar/pull/434/commits/cbc941f19388c222f501ebf5520c5a36adcedd6f).

Changelog is updated as well.

PR #436, opened 2022-05-09
Fix Polymer Chain Instance Coloring

JonStargaryen:
- currently, instances of the same chain get colored identically
- looks like a typo

instance-color


arose:
Looks good, thanks @JonStargaryen

PR #445, opened 2022-05-27
Add check that a data block contains volume data before parsing

aliaksei-chareshneu:


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ Note that since we don't clearly distinguish between a public and private interf
 
 ## [Unreleased]
 
+## [v3.8.3] - 2022-05-27
+- Add check that a data block contains volume data before parsing
```
dsehnal, line 10:
```suggestion
- Add check that a data block contains volume data before parsing
```

This should be in the unreleased section until we make a new release with the changes.

src/mol-plugin-state/formats/volume.ts
```diff
@@ -197,15 +198,17 @@ export const DscifProvider = DataFormatProvider({
     parse: async (plugin, data, params?: DsCifParams) => {
         const cifCell = await plugin.build().to(data).apply(StateTransforms.Data.ParseCif).commit();
         const b = plugin.build().to(cifCell);
-        const blocks = cifCell.obj!.data.blocks.slice(1); // zero block contains query meta-data
+        const blocks = cifCell.obj!.data.blocks.slice(0); // iterate over all blocks as even 0th can contain data
```
dsehnal, line 201:
```suggestion
        const blocks = cifCell.obj!.data.blocks;
```

No need to call the slice since we are iterating all of them.

src/mol-plugin-state/formats/volume.ts
```diff
@@ -197,15 +198,17 @@ export const DscifProvider = DataFormatProvider({
...
 
-        if (blocks.length !== 1 && blocks.length !== 2) throw new Error('unknown number of blocks');
+        if (blocks.length !== 1 && blocks.length !== 2 && blocks.length !== 3) throw new Error('unknown number of blocks');
```
dsehnal, line 203:
```suggestion
        if (blocks.length === 0) throw new Error('no data blocks');
```

Now that we are checking if it is a valid volume block, we should only check if there are 0 data blocks.

src/mol-plugin-state/formats/volume.ts
```diff
@@ -197,15 +198,17 @@ export const DscifProvider = DataFormatProvider({
...
 
         const volumes: StateObjectSelector<PluginStateObject.Volume.Data>[] = [];
         let i = 0;
         for (const block of blocks) {
             const entryId = Array.isArray(params?.entryId) ? params?.entryId[i] : params?.entryId;
-            volumes.push(b.apply(StateTransforms.Volume.VolumeFromDensityServerCif, { blockHeader: block.header, entryId }).selector);
+            if (block.categories['volume_data_3d']?.rowCount > 0) {
```
dsehnal, line 209:
```suggestion
            if (block.categories['volume_data_3d_info']?.rowCount > 0) {
```

Let's check for the info category, as technically speaking an empty volume is a valid node and should be added.

dsehnal (commented):
Thanks for your 1st contribution  to Mol*!

src/mol-plugin-state/formats/volume.ts
```diff
@@ -197,16 +198,18 @@ export const DscifProvider = DataFormatProvider({
...
-        const blocks = cifCell.obj!.data.blocks.slice(1); // zero block contains query meta-data
+        const blocks = cifCell.obj!.data.blocks;
 
-        if (blocks.length !== 1 && blocks.length !== 2) throw new Error('unknown number of blocks');
+        if (blocks.length === 0) throw new Error('no data blocks');
 
         const volumes: StateObjectSelector<PluginStateObject.Volume.Data>[] = [];
         let i = 0;
         for (const block of blocks) {
             const entryId = Array.isArray(params?.entryId) ? params?.entryId[i] : params?.entryId;
```
dsehnal, line 208:
```suggestion
            // Skip "server" data block.
            if (block.header.toUpperCase() === 'SERVER') continue;
            
            const entryId = Array.isArray(params?.entryId) ? params?.entryId[i] : params?.entryId;
```

Let's include this check as well.

dsehnal (approved):

PR #446, opened 2022-05-28
Webgl timing support

arose:
image

### DIFF
--- a/src/mol-gl/compute/grid3d.ts
+++ b/src/mol-gl/compute/grid3d.ts
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2020 mol* contributors, licensed under MIT, See LICENSE file for more info.
+ * Copyright (c) 2020-2022 mol* contributors, licensed under MIT, See LICENSE file for more info.
  *
  * @author David Sehnal <david.sehnal@gmail.com>
  */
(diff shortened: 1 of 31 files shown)


src/mol-gl/compute/grid3d.ts
```diff
@@ -191,23 +194,31 @@ export function createGrid3dComputeRenderable<S extends RenderableSchema, P, CS>
 
                 if (spec.cumulative.yieldPeriod && i !== states.length - 1) {
                     if (i % yieldPeriod === yieldPeriod - 1) {
-                        webgl.readPixels(0, 0, 1, 1, array);
+                        webgl.waitForGpuCommandsCompleteSync();
```
dsehnal, line 197: Wasn't the reason for the readPixel that the wait wasn't working propertly?

dsehnal (approved):
Nice.

src/mol-gl/compute/grid3d.ts
arose: `webgl.waitForGpuCommandsComplete` doesn't work well, too slow here

`webgl.waitForGpuCommandsCompleteSync` is just doing the readPixels thing, so the change is mostly cosmetic

PR #447, opened 2022-05-28
use drawbuffers for picking

arose:


dsehnal (approved):
Cool. Does this significantly improve performance for large scenes?

arose:
> Cool. Does this significantly improve performance for large scenes?

for picking it is almost 4x faster, especially good if the scene is vertex shader bound as rendering to a smaller sized buffer doesn't help

PR #452, opened 2022-06-10
DownloadScreenshotControls: componentWillUnmount

simeonborko:
This should solve the error:

Warning: Can't perform a React state update on an unmounted component.
This is a no-op, but it indicates a memory leak in your application.
To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.


dsehnal (approved):
Nice fix, thanks. This actually happens in a few more places in the code, I will add the call there too (tho ideally we would move to hook based code at this point).

PR #466, opened 2022-06-26
Cellpack tweaks

arose:
- Add ``instanceGranularity`` option for marker, transparency, clipping, overpaint, substance data to save memory
- CellPack extension tweaks
    - Use instancing to create DNA/RNA curves to save memory
    - Enable ``instanceGranularity`` by default


arose:
Hey @corredD here are some cellpack updates. With this you can render the ISG without errors. Example https://www.dropbox.com/sh/sbhdy724vziriv5/AAAFe2W4VKtnQgXl_TIs7koha?dl=0

arose:
Hey @corredD here are some cellpack updates. With this you can render the ISG without errors. Example https://www.dropbox.com/sh/sbhdy724vziriv5/AAAFe2W4VKtnQgXl_TIs7koha?dl=0.

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ Note that since we don't clearly distinguish between a public and private interf
 
 ## [Unreleased]
 
+- Add ``instanceGranularity`` option for marker, transparency, clipping, overpaint, substance data to save memory
+- CellPack extension tweaks
+    - Use instancing to create DNA/RNA curves to save memory
+    - Enable ``instanceGranularity`` by default
```
dsehnal, line 12: I think this seems to be missing here. I tried looking for ``instanceGranuality`` in the ``cellpack`` folder and only found ``forceInstanceTheme: true``, but this property doesn't seem to be actually read anywhere.

I also tried to load some cellpack models and it still seems to use the "group" granuality. Am I missing something?

arose: yeah, thanks, missed that when renaming forceInstanceTheme -> instanceGranularity

arose: thinking about making instanceGranularity a quality param

dsehnal: That might be a good improvement. What would the threshold be?

dsehnal (approved):
Very cool, works now.

CHANGELOG.md
arose: Thought about it a bit and decided not to handle it as a quality param. The current quality levels handle systems up to 1,000,000 atoms, everything beyond gets generally the same quality. The ``instanceGranularity`` settings only really starts to make sense for systems with 100,000,000 atoms and more. In short it is quite specific so for now it should be set manually in specific context like for cellpack models.

Issue #361, opened 2022-01-31
Molstar plugin fails the build when imported in a project with isolatedModules set to true in tsconfig


azangru:
When I tried to use the Molstar plugin following the [docs](https://molstar.org/docs/plugin/) (the "PluginContext without built-in React UI" option), I ran up against the following error during the build:

![image](https://user-images.githubusercontent.com/6834224/151825682-6878e64a-18ae-4506-896b-e61ac9e054bc.png)

This, as it turned out, is due to the fact that Molstar uses enums as consts in file exports, and this language feature is not supported by typescript when the compiler is configured to use the `isolatedModules` option.

[Repro case](https://github.com/azangru/molstar-experiments/tree/1c1d92d7e73eda2e5db9d9466f2ff97f9381291c) — a newly scaffolded project based on Vite, which puts `isolatedModules` in the tsconfig by default.

While I understand the attractiveness of const enums — they are compiled away by typescript resulting in a slightly cleaner and smaller javascript file — won't the fact that Molstar can be used as a plugin by third-party projects with different typescript settings serve as a sufficient reason for avoiding this language feature (e.g. by using regular enums)?

**References:**

- [Relevant section in typescript docs](https://www.typescriptlang.org/tsconfig#isolatedModules) about the incompatibility of const enums with the `isolatedModules` setting in tsconfig 
- A [discussion](https://stackoverflow.com/a/62175702/3925302) on Stack Overflow urging library authors not to export const enums
- A [blog post](https://ncjamieson.com/dont-export-const-enums) by an RxJS core team member making the same case of avoiding const enum exports when building a library

arose:
Yeah, we did consider that before but it wasn't much of a problem for us. I think for most `const enum`s we can just use `enum`. It is not so much that we want a smaller js file but we enums in some tight-ish loops. We would need to do some performance test.

arose:
@azangru would you be willing to do a PR that uses `enum` instead of `const enum` throughout? I can do some performance test using it then. I don't know if there are other issues apart from `const enum`s.

@dsehnal any objections to enabling isolatedModules? You did something related to them before, didn't you.

dsehnal:
I've ran into the issue before too, but I just worked around it locally. 

I think we could just find the instances where we use the const enum in the loops and optimize those. 

There might also be more problems with the isolatedModules related to how we use namespaces so that it will require a bit more refactoring.

azangru:
[...]

PR #481, opened 2022-07-10
Refactor away from namespaces for several files

azangru:
First step to the ultimate goal of being able to compile molstar using the `isolatedModules` flag. See https://github.com/molstar/molstar/issues/361#issuecomment-1166641617 for context.

This PR is just a refactoring for several files to move them away from using namespaces.

Ran `npm test`, and `npm run rebuild`, which completed successfully. Did not check whether the compiled code works.

### DIFF
--- a/src/examples/proteopedia-wrapper/helpers.ts
+++ b/src/examples/proteopedia-wrapper/helpers.ts
@@ -16,8 +16,8 @@ export interface ModelInfo {
     preferredAssemblyId: string | undefined
 }
 
-export namespace ModelInfo {
-    async function getPreferredAssembly(ctx: PluginContext, model: Model) {
+export const ModelInfo = {
+    async getPreferredAssembly(ctx: PluginContext, model: Model) {
         if (model.entryId.length <= 3) return void 0;
         try {
             const id = model.entryId.toLowerCase();
@@ -37,16 +37,15 @@ export namespace ModelInfo {
         } catch (e) {
             console.warn('getPreferredAssembly', e);
         }
-    }
-
-    export async function get(ctx: PluginContext, model: Model, checkPreferred: boolean): Promise<ModelInfo> {
+    },
+    async get(ctx: PluginContext, model: Model, checkPreferred: boolean): Promise<ModelInfo> {
         const { _rowCount: residueCount } = model.atomicHierarchy.residues;
         const { offsets: residueOffsets } = model.atomicHierarchy.residueAtomSegments;
         const chainIndex = model.atomicHierarchy.chainAtomSegments.index;
         // const resn = SP.residue.label_comp_id, entType = SP.entity.type;
 
         const pref = checkPreferred
-            ? getPreferredAssembly(ctx, model)
+            ? this.getPreferredAssembly(ctx, model)
             : void 0;
 
         const hetResidues: ModelInfo['hetResidues'] = [];
@@ -78,7 +77,7 @@ export namespace ModelInfo {
             preferredAssemblyId
         };
     }
-}
+};
 
 export type SupportedFormats = 'cif' | 'pdb'
 export interface LoadParams {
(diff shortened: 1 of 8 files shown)


dsehnal:
Hi, thank you for the PR. I will have closer a look at this soon.

dsehnal (commented):
Went through the code. I am not too much opposed to turning most of the namespaces into ``const`` objects, but definitely need to be careful about doing this.

>  Did not check whether the compiled code works.

There is at least one new bug which you introduced in the code (``pos => Vec()`` in the membrane code). If we are going to do this we should probably do it in a way where we test related code with each change (at least by opening a file and seeing if that function works).

src/extensions/anvil/prop.ts
```diff
@@ -36,26 +34,24 @@ interface MembraneOrientation {
...
+    symbols: {
         isTransmembrane: QuerySymbolRuntime.Dynamic(CustomPropSymbol('computed', 'membrane-orientation.is-transmembrane', Type.Bool),
             ctx => {
                 const { unit, structure } = ctx.element;
                 const { x, y, z } = StructureProperties.atom;
                 if (!Unit.isAtomic(unit)) return 0;
                 const membraneOrientation = MembraneOrientationProvider.get(structure).value;
                 if (!membraneOrientation) return 0;
-                Vec3.set(pos, x(ctx.element), y(ctx.element), z(ctx.element));
+                Vec3.set(Vec3(), x(ctx.element), y(ctx.element), z(ctx.element));
```
arose, line 49: `pos` should be reused, e.g., create it at the module/file level

src/extensions/anvil/algorithm.ts
```diff
@@ -560,10 +559,9 @@ namespace HphobHphil {
             }
         }
         return { hphob, hphil };
-    }
-
-    const testPoint = v3zero();
```
arose, line 565: I would create that at the module/file level for reuse, not as part of the object, so it is not accessible elsewhere

arose (commented):
When should we replace namespaces by an object and when by exporting from a (new) file?

src/examples/proteopedia-wrapper/helpers.ts
```diff
@@ -37,16 +37,15 @@ export namespace ModelInfo {
...
+    },
+    async get(ctx: PluginContext, model: Model, checkPreferred: boolean): Promise<ModelInfo> {
         const { _rowCount: residueCount } = model.atomicHierarchy.residues;
         const { offsets: residueOffsets } = model.atomicHierarchy.residueAtomSegments;
         const chainIndex = model.atomicHierarchy.chainAtomSegments.index;
         // const resn = SP.residue.label_comp_id, entType = SP.entity.type;
 
         const pref = checkPreferred
-            ? getPreferredAssembly(ctx, model)
+            ? this.getPreferredAssembly(ctx, model)
```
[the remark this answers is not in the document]
azangru: Ouch! I haven't programmed in this style for a long time, and probably never ever assigned object methods to variables. I even quickly checked the rules of `this` before making that change, but missed this crucial use case. Thank you for spotting it. Fixed as you suggested.

src/extensions/anvil/prop.ts
azangru: After re-reading the code, it seems that `pos` is only used inside of `MembraneOrientation`'s `ctx` function? If so, then declaring the `pos` inside of that function should work?

```
ctx => {
    ...           

    const pos = Vec3();
    Vec3.set(pos, x(ctx.element), y(ctx.element), z(ctx.element));

    ...

    return isInMembranePlane(pos, normalVector, planePoint1, planePoint2);
})
```

azangru:
> Went through the code. I am not too much opposed to turning most of the namespaces into `const` objects, but definitely need to be careful about doing this.
> 
> > Did not check whether the compiled code works.
> 
> There is at least one new bug which you introduced in the code (`pos => Vec()` in the membrane code). If we are going to do this we should probably do it in a way where we test related code with each change (at least by opening a file and seeing if that function works).

My trouble is that I don't know how to test the changes other than running `tsc` for checking the type correctness, and running the test suite to confirm that it still passes 🙂

dsehnal:
> My trouble is that I don't know how to test the changes other than running tsc for checking the type correctness, and running the test suite to confirm that it still passes 🙂

You can run `npm run watch-viewer` and the open/serve build/viewer/index.html locally to use the current build of the viewer.

Issue #486, opened 2022-07-13
Bonds of ligand in `pdb` do not appear


ptourlas:
Greetings,

The viewer fails to render the bonds between the atoms of a ligand in a pdb file.
However, if I isolate the atoms and their corresponding `CONECT` entries in a separate file, the bonds appear normally.
I attach the two files so that you can reproduce.
[pdbPair.zip](https://github.com/molstar/molstar/files/9106579/pdbPair.zip)

Thank you.

dsehnal:
If you give the ligand unique chain id, it will also work, e.g.

```
ATOM   3271  C0  MOL B 219      49.767  38.984  24.904  0.00  0.00           C  
ATOM   3272  C1  MOL B 219      50.467  37.942  24.356  0.00  0.00           C  
ATOM   3273  C2  MOL B 219      49.957  36.636  24.278  0.00  0.00           C  
ATOM   3274  C3  MOL B 219      48.664  36.363  24.662  0.00  0.00           C  
ATOM   3275  C4  MOL B 219      47.901  37.473  25.082  0.00  0.00           C  
ATOM   3276  C5  MOL B 219      48.381  38.785  25.282  0.00  0.00           C  
ATOM   3277  C6  MOL B 219      47.988  35.072  24.334  0.00  0.00           C  
ATOM   3278  O7  MOL B 219      46.764  34.988  24.283  0.00  0.00           O  
ATOM   3279  N8  MOL B 219      48.742  34.058  24.022  0.00  0.00           N  
ATOM   3280  H9  MOL B 219      50.187  39.985  24.965  0.00  0.00           H  
ATOM   3281  H10 MOL B 219      51.443  38.156  23.941  0.00  0.00           H  
ATOM   3282  H11 MOL B 219      50.623  35.967  23.743  0.00  0.00           H  
ATOM   3283  H12 MOL B 219      46.877  37.275  25.392  0.00  0.00           H  
ATOM   3284  H13 MOL B 219      47.774  39.538  25.774  0.00  0.00           H  
ATOM   3285  H14 MOL B 219      48.226  33.219  23.762  0.00  0.00           H  
ATOM   3286  H15 MOL B 219      49.655  34.037  24.460  0.00  0.00           H  
```

----

The issue is that the PDB parser auto-creates entities, and the polymer and ligand end up in different ones. https://github.com/molstar/molstar/pull/487 hopefully fixes this (without breaking anything elsewhere).

ptourlas:
Problem solved!
[...]

PR #487, opened 2022-07-17
struct_conn parsing fix

dsehnal:
- scan all entities when looking for ``struct_conn`` entries
- solves PDB loading issue discussed in #486


arose:
Seems OK to me.

Issue #490, opened 2022-07-19
camera manualReset option is ignored if canvas becomes empty


giagitom:
In the situation in which a representation is updated using an empty structure ( so the canvas becomes empty ), even if I use manualReset option, camera clip space seems to be updated to 0. Is this something wanted?

dsehnal:
I think this might be some edge-case handling going haywire. Maybe you can have a look in `mol-canvas3d/canvas3d.ts` where this logic is implemented and make a PR with a fix?

Alternatively, you can try restoring camera snapshots manually in your code to prevent this.

PR #494, opened 2022-07-20
only update camera state if manualReset is off

giagitom:
Hi,
This pull references issue #490

### DIFF
--- a/src/mol-canvas3d/canvas3d.ts
+++ b/src/mol-canvas3d/canvas3d.ts
@@ -574,7 +574,7 @@ namespace Canvas3D {
             }
             if (oldBoundingSphereVisible.radius === 0) nextCameraResetDuration = 0;
 
-            camera.setState({ radiusMax: scene.boundingSphere.radius }, 0);
+            if (!p.camera.manualReset) camera.setState({ radiusMax: scene.boundingSphere.radius }, 0);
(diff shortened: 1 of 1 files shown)


src/mol-canvas3d/canvas3d.ts
```diff
@@ -571,10 +571,10 @@ namespace Canvas3D {
             if (helper.debug.isEnabled) helper.debug.update();
             if (!p.camera.manualReset && (reprCount.value === 0 || shouldResetCamera())) {
                 cameraResetRequested = true;
+                camera.setState({ radiusMax: scene.boundingSphere.radius }, 0);
             }
             if (oldBoundingSphereVisible.radius === 0) nextCameraResetDuration = 0;
 
-            camera.setState({ radiusMax: scene.boundingSphere.radius }, 0);
             reprCount.next(reprRenderObjects.size);
```
dsehnal, line 578: This changes the semantics and might break some stuff (I think the reason for this extra line was dealing with animating state).

Adding the condition in front of the original line

```ts
if (!p.camera.manualReset) camera.setState({ radiusMax: scene.boundingSphere.radius }, 0);
```

might do the trick better.

dsehnal (commented):
Thanks for the PR, a simple change in the end, tho needs a tweak.

arose:
thanks

PR #496, opened 2022-07-21
Download CCD from Configurable URL

JonStargaryen:
- add CLI arguments that allow downloading the CCD from non-standard URLs
- useful to process CCD data that isn't released yet and for custom CCDs
- technically a breaking change to the signature of `ensureDataAvailable()` -- let me know if this should be done in a less invasive fashion


arose:
It's fine, just add a note the the changelog.

JonStargaryen:
Thanks for the release :)

PR #499, opened 2022-07-23
enable immediateUpdate for iso level

arose:


arose:
@dsehnal any objections? bad idea?

dsehnal:
@arose Hmm, now about at least introducing a ~0.5s delay for the change?

Otherwise I can see this leading to a very unpleasant UX for some cases.

arose:
How would you do that? We don't do for modelIndex update.

dsehnal:
Change ``{ immediateUpdate?: boolean | { debounceTimeMs: number } }`` and then pass this down to the control. The update itself would then be wrapped in a ``BehaviorSubject`` with ``.pipe(debounceTime(...)).subscribe()`` or just manual ``setTimeout``.

dsehnal:
On a second thought, I think we can just do this and implement the debounce later if it causes any problems.

arose:
> On a second thought, I think we can just do this and implement the debounce later if it causes any problems.

:)

PR #507, opened 2022-08-01
Add labels for Confal pyramids

MadCatX:
My original intention was to just add labels to Confal pyramids. In retrospect I realized that some of the naming used in the code was quite misleading so I tried to improve upon that as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ Note that since we don't clearly distinguish between a public and private interf
 
 
 ## [Unreleased]
+- Add custom labels to Confal pyramids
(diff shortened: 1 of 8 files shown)


arose:
Looks good to me and reads cleaner than before.

Unrelated to the changes here but could you add a small example mmcif file to `./examples/` to simplify cursory testing of the confal pyramids?

arose:
Please add a short note to the Changelog even if you just did internal refactoring.

MadCatX:
Thanks! I've added an example file and updated the changelog accordingly. Hopefully I'll have some code less mundane that this to propose in the upcoming weeks...

arose:
Thank you for the changes.

PR #513, opened 2022-08-06
expose inter-bonds props & improve performance

arose:
- Expose inter-bonds compute params in structure
- Improve performance of inter/intra-bonds compute


src/mol-math/geometry/primitives/box3d.ts
```diff
@@ -130,6 +130,14 @@ namespace Box3D {
             v[2] < box.min[2] || v[2] > box.max[2]
         ) ? false : true;
     }
+
+    export function overlaps(a: Box3D, b: Box3D) {
+        return (
+            a.max[0] < b.min[0] || a.min[0] > b.max[0] ||
+            a.max[1] < b.min[1] || a.min[1] > b.max[1] ||
+            a.max[2] < b.min[2] || a.min[2] > b.max[2]
+        ) ? false : true;
```
dsehnal, line 139: I am gonna assume this one is correct :)

```suggestion
        return !(
            a.max[0] < b.min[0] || a.min[0] > b.max[0] ||
            a.max[1] < b.min[1] || a.min[1] > b.max[1] ||
            a.max[2] < b.min[2] || a.min[2] > b.max[2]
        );
```

src/mol-model/structure/structure/structure.ts
```diff
@@ -1180,23 +1204,27 @@ namespace Structure {
...
         for (const unit of structure.units) {
             if (!validUnit(unit)) continue;
 
             const bs = unit.boundary.sphere;
+            Box3D.expand(bbox, unit.boundary.box, rvec);
             Vec3.transformMat4(imageCenter, bs.center, unit.conformation.operator.matrix);
             const closeUnits = lookup.findUnitIndices(imageCenter[0], imageCenter[1], imageCenter[2], bs.radius + maxRadius);
             for (let i = 0; i < closeUnits.count; i++) {
                 const other = structure.units[closeUnits.indices[i]];
+                if (!Box3D.overlaps(bbox, other.boundary.box)) continue;
```
dsehnal, line 1227: Is this more accurate then using a sphere? Since we have the spheres here, would it make sense to do both checks?

dsehnal (approved):

src/mol-math/geometry/primitives/box3d.ts
arose: :)

src/mol-model/structure/structure/structure.ts
arose: Isn't the sphere test already coverred by `lookup.findUnitIndices`?

Doing the box additionally because the box is usually calculated from the sphere3d extrema and often tighter in some ways than the sphere.

PR #514, opened 2022-08-07
Integration of pymol/vmd/jmol transpilers from MolQL project (WIP rasmol transpiler)

yakomaxa:
Integration of MolQL transpilers into Molstar.

# Explanation of this PR
## Imports from previous contribution (see below) and MolQL repository:
(1) vmd/pymol/jmol transpilers were imported from the previous contribution into `src/mol-script/transpilers/` with some modifications
(2) `src/mol-script/transpilers/_spec
` was imported from the previous contribution with some modifications though I do not understand how to deal with these spec things.
(3)` transpile.ts` was taken into src/mol-script/ from MolQL with some modifications.
(4) experimental rasmol traspiler  (WIP) was branched from jmol transpiler. 

## Modifications:
(5) `src/mol-script/script.ts`  : function `toExpression` was modified to use new transpilers. Default values for function `Script `were added. 
(6) `src/mol-util/monadic-parser.ts` : function `seqMap` was implemented so that it can be used by `src/mol-script/transpilers/helper.ts`. The implementation is based on the one in persimmon.  In addition, function `createLanguage` was un-commented out.
(7) `src/mol-plugin-state/transforms/model.ts` : the script-language to be used was hard-coded at line 789 in `StructureSelectionFromScript`. Currently, pymol is selected.

## ToDo (help wanted if possible):
(8) 

- [x] Add a language selector because the selected language is now hardcoded in `model.ts`. 

(9) 

- [x] Manage things about spec tests.

## Miscellaneous:
(10) In `src/mol-script/transpilers/vmd/parser.ts`, function `valuesTest` from `src/mol-script/transpilers/helper.ts` is not working for unknown reasons so 'atom-test' is now hard-coded.
(11) Currently, intersecting selection such as "chain A and residue 10 to 50" fails with Error saying 

> Error: Symbol 'structure-query.modifier.intersect-by' is not implemented.

How can this be fixed? -> Update-2022-08-07: AND and NOT operations were implemented by adding responsible function to `src/mol-script/runtime/query/table.ts`

## Previous contributions:
This PR is based largely on a previous PR by @ptourlas below.
**Integrate pymol/vmd transpilers from MolQL project** https://github.com/molstar/molstar/pull/325
Therefore, author information included in the codes from the branch mentioned in that previous PR is left unmodified.

## Updates 2022-8-12
Finally, I finished implementing and debugging most of the functions used in MolQL transpilers (See below for the process), which allows nearly identical behavior to MolQL explorer.


yakomaxa:
On point (11),  `structure-query.modifier.intersect-by `was implemented by editing `src/mol-script/runtime/query/table.ts`.

yakomaxa:
Related to point (11), `structure-query.generator.query-in-selection` was implemented by editing `src/mol-script/runtime/query/table.ts`. Also `src/mol-script/transpilers/helper.ts` was modified. 

This enables NOT operation.

yakomaxa:
New problems:
(12) Some of operators are not working such as PyMOL within and VMD exwithin.
(13) B.struct.generator.atomGroups() seems not to work correctly, especially when passed no arguments.

yakomaxa:
On point (12), `B.struct.filter.within` seems not working correctly.

arose:
Thank you @yakomaxa, I will have a good look over the next couple of days.

yakomaxa:
**Updates:**
Two functions in `transpilers/helper.ts` were debugged, which enables secondary structure selection in vmd. 

New problems.
(14) The `same` operator for VMD is not working. The cause may be similar to the case of `within`  for PyMOL and `exwithin`  for VMD.  In short, these operators  call functions classified as `Filter` in  `src/mol-script/runtime/query/table.ts`.

yakomaxa:
Memo: PyMOL operators work-or-not-work list
o -> work
x -> not work
un -> unsupported

> 
> o	name: 'not',
> o      name: 'and',
> o       name: 'or',
> x       name: 'in',
> x      name: 'like',
> x       name: 'gap',
> x       name: 'around',
> o      name: 'expand',
> x       name: 'within',
> x       name: 'near_to',
> x       name: 'beyond',
> o      name: 'byresidue',  
> o      name: 'bycalpha',
>  x      name: 'bymolecule',  structure-query.atom-property.topology.connected-component-key' is not implemented
>  un     name: 'byfragment',
>  o      name: 'bysegment', h.asAtom
>  un    name: 'byobject',
>  un      name: 'bycell',
>   x      name: 'byring', structure-query.atom-set.atom-count' is not implemented.
>   o    name: 'neighbour', 
>   o     name: 'bound_to', works but the definition may be different from PyMOL
>   o      name: 'extend', 

The not-working operators all call functions `B.struct.filter.something`

yakomaxa:
New problems:
VMD's keywords such as `protein` and `nucleic `are not working because of '`structure-query.atom-set.property-set` is not implemented'. It seems that  `structure-query.atom-set.somthing` is totally missing from molstar implementation.

**Memo**
`structure-query.generator.something` : working
`structure-query.modifier.something` : working
`structure-query.combinator.something` : working
`structure-query.filter.something` : existing but not working
`structure-query.atom-set.something` :  not existing

yakomaxa:
Updates:
`structure-query.atom-set.atom-count`, `structure-query.atom-set.count-query`, and  `structure-query.atom-set.property-set` were implemented by importing into `src/mol-model/structure/query/queries/atom-set.ts`  from 
`molql/src/reference-implementation/molql/runtime/structure/atom-set.ts`. Current implementation is subset of original one.

Although these function are available now, some operators/keyword like VMD _protein_, PyMOL _byring_, Jmol _protein_ are not working probably because `B.struct.filter.pick` is not working correctly.

yakomaxa:
**Update:**
Finally, PyMOL _within_  operator got enabled!!
The key was to replace `xs['max-radius'] `to `xs['max-radius'](ctx)` in the definition of `MolScript.structureQuery.filter.within` in src/mol-script/runtime/query/table.ts.

Similar tweak will be required for min-radius.
In the same way, I try to fix other problems around `B.struct.filter.something`.

yakomaxa:
Though `B.struct.filter.within` works fine now,  I'm still struggling with fixing other `B.struct.filter.something`.

**B.struct.filter.pick**
- `B.struct.filter.pick` is not working since `pred(ctx)` is always false at the line shown below.
-  https://github.com/molstar/molstar/blob/master/src/mol-model/structure/query/queries/filters.ts#L26

**B.struct.filter.withSameAtomProperties**
- `B.struct.filter.withSameAtomProperties` is not working since the line shown below is evaluated always true.
- https://github.com/molstar/molstar/blob/master/src/mol-model/structure/query/queries/filters.ts#L105

yakomaxa:
Updates
I re-implemented atomCount in atom-set.ts and found`B.struct.filter.pick` was not responsible for the troubles.

yakomaxa:
Now, the problem around `B.struct.filter.something` seems fixed. The rest of the problems are that VMD keyword such as _backbone_ and _protein_  or PyMOL operators like _in_ and _like_ are not working, and some behaves differently from original software (inorganic and bound_to ). The former one would be caused by B.ammp or something like that.

yakomaxa:
**Updates: most of the molql behavior for VMD and PyMOL imported.**

`pick` in `filter.ts` was debugged and now operator/keyword using this function, where `B.ammp `was not responsible. This enables keywords and operators like VMD _protein_.  The key was to change the type definition of arguments of responsible functions.

yakomaxa:
Problem remaining:

- VMD: _same_ operator is unstable
- - not work: `same residue as resname ALA` for any
- - work: `same residue as name FE` for 1TQN
- - not work: `same residue as name FE` for 4HHB
- - work: `same chain as resname ADP` for 1BMF
- - not work: `same chain as resname ANP` for 1BMF
- It seems it fails when selection result in multiple residues or chains

- PyMOL _like_ and _in_ is not working
- - result in null selection.

yakomaxa:
@arose 
Thank you for paying attention to this PR. Now I'm replying to you because I've nearly finished integration of MolQL functionality for PyMOL and VMD. 

To summarize this thread, I made modifications listed below in addition to the ones listed in the beginning of this PR.

- Functions in `src/mol-script/transpilers/helper.ts` and some hardcoded lines in `transpilers/vmd/parser.ts`  were modified to be compatible with mol-script instead of molql-script. For example, structure -> structure-query  property.head -> property.head.name. 

- Some functions were implemented in `src/mol-model/structure/query/queries/atom-set.ts`, which are included in the list Queries in `src/mol-model/structure/query.ts` to enable command classified as `structure-query.atom-set.something`.  

- Some missing functions were implemented, though I just created alias, in `src/mol-script/runtime/query/table.ts`. This enabled _and_ and _not_ operations.

-  As I found type-mismatch caused problems, the type definitions of function arguments in `src/mol-model/structure/query/queries/filters.ts` were changed so that they can take input without casting to any in `src/mol-script/runtime/query/table.ts`. 

As I made many modifications, I hope this will help you to catch what's done here.

I have done nothing about language-selector and _spec yet. Especially for language-selector UI, I would like your advices and helps if possible because I'm not familiar with React .

yakomaxa:
Update:
Added spec and checked they pass.

src/mol-plugin-state/transforms/model.ts
```diff
@@ -786,7 +786,7 @@ const StructureSelectionFromScript = PluginStateTransform.BuiltIn({
     from: SO.Molecule.Structure,
     to: SO.Molecule.Structure,
     params: () => ({
-        script: PD.Script({ language: 'mol-script', expression: '(sel.atom.atom-groups :residue-test (= atom.resname ALA))' }),
+        script: PD.Script({ language: 'pymol', expression: 'all' }),
```
arose: If we really want to we can add a PluginConfig options to change the default here.

src/mol-model/structure/query/queries/atom-set.ts
```diff
@@ -0,0 +1,44 @@
+/**
```
arose, line 1: Not sure if we really need this ported from molql or if there is something like it already in molstar. @dsehnal will know.

src/mol-script/runtime/query/table.ts
```diff
@@ -211,13 +211,13 @@ const symbols = [
...
     D(MolScript.structureQuery.filter.first, (ctx, xs) => Queries.filters.first(xs[0] as any)(ctx)),
-    D(MolScript.structureQuery.filter.withSameAtomProperties, (ctx, xs) => Queries.filters.withSameAtomProperties(xs[0] as any, xs['source'] as any, xs['property'] as any)(ctx)),
+    D(MolScript.structureQuery.filter.withSameAtomProperties, (ctx, xs) => Queries.filters.withSameAtomProperties(xs[0] as any, xs['source'] as any, xs['property'])(ctx)),
     D(MolScript.structureQuery.filter.intersectedBy, (ctx, xs) => Queries.filters.areIntersectedBy(xs[0] as any, xs['by'] as any)(ctx)),
     D(MolScript.structureQuery.filter.within, (ctx, xs) => Queries.filters.within({
         query: xs[0] as any,
         target: xs['target'] as any,
         minRadius: xs['min-radius'] as any,
-        maxRadius: xs['max-radius'] as any,
+        maxRadius: xs['max-radius'](ctx) as any,
```
arose, line 220: @dsehnal do we need to evaluate all of them here or do we generally do that later?

arose (commented):
Hey @yakomaxa, I had a first look at the code and played around with it in the viewer. Looks generally very good, thank you. I amde a few comments and added a few questions to @dsehnal who wrote most of molql and mol-script.

I also took the liberty and implemented a simple language selector for the parameter UI.

package.json
```diff
@@ -90,7 +90,7 @@
     "Panagiotis Tourlas <panagiot_tourlov@hotmail.com>"
```
arose, line 90: Feel free to add yourself as an author here

yakomaxa:
@arose 
Thank you for reviewing and commenting on codes, and questioning @dsehnal about the details of mol-script/molql implementation. I will check them and fix what you pointed out. I also thank you very much about the language selector. I'm so happy to see the selector UI working nicely.

yakomaxa:
Now I start answering the reviwer's comments

src/mol-util/monadic-parser.ts
```diff
@@ -455,6 +475,23 @@ export namespace MonadicParser {
...
+
+    export function of(A: any) {
+        return succeed(A);
+    }
+
+    export function regex(A: any) {
+        return regexp(A);
+    }
+
+    MonadicParser.createLanguage = createLanguage;
```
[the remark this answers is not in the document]
yakomaxa: These were needless so that I deleted them.

src/mol-util/monadic-parser.ts
```diff
@@ -3,11 +3,14 @@
  *
  * @author David Sehnal <david.sehnal@gmail.com>
  */
-
 /**
  * Adapted from Parsimmon (https://github.com/jneen/parsimmon)
  * Copyright (c) 2011-present J. Adkisson (http://jneen.net).
  */
+/**
+*  @author Koya Sakuma
```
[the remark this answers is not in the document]
yakomaxa: The author information was moved to the top and removed the notes on what I did.

src/mol-script/transpilers/rasmol/special_keywords.ts
```diff
@@ -0,0 +1,232 @@
+/**
```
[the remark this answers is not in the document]
yakomaxa: I removed `special_keywords.ts` and `special_operators.ts` because they are not needed.  And I found these filenames were not good, I changed `special_properties` to `macroproperties ` as it's used to define selection macros. 

I can re-change the name if it's better to place hyphen between the words (`macro-properties.ts` rather than macroproperties.ts).

src/mol-plugin-state/transforms/model.ts
yakomaxa: It would be excellent if we can set default language. 

By the way, it was so nice that you placed the language selector where it is now. We can switch languages and perform multi-language selection onto single structure in layer-by-layer manner. Multi-language operation in single viewer is new and interesting experience.

package.json
yakomaxa: Thank you very much. I added myself here.

src/mol-script/runtime/query/table.ts
dsehnal: This depends on the type of the params. The radii need to be evaluated as it expects a number. The query and target are evaluated inside.

src/mol-model/structure/query/queries/atom-set.ts
```diff
@@ -0,0 +1,44 @@
...
+import { QueryContext, QueryFn } from '../context';
+
+
+export function atomCount(ctx: QueryContext) {
+    return (ctx: QueryContext) => {
+        const all = StructureSelection.Singletons(ctx.currentStructure, ctx.currentStructure);
+        const x: number = StructureSelection.structureCount(all);
+        return x;
+    };
+}
```
dsehnal, line 24: You are adding an unnecessary layer of "context" here:

```ts
export function atomCount(ctx: QueryContext) {
  return ctx.currentStructure.elementCount;
}
```

and later

```ts
D(MolScript.structureQuery.atomSet.atomCount,
        function structureQuery_atomset_atomCount(ctx, xs) {
	    return Queries.atomset.atomCount(ctx);
        }),
```

src/mol-script/runtime/query/table.ts
dsehnal, line 220: Oops, guess this and the other ones were an omission by me, need to conditionally evaluate these.

```suggestion
        maxRadius: xs['max-radius']?.(ctx) as any,
```

src/mol-script/runtime/query/table.ts
```diff
@@ -211,13 +211,13 @@ const symbols = [
     // ============= FILTERS ================
     D(MolScript.structureQuery.filter.pick, (ctx, xs) => Queries.filters.pick(xs[0] as any, xs['test'])(ctx)),
     D(MolScript.structureQuery.filter.first, (ctx, xs) => Queries.filters.first(xs[0] as any)(ctx)),
-    D(MolScript.structureQuery.filter.withSameAtomProperties, (ctx, xs) => Queries.filters.withSameAtomProperties(xs[0] as any, xs['source'] as any, xs['property'] as any)(ctx)),
+    D(MolScript.structureQuery.filter.withSameAtomProperties, (ctx, xs) => Queries.filters.withSameAtomProperties(xs[0] as any, xs['source'] as any, xs['property'])(ctx)),
     D(MolScript.structureQuery.filter.intersectedBy, (ctx, xs) => Queries.filters.areIntersectedBy(xs[0] as any, xs['by'] as any)(ctx)),
     D(MolScript.structureQuery.filter.within, (ctx, xs) => Queries.filters.within({
         query: xs[0] as any,
         target: xs['target'] as any,
         minRadius: xs['min-radius'] as any,
```
dsehnal, line 219:
```suggestion
        minRadius: xs['min-radius']?.(ctx) as any,
```

src/mol-script/runtime/query/table.ts
```diff
@@ -211,13 +211,13 @@ const symbols = [
...
-        maxRadius: xs['max-radius'] as any,
+        maxRadius: xs['max-radius'](ctx) as any,
         elementRadius: xs['atom-radius'] as any,
         invert: xs['invert'] as any
```
dsehnal, line 222:
```suggestion
        invert: xs['invert']?.(ctx) as any
```

src/mol-script/runtime/query/table.ts
```diff
@@ -353,6 +357,23 @@ const symbols = [
     D(MolScript.structureQuery.atomProperty.macromolecular.secondaryStructureFlags, atomProp(StructureProperties.residue.secondary_structure_type)),
     D(MolScript.structureQuery.atomProperty.macromolecular.chemCompType, atomProp(StructureProperties.residue.chem_comp_type)),
 
+    // ============= ATOM SET ================
+
+    D(MolScript.structureQuery.atomSet.atomCount,
+        function structureQuery_atomset_atomCount(ctx, xs) {
+	    return Queries.atomset.atomCount(ctx)(ctx);
```
dsehnal, line 364:
```suggestion
	    return Queries.atomset.atomCount(ctx);
```
(together with the change above).

src/mol-model/structure/query/queries/atom-set.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export function countQuery(ctx: QueryContext, query: StructureQuery) {
+    return (ctx: QueryContext) => {
+        const sel = query(ctx);
+        const x: number = StructureSelection.structureCount(sel);
+        return x;
+    };
+}
```
dsehnal, line 32:
```suggestion
export function countQuery(query: StructureQuery) {
    return (ctx: QueryContext) => {
        const sel = query(ctx);
        const x: number = StructureSelection.structureCount(sel);
        return x;
    };
}
```

src/mol-script/runtime/query/table.ts
```diff
@@ -353,6 +357,23 @@ const symbols = [
...
+        }),
+
+    D(MolScript.structureQuery.atomSet.countQuery,
+        function structureQuery_atomset_countQuery(ctx, xs) {
+	    return Queries.atomset.countQuery(ctx, xs[0] as any)(ctx);
```
dsehnal, line 369:
```suggestion
	    return Queries.atomset.countQuery(xs[0] as any)(ctx);
```

(together with the change above)

src/mol-model/structure/query/queries/atom-set.ts
```diff
@@ -0,0 +1,44 @@
...
+
+
+// export function propertySet(ctx: QueryContext, prop: UnitTypeProperties) {
+export function propertySet(ctx: QueryContext, prop: QueryFn<any>) {
```
dsehnal, line 36:
```suggestion
export function propertySet(prop: QueryFn<any>) {
```

src/mol-script/runtime/query/table.ts
```diff
@@ -353,6 +357,23 @@ const symbols = [
...
+        }),
+
+    D(MolScript.structureQuery.atomSet.propertySet,
+        function structureQuery_atomset_propertySet(ctx, xs) {
+	  return Queries.atomset.propertySet(ctx, xs[0] as any)(ctx);
```
dsehnal, line 374:
```suggestion
	  return Queries.atomset.propertySet(xs[0] as any)(ctx);
```

package.json
```diff
@@ -87,10 +87,11 @@
     "Ludovic Autin <autin@scripps.edu>",
     "Michal Malý <michal.maly@ibt.cas.cz>",
     "Jiří Černý <jiri.cerny@ibt.cas.cz>",
-    "Panagiotis Tourlas <panagiot_tourlov@hotmail.com>"
+    "Panagiotis Tourlas <panagiot_tourlov@hotmail.com>",
+    "Koya Sakuma <koya.sakuma.work@gmail.com>"
   ],
   "license": "MIT",
-  "devDependencies": {
+    "devDependencies": {
```
dsehnal, line 94:
```suggestion
  "devDependencies": {
```

package.json
```diff
@@ -132,7 +133,7 @@
     "webpack": "^5.73.0",
     "webpack-cli": "^4.10.0"
   },
-  "dependencies": {
+    "dependencies": {
```
dsehnal, line 136:
```suggestion
  "dependencies": {
```

dsehnal (commented):
Thanks for this PR, heroic work :)

Some comments after going though the code.

yakomaxa:
@dsehnal 
Thank you for detailed review on codes. Your suggestions are very helpful as It was difficult for me to guess how to deal with `contexts` in certain functions. Soon I will adopt your suggestions.

src/mol-model/structure/query/queries/atom-set.ts
yakomaxa: Thank you @dsehnal , I've adopted your suggestions and checked that they works fine.

yakomaxa:
Now adopted all the changes suggested by reviewers.

arose:
Thank you @yakomaxa for adopting the suggestions.

I added a little this is an 'experimental feature' warning to the script param UI so it is easier to justify releasing this.

It would be nice to add the specs https://github.com/MolQL/molql/blob/master/src/transpilers/spec/examples.spec.ts (without the json one) and https://github.com/MolQL/molql/blob/master/src/transpilers/spec/jmol.spec.ts. If there is something special in the rasmol transpiler it would be good to have spec tests for that as well. We could also add a few specs comparing the transpiled expressions to expected builder output.

There are a few commented out pieces of code and console.log statements we should remove before merging this.

arose:
We would also need a Changlog entry mentioning the supported languages and that it is, I suggest, an experimental feature for now.

src/mol-script/runtime/query/table.ts
```diff
@@ -211,15 +211,15 @@ const symbols = [
...
-        minRadius: xs['min-radius'] as any,
-        maxRadius: xs['max-radius'] as any,
+        minRadius: xs['min-radius']?.(ctx) as any,
+        maxRadius: xs['max-radius']?.(ctx) as any,
         elementRadius: xs['atom-radius'] as any,
-        invert: xs['invert'] as any
+        invert: xs['invert']?.(ctx) as any
     })(ctx)),
     D(MolScript.structureQuery.filter.isConnectedTo, (ctx, xs) => Queries.filters.isConnectedTo({
         query: xs[0] as any,
```
dsehnal, line 225: I've realized the ``?.(ctx)`` is missing for ``disjunct`` and ``invert`` parameters here too, can you please update that too here?

dsehnal:
Agree with making this an experimental feature too for now as it's quite a bit of new code which will need testing.

yakomaxa:
@arose @dsehnal 
Thank you @arose for adding warning for transpiled languages. I agree with making this as experimental. It should be tested more.

src/mol-script/runtime/query/table.ts
yakomaxa: I added `?.(ctx)` for `disjunct` and `invert`. 

I also added `?.(ctx)` to `inComplement` in `Queries.generators.querySelection`. Could you check if it is appropriate?

yakomaxa:
@arose 
I've finished removing needless comments and console.log statements. Please check if there's no comments or console.log statements left to be removed. 

I also added example.spec.ts and spec for RasMol. There have not been yet no code to compare transpiled expressions to expected builder output.

A sentence on these transpilers in CHANGELOG was added. Please refine it to a better sentence if needed.

src/mol-model/structure/query/queries/atom-set.ts
```diff
@@ -0,0 +1,39 @@
...
+        const x: number = StructureSelection.structureCount(sel);
+        return x;
+    };
+}
+
+export function propertySet(prop: QueryFn<any>) {
+    return (ctx: QueryContext) => {
+        const set = new Set();
+        const x = getCurrentStructureProperties(ctx, prop, set);
+        //	console.log(x)
```
arose, line 35: please remove console.log

src/mol-model/structure/query/queries/atom-set.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export function atomCount(ctx: QueryContext) {
+    return ctx.currentStructure.elementCount;
+}
+
+
+export function countQuery(query: StructureQuery) {
+    return (ctx: QueryContext) => {
+        const sel = query(ctx);
+        const x: number = StructureSelection.structureCount(sel);
```
arose, line 26: directly return `x`?

src/mol-model/structure/query/queries/atom-set.ts
arose: ping @dsehnal

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ Note that since we don't clearly distinguish between a public and private interf
 - Add custom labels to Confal pyramids
 - Improve naming of some internal types in Confal pyramids extension coordinate
 - Add example mmCIF file with categories necessary to display Confal pyramids
+- Add traspilers for selection scripts that experimentally support PyMOL, VMD, Jmol and RasMol selection syntaxes
```
arose, line 13: [Experimental] Add support for PyMOL, VMD, Jmol and RasMol atom expressions in selection scripts

src/mol-model/structure/query/queries/filters.ts
```diff
@@ -6,7 +6,8 @@
 
 import { SetUtils } from '../../../../mol-util/set';
 import { Unit } from '../../structure';
-import { QueryContext, QueryFn, QueryPredicate } from '../context';
+// import { QueryContext, QueryFn, QueryPredicate } from '../context';
```
arose, line 9: please remove

src/mol-script/transpilers/_spec/utils.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+export function testProperties(properties: PropertyDict, transpiler: Transpiler) {
+    for (const name in properties) {
+        const p = properties[name];
+        p['@examples'].forEach(example => {
+            it(name, () => {
+                if (!p.isUnsupported) {
+                    transpiler(example);
+                    //                  compile(expr);
```
arose, line 36: please remove

src/mol-script/transpilers/_spec/utils.ts
```diff
@@ -0,0 +1,72 @@
...
+    }
+}
+
+export function testOperators(operators: OperatorList, transpiler: Transpiler) {
+    operators.forEach(o => {
+        o['@examples'].forEach(example => {
+            it(o.name, () => {
+                if (!o.isUnsupported) {
+                    transpiler(example);
+                    //                compile(expr);
```
arose, line 58: please remove

src/mol-script/transpilers/all.ts
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright (c) 2017 MolQL contributors, licensed under MIT, See LICENSE file for more info.
+ *
+ * @author David Sehnal <david.sehnal@gmail.com>
+ */
+
+import { transpiler as jmol } from './jmol/parser';
+// import json from './json/parser'
```
arose, line 8: please remove commented out code

src/mol-script/transpilers/helper.ts
```diff
@@ -0,0 +1,410 @@
+/**
+ * Copyright (c) 2017-2021 mol* contributors, licensed under MIT, See LICENSE file for more info.
+ *
+ * @author Alexander Rose <alexander.rose@weirdbyte.de>
+ * @author Panagiotis Tourlas <panagiot_tourlov@hotmail.com>
+ * @author Koya Sakuma <koya.sakuma.work@gmail.com>
+ */
+// import * as Q from 'parsimmon';
```
arose, line 8: please remove

src/mol-script/transpilers/helper.ts
```diff
@@ -0,0 +1,410 @@
...
+            return level.type(level.rule, acc, map);
+        },
+        rule
+    );
+    return x;
+}
+
+export function infixOp(re: RegExp, group: number = 0) {
+    return P.MonadicParser.whitespace.then(P.MonadicParser.regexp(re, group).skip(P.MonadicParser.whitespace));
+    // return P.optWhitespace.then(P.MonadicParser.regexp(re, group).lookahead(P.whitespace))
```
arose, line 134: please remove commented out code (or note why it is useful to keep it)

src/mol-script/transpilers/helper.ts
```diff
@@ -0,0 +1,410 @@
...
+    // return P.MonadicParser.regexp(re, group).skip(P.whitespace
+}
+
+export function prefixOp(re: RegExp, group: number = 0) {
+    return P.MonadicParser.regexp(re, group).skip(P.MonadicParser.whitespace);
+}
+
+export function prefixOpNoWhiteSpace(re: RegExp, group: number = 0) {
+    //    return P.MonadicParser.regexp(re, group).skip(P.MonadicParser.regexp(/\s*/));
```
arose, line 143: please remove

src/mol-script/transpilers/helper.ts
```diff
@@ -0,0 +1,410 @@
...
+    return P.MonadicParser.regexp(re, group).skip(P.MonadicParser.regexp(/\s*/));
+}
+
+export function postfixOp(re: RegExp, group: number = 0) {
+    return P.MonadicParser.whitespace.then(P.MonadicParser.regexp(re, group));
+}
+
+// export function functionOp (re: RegExp, rule: P.MonadicParser<any>) {
```
arose, line 151: please remove

src/mol-script/transpilers/helper.ts
```diff
@@ -0,0 +1,410 @@
...
+        const fs = functions[name];
+        const mapFn = fs.map ? fs.map : makeError(`function '${name}' not supported`);
+        const rule = P.MonadicParser.regexp(new RegExp(name, 'i')).skip(begRule).then(argRule).skip(endRule).map(mapFn);
+        functionsList.push(rule);
+    });
+
+    return functionsList;
+}
+
+// const rule = P.regex(getNamesRegex(name, ps.abbr)).lookahead(lookahead).map(() => {
```
arose, line 307: please remove

src/mol-script/transpilers/jmol/operators.ts
```diff
@@ -0,0 +1,42 @@
...
+ * Adapted from MolQL project
+ */
+
+
+import * as P from '../../../mol-util/monadic-parser';
+import * as h from '../helper';
+import { MolScriptBuilder } from '../../../mol-script/language/builder';
+const B = MolScriptBuilder;
+import { OperatorList } from '../types';
+// import { Expression } from '../../language/expression';
```
arose, line 13: please remove

src/mol-script/transpilers/jmol/parser.ts
```diff
@@ -0,0 +1,256 @@
...
+import { MolScriptBuilder } from '../../../mol-script/language/builder';
+const B = MolScriptBuilder;
+import { properties, structureMap } from './properties';
+import { operators } from './operators';
+import { keywords } from './keywords';
+import { AtomGroupArgs } from '../types';
+import { Transpiler } from '../transpiler';
+import { OperatorList } from '../types';
+
+// const propertiesDict = h.getPropertyRules(properties);
```
arose, line 19: please remove

src/mol-script/transpilers/jmol/parser.ts
```diff
@@ -0,0 +1,256 @@
...
+
+// const slash = P.MonadicParser.string('/');
```
arose, line 21: please remove

src/mol-script/transpilers/jmol/parser.ts
```diff
@@ -0,0 +1,256 @@
...
+    Chainname: () => P.MonadicParser.regexp(/:([A-Za-z]{1,3})/, 1).desc('chainname'),
+    Model: () => P.MonadicParser.regexp(/\/([0-9]+)/, 1).map(Number).desc('model'),
+    Element: () => P.MonadicParser.regexp(/_([A-Za-z]{1,3})/, 1).desc('element'),
+    Atomname: () => P.MonadicParser.regexp(/\.([a-zA-Z0-9]{1,4})/, 1).map(B.atomName).desc('atomname'),
+    Resname: () => P.MonadicParser.regexp(/[a-zA-Z0-9]{1,4}/).desc('resname'),
+    Resno: (r: any) => r.Integer.desc('resno'),
+    Altloc: () => P.MonadicParser.regexp(/%([a-zA-Z0-9])/, 1).desc('altloc'),
+    Inscode: () => P.MonadicParser.regexp(/\^([a-zA-Z0-9])/, 1).desc('inscode'),
+
+    // BracketedResname: function (r) {
```
arose, line 172: can we handle bracketed resnames, like [CA] yet? if not should have a todo somewhere (maybe a future issue)

src/mol-script/transpilers/jmol/parser.ts
```diff
@@ -0,0 +1,256 @@
...
+    //   return P.MonadicParser.regexp(/\.([a-zA-Z0-9]{1,4})/, 1)
+    //     .desc('bracketed-resname')
+    //   // [0SD]
+    // },
+
+    // ResnoRange: function (r) {
```
arose, line 178: can we handle resno ranges yet? if not should have a todo somewhere (maybe a future issue)

src/mol-script/transpilers/pymol/properties.ts
```diff
@@ -0,0 +1,175 @@
...
+ * @author Panagiotis Tourlas <panagiot_tourlov@hotmail.com>
+ * @author Koya Sakuma
+ */
+
+import { MolScriptBuilder } from '../../../mol-script/language/builder';
+const B = MolScriptBuilder;
+import { PropertyDict } from '../types';
+
+const reFloat = /[-+]?[0-9]*\.?[0-9]+/;
+// const rePosInt = /[0-9]+/;
```
arose, line 14: please remove

src/mol-script/transpilers/rasmol/keywords.ts
```diff
@@ -0,0 +1,391 @@
...
+                B.ammp('label_comp_id')
+	  ])
+        }),
+        B.struct.filter.pick({
+	    0: B.struct.generator.atomGroups({
+                'group-by': B.ammp('residueKey')
+	    }),
+	    test: B.core.logic.and([
+                B.core.set.isSubset([
+                // B.core.type.set([ 'P', 'O1P', 'O2P' ]),
```
arose, line 33: please remove

src/mol-script/transpilers/_spec/rasmol.spec.ts
```diff
@@ -0,0 +1,62 @@
+/**
```
arose, line 1: Is there anything in Rasmol expressions that is NOT supported in Jmol (actual Jmol, not the transpiler here which may miss some features)? I am wondering if it is necessary to have both or if Jmol is enough.

src/mol-script/transpilers/vmd/functions.ts
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * @author Alexander Rose <alexander.rose@weirdbyte.de>
+ * @author Panagiotis Tourlas <panagiot_tourlov@hotmail.com>
+ */
+
+import { MolScriptBuilder } from '../../../mol-script/language/builder';
+const B = MolScriptBuilder;
+import { FunctionDict } from '../types';
+
+// import * as P from 'parsimmon'
```
arose, line 12: please remove

src/mol-script/transpilers/vmd/keywords.ts
```diff
@@ -0,0 +1,302 @@
...
+}
+
+function nucleicExpr() {
+    return B.struct.filter.pick({
+        0: B.struct.generator.atomGroups({
+            'group-by': B.ammp('residueKey')
+        }),
+        test: B.core.logic.and([
+            B.core.set.isSubset([
+                // B.core.type.set([ 'P', 'O1P', 'O2P' ]),
```
arose, line 32: please remove

arose (changes requested):

src/mol-script/transpilers/jmol/parser.ts
yakomaxa: Not yet implemented, so I added TODO comment on the top of these comment-out functions.
This is why I made RasMol mode based on PyMOL parser rather than Jmol parser.

src/mol-script/transpilers/_spec/rasmol.spec.ts
yakomaxa: I understand that Jmol supports all expressions RasMol does. As actual Jmol expression seems to be superset of RasMol,  I agree that they should be merged into one as Jmol if possible.

However, as for transpilers, I made RasMol transpiler support expression like [LYS]100:A.CA (<- C-alpha of 100th Lys on chain A ) and residue-range expression, which are left unsupported in Jmol transpiler.  The difference from actual RasMol expression is that residue name has to be in brackets and residue indexes has to be parenthesized. This is because I implemented these in similar manner as PyMOL selection macro using `[]`, `()`, `:` and `. `as keys to make them recognized as macro. In my implementation, RasMol is very similar to PyMOL rather than Jmol.

I tried to integrate these with Jmol expression, but I found it difficult for them to co-exist because Jmol and RasMol expression lacks property names  like _resi_ or _resn_ in PyMOL and _residue_ or _resname_ in VMD, and such raw expressions like _100_ or _LYS_ are confused with keywords.

yakomaxa:
@arose 
Thank you for searching the lines to be removed. There were so many lines I was not able to discover...
Now I believe I removed the comment-out lines suggested and adopted some other suggestions.

src/mol-script/transpilers/_spec/rasmol.spec.ts
yakomaxa: Update: 
Now it's made possible to use un-parenthesized resno-range in RasMol. This should be ported to Jmol.

yakomaxa: I found that my RasMol implementation cannot deal with negative-valued residue index because hyphen is used to represent the index range like 1-200. The problem is that Jmol spec test have an examination of negative-valued residue index, so it's impossible to simply import the implementation of residue range to Jmol from RasMol. 
There are three options:
(1) Import it and change spec of Jmol
(2) Give up importing
(3) Extend residue range selection to negatively valued index (desirable)

yakomaxa: Finally, I made it possible to perform residue range selection including negative-valued index in RasMol. This could be ported to Jmol.

arose:
Thanks @yakomaxa! I will need some time to look at this.

src/mol-script/transpilers/pymol/keywords.ts
```diff
@@ -0,0 +1,272 @@
...
+    'bFCLHMz55tjm16c0': {
+        '@desc': 'nucleic acid',
+        abbr: ['polymer.nucleic'],
+        map: () => B.struct.generator.atomGroups({
+            'residue-test': B.core.set.has([
+                B.core.type.set(ResDict.nucleic),
+                B.ammp('label_comp_id')
+            ])
+        })
+    }
```
arose, line 271: Where do these come from?

arose:
Hey @yakomaxa, went over the PR again, such nice work, thank you!

I implemented the rasmol functionality in the jmol transpiler and took the liberty to remove the rasmol transpiler.

There are quite a few keywords from Jmol and other not yet implemented but we can do that in a separate PR.

It would be good to have some kind of end to end test where we load a structure, run a query and check if the output is as expected. Again, for another PR.

Are there things from you side, @yakomaxa, you think need fixing before merging this?

src/mol-script/transpilers/pymol/keywords.ts
yakomaxa: I intended to mimic polymer.protein and polymer.nucleic by these keywords. 

These keywords are found in PyMOL 2.1 or later and I heavily use when selecting protein and nucleic acid seperately. 
https://pymolwiki.org/index.php/Selection_Algebra

I could not use dots in the key of this keyword dict, so I defined them as random string and make abbr. corresponds to actual keywords. Is there any ways to work this around?

The definition is taken from VMD and modified to fit in PyMOL residue dictionary, but these definition are not strict because they also select  ligands (not-polymer) amino-acid residues such as in amino-acyl-tRNA (like 1ttt). 

There's not problem to remove these keywords if you worry about the definitions are not strict to PyMOL's

yakomaxa:
@arose
Thank you for reviewing codes and merging RasMol to Jmol.

I built you version and found current implementation of Jmol transpiler seems unable to parse some RasMol-like(?) selection command like "19-32:A", which I had enabled in previous RasMol trasnspiler. Unfortunately, I have no idea to re-enable them keeping your way of implementation intact. In addition, I think this lack is trivial for Jmol transpiler at this stage. So, In this PR, there's nothing to add or fix from my side any more. 

As you suggested, let's go on to merge current implementation and continue to another PR to refine and test what we added here. 

By the way, is it acceptable to open another PR to re-implement RasMol transpiler? I came to think I want to make RasMol back  since current Jmol transpiler behave a bit different from RasMol.  I believe RasMol transpiler had almost fully covered RasMol selection syntax, but the implementation was complicated and seems unable to co-exist with your Jmol implementation. So I came to think these two should be implemented as independent transpilers (of course in future PR!).

src/mol-script/transpilers/pymol/keywords.ts
arose: OK, thanks. You can actually use the dots in the keyword. At least it works for me, pushed a change. I think there are close enough for now. It is the same as for the polymer keyword I added initially. One thing I always try it to use the documentation from the PyMOL, Jmol, VMD docs in the transpiler. I changed that too.

yakomaxa: Thank you for modifying the key and adding explanation for PyMOL version.

arose:
Hey @yakomaxa, I just added support for "19-32:A" in the jmol transpiler, see https://github.com/molstar/molstar/pull/514/commits/70fc1a9579c71e54836540f998710ecefef034e5.

I am not in principle opposed to a separate rasmol transpiler, but I need to be convinced it is really required :) To my understanding the jmol atom expressions evolved out of rasmol and should include all (or most?) of it, but I might be mistaken. Do you have some docs for the rasmol syntax?

arose (approved):

yakomaxa:
@arose 
I checked your modification! it works! Thank you very much!

I don't mean to stick to RasMol, and it does not matter whether it’s named RasMol or Jmol as long as it works fine.
The reason I wanted RasMol to back was  I’m not familiar with Jmol and have no means to validate the command.
So if I get confident that Jmol actually supports superset of RasMol atom selection, I will make PR for Jmol transpiler when I notice something missing,  instead of opening PR for RasMol. 

For RasMol transpiler, I read this official manual 
 http://www.openrasmol.org/doc/#chexprs 
and this unofficial one
https://chemapps.stolaf.edu/pe/protexpl/htm/seleccmd.htm
and examined the commands on RasMol 2.7.4.2 on my laptop.

Which Jmol documentation are you referring to? I feel I need to learn about Jmol for the reason above.

arose:
Thank you for the doc links, @yakomaxa. The unofficial one is actually part of Jmol.

For Jmol atom expression docs I look at https://chemapps.stolaf.edu/jmol/docs/#atomexpressions. It has a section called "RasMol biomolecular residue specifications".

You may try Jmol here https://chemapps.stolaf.edu/jmol/jsmol/jsmol.htm. Right clicking the canvas will show a context menu with a "Console" entry. In the console you can execute commands.

arose:
@dsehnal what do you think, merge it?

yakomaxa:
@arose 
Thank you for the reference for Jmol and link for the app. I'll try it.

I found one thing to be fixed in `transpilers/helper.ts`.
I added there functions `prefixRemoveKet` and `prefixOpNoWhiteSpace` for RasMol transpiler, but now they are not referenced from any codes. Do you think these should be removed?

dsehnal:
I am ok with merging this and continue iterating on the functionality in future PRs

arose:
Good point @yakomaxa, I removed the unused functions

yakomaxa:
@arose @dsehnal 
I appreciate your help in this PR. 
This could not be accomplished without your help in implementation of GUI, debugging functions and many other contributions.
Thank you both!

arose:
Thank you @yakomaxa, well done, without your effort and initiative, who knows how much longer the molql code would have been unused!

PR #519, opened 2022-08-11
Allow download of Gzipped files

MadCatX:
As far as I can tell, Molstar Viewer will automatically decompress assets only when come from a local file. There is support for remote Zip archives but getting Gzipped files from a remote source does not seem possible. This little patch adds the option to download Gzipped files.

Is there a reason why there is no direct support for download of compressed structure files etc? I might be missing something obvious...

### DIFF
--- a/src/mol-plugin-state/actions/file.ts
+++ b/src/mol-plugin-state/actions/file.ts
@@ -83,7 +83,7 @@ export const DownloadFile = StateAction.build({
     display: { name: 'Download File', description: 'Load one or more file from an URL' },
     from: PluginStateObject.Root,
     params: (a, ctx: PluginContext) => {
-        const options = [...ctx.dataFormats.options, ['zip', 'Zip'] as const];
+        const options = [...ctx.dataFormats.options, ['zip', 'Zip'] as const, ['gzip', 'Gzip'] as const];
         return {
             url: PD.Url(''),
             format: PD.Select(options[0][0], options),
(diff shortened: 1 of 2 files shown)


src/mol-plugin-state/actions/file.ts
```diff
@@ -96,17 +96,27 @@ export const DownloadFile = StateAction.build({
...
+                    const zippedFiles = await unzip(taskCtx, (data.obj?.data as Uint8Array).buffer);
+                    for (const [fn, filedata] of Object.entries(zippedFiles)) {
+                        if (!(filedata instanceof Uint8Array) || filedata.length === 0) continue;
+
+                        const asset = Asset.File(new File([filedata], fn));
+
+                        await processFile(asset, plugin, 'auto', params.visuals);
+                    }
+                } else {
+                    const url = Asset.getUrl(params.url);
```
arose, line 113: Can you use `getFileInfo` from 'mol-util/file-info.ts' to get the name and ext?

arose (changes requested):
Thank you @MadCatX! Was not pressing to implement this as most files you load over the network are automatically compressed/decompressed while in transfer.

- Did you check if this works with state and session saving?
- Please add a changelog note

src/mol-plugin-state/actions/file.ts
MadCatX: Sure, done.

MadCatX:
> Thank you @MadCatX! Was not pressing to implement this as most files you load over the network are automatically compressed/decompressed while in transfer.
This becomes an issue when PDB/CIF files are provided only on gzipped form. Getting such files loaded then becomes a bit of a hassle...

>     * Did you check if this works with state and session saving?
It does work with session but not with state. When a state is restored, it looks like the remote file is downloaded but not extracted and everything than depends on it then fails. Unless I broke something it seems that remote zip files have the same issue.

src/mol-plugin-state/actions/file.ts
arose: Thanks

arose:
> It does work with session but not with state. When a state is restored, it looks like the remote file is downloaded but not extracted and everything than depends on it then fails. Unless I broke something it seems that remote zip files have the same issue.

OK, thanks for checking. Let's merge this as it is better than before.

PR #520, opened 2022-08-12
Change the lookup logic of NtC steps from residues

MadCatX:
Current code uses a rather clumsy way to match NtC steps with the corresponding residues in the structure. It currently takes the two residues that make up a step, looks up a bunch of information about it and then constructs a "step name" string. The step name is a key in a map that maps step names to indices in the array of all available steps. This is rather slow and involves some crufty code.

Instead of using the step name as the identifier, this patch splits the mapping up per model -> chain -> residue. Since a step is made up of two consecutive residues, the seq_id of the first residue in a step can be used as the key in the map that contains the final mapping of residues to elements the array of steps.
If there are alternate positions, there may be multiple steps with the same seq_id (the altpos property of atoms is "extended" to the entire step), we just map the seq_id to an array of step indices.

This simplifies the code a lot (IMO) and in the worst case scenario of 5MMM reduces the pyramid draw time by ~25 %. I'm almost done with an alternate visualization that will benefit from this new mapping in other ways.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Note that since we don't clearly distinguish between a public and private interf
 - Add custom labels to Confal pyramids
 - Improve naming of some internal types in Confal pyramids extension coordinate
 - Add example mmCIF file with categories necessary to display Confal pyramids
+- Change the lookup logic of NtC steps from residues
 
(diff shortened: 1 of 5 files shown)


arose:
Nice, looks good to me

PR #525, opened 2022-08-18
set some IndexPairBonds maxDistance to Infinity

arose:
- for MOL/SDF and MOL2 (without symmetry) models
- avoid filtering by element-based rules


dsehnal (approved):
LGTM

PR #526, opened 2022-08-19
Add NtC Tube representation

MadCatX:
This adds another representation to visualize NtC conformers. The representation uses the same data as Confal Pyramids. This representation draws a tube through C5' -> O3' -> C5' -> ... atoms of nucleic acids and paints the individual segments of the tube with colors to show what NtCs correspond to different parts of the NA structure. The coloring logic is the same as for Confal Pyramids. Additionally, there are two "markers" on the tube. A Residue Marker is drawn close to the position of the P atom and s Step (or rather Segment) Marker that is drawn at the end of each tube segment.

An example of 4U55 visualized with NtC Tube:
ntc_tube_example

We still have to work out reasonable defaults such as the thickness of the tube, whether to display the markers, etc...

It seems that the Loci marking code does shallow comparison of Loci objects to decide if the currently picked Loci is the same as the previous one and redraws the highlighting wireframe only if the Locis differ. To get this to work properly I need to pre-generate all the DataLoci for the mesh and keep them somewhere. I use a `lociCache` variable that lives in the `CustomProperty` data of the `NtCTubeProvider`. That's the only way I could think if that doesn't break even when I load the same structure multiple times, display multiple models from the same trajectory simultaneously and also doesn't leak memory. The solution of creating DataLocis in `createNtCTubeMesh()` and storing them in what should be treated as read-only data feels rather sketchy though.

Any comments and suggestions will be much appreciated!

Demo of a patched viewer with a few example structures:

https://dnatco.datmos.org/ntc_viewer/
https://dnatco.datmos.org/v4.1/RCSB/cif_dnatco_updated/4u55_v41C35A23.cif.gz
https://dnatco.datmos.org/v4.1/RCSB/cif_dnatco_updated/7k00_v41C35A23.cif.gz
https://dnatco.datmos.org/v4.1/RCSB/cif_dnatco_updated/1hmh_v41C35A23.cif.gz


arose:
Nice new visual.

Could you please create a single behavior that registers all dnatco representations and themes?

Hmm, the DataLoci `data` property is not supposed to be create anew for each loci (with the exception of a single level of object properties). The `elements` should be created for each loci and should refer to something in `data`. That way you only need to have a single number for each thing you want to refer to.

arose:
Thanks @MadCatX ! I will need some time to look at this.

MadCatX:
Sure! Some of the comments need some improvements anyway as I just realized. Thanks!

MadCatX:
This took a little longer than I'd hoped but I have rebased the code against latest Molstar and made a few minor tweaks.

One thing of note:
I came across some peculiar structures that appear to be drawn wrong by this visualization. An example could be https://dnatco.datmos.org/v4.1/RCSB/cif_dnatco_updated/3hl2_v41C35A23.cif.gz. The issue is that this structure contains two models of the NA denoted only by different alt_ids of all atoms of the NA. This confuses the code that computes the NtC annotation data and it puts the steps in the mmCif in a wrong order as a result.
Mesh generator of this visualization cannot deal with "disconnected" steps. I cannot think of a way around this apart from some convoluted heuristics that would put the steps in a proper chemically connected order before drawing them.

I think it should be the responsibility of the code that computes NtC annotations to order the steps correctly. Is it acceptable to just document this issue and declare some structures as broken? I can provide a simpler visualization style for these structures that wouldn't care about wrongly ordered steps but that'd have to go in as another PR.

MadCatX:
Any chance I could get an update on this from you? 😇 At the very least, could we discuss merging the first two patches? They fix a few corner-cases and it'd help me to maintain the NtC Tube visual out-of-tree, if necessary,

src/extensions/dnatco/ntc-tube/representation.ts
```diff
@@ -0,0 +1,458 @@
...
+const v3slerp = Vec3.slerp;
+const v3spline = Vec3.spline;
+const v3sub = Vec3.sub;
+const v3toArray = Vec3.toArray;
+
+const NtCTubeMeshParams = {
+    ...UnitsMeshParams,
+    linearSegments: PD.Numeric(4, { min: 2, max: 8, step: 1 }, BaseGeometry.CustomQualityParamInfo),
+    radialSegments: PD.Numeric(22, { min: 4, max: 56, step: 2 }, BaseGeometry.CustomQualityParamInfo),
+    markResidues: PD.Boolean(true),
```
arose, line 58: This does not seem to have an effect. May it has been superseded by the color option?

src/extensions/dnatco/ntc-tube/representation.ts
```diff
@@ -0,0 +1,458 @@
...
+    residueMarkerWidth: PD.Numeric(0.05, { min: 0.01, max: 0.25, step: 0.01 }),
+    markSegmentBoundaries: PD.Boolean(true),
```
arose, line 60: This does not seem to have an effect. May it has been superseded by the color option?

src/extensions/dnatco/index.ts
```diff
@@ -1,8 +1,8 @@
 /**
- * Copyright (c) 2018-2020 mol* contributors, licensed under MIT, See LICENSE file for more info.
+ * Copyright (c) 2018-2022 mol* contributors, licensed under MIT, See LICENSE file for more info.
  *
  * @author Michal Malý <michal.maly@ibt.cas.cz>
  * @author Jiří Černý <jiri.cerny@ibt.cas.cz>
  */
 
-export { DnatcoConfalPyramids } from './confal-pyramids/behavior';
+export { DnatcoNtCs } from './behavior';
```
arose, line 8: You should add a "[Breaking] note..." in the Changelog

arose:
Is it on purpose that the Pyramids are not highlighted/selected?

src/extensions/dnatco/ntc-tube/representation.ts
MadCatX: Oops, indeed. I removed the props from the MeshParams.

src/extensions/dnatco/ntc-tube/representation.ts
MadCatX: ditto

src/extensions/dnatco/index.ts
MadCatX: I see. Done.

MadCatX:
> Is it on purpose that the Pyramids are not highlighted/selected?

Kind of yes. The pyramids are not really supposed to be used as a standalone visual but in combination with cartoon or ball-n-stick. Having the pyramids "passive" hopefully lowers the risk of misleading the user that they are somehow an actual part of the structure.

arose:
Thanks @MadCatX, merged it.

When you have it in use on your site, please let me know so we can add it to the Community Highlights on molstar.org if you like.

PR #529, opened 2022-08-20
Webgl state

arose:
- Reduce number of webgl state changes
    - Add ``viewport`` and ``scissor`` to state object
    - Add ``hasOpaque`` to scene object
- Handle edge cases where some renderables would not get (correctly) rendered
    - Fix text background rendering for opaque text
    - Fix helper scenes not shown when rendering directly to draw target


dsehnal (approved):
LGTM

PR #531, opened 2022-08-23
add missing depth renderbuffer to wboit pass

arose:
- fix wboit in Safari >=15


dsehnal (approved):
Great! Works in Safari 15.6.1 and 16.0 tech preview on Mac.

PR #533, opened 2022-08-23
Integration of Dual depth peeling - OIT method

giagitom:
This is the initial integration of Dual Depth Peeling algorithm adapted from https://github.com/tsherif/webgl2examples/blob/master/oit-dual-depth-peeling.html

Tested on a linux machine on Chrome and Firefox

### Known Issues ###
- There is a wrong blending on transparent or non-black background


src/mol-canvas3d/canvas3d.ts
```diff
@@ -124,7 +124,8 @@ namespace Canvas3DContext {
         pickScale: 0.25,
         /** extra pixels to around target to check in case target is empty */
         pickPadding: 1,
-        enableWboit: true,
+        enableWboit: false,
+        enableDpoit: true,
```
dsehnal, line 128: Please throw and error in the constructor if both WBOIT and DPOIT are true.

Also, as far as I understand, the DPOIT is quite a bit slower, so we might keep the WBOIT as default

src/mol-canvas3d/passes/dpoit.ts
```diff
@@ -0,0 +1,285 @@
...
+
+    private _supported = false;
+    get supported() {
+        return this._supported;
+    }
+
+    bind() {
+        const { state, gl, extensions : { blendMinMax, drawBuffers } } = this.webgl;
+
+        //initialize
```
dsehnal, line 96: Please fix all the linting errors (``npm run lint-fix`` should do the trick).

dsehnal (commented):
Thanks for the PR!

Some quick feedback (didn't really review the pass itself or the shaders).

- Please add your name to the headers of the files you've modified and to the contributors section of package.json
- Add a line with description of the feature to changelog.md
- For the code you've adapted from https://github.com/tsherif/webgl2examples/blob/master/oit-dual-depth-peeling.html, please add a link/credit to it

----------------------

We might also consider and option to re-create the canvas during runtime so that the transparency method could be changed on the fly. @arose do you have any ideas for this?

giagitom:
@dsehnal I should have done everything. Let me know if something is missing

src/mol-canvas3d/canvas3d.ts
```diff
@@ -124,13 +125,17 @@ namespace Canvas3DContext {
         pickScale: 0.25,
         /** extra pixels to around target to check in case target is empty */
         pickPadding: 1,
-        enableWboit: true,
+        enableWboit: false,
```
dsehnal, line 128: Make the ``wboit`` default please.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ Note that since we don't clearly distinguish between a public and private interf
 
 ## [v3.15.0] - 2022-08-23
 
+- Integration of Dual depth peeling - OIT method
```
dsehnal, line 11: Put this into the "unreleased" section please.

src/mol-canvas3d/canvas3d.ts
```diff
@@ -286,7 +291,8 @@ namespace Canvas3D {
 
     export function create({ webgl, input, passes, attribs, assetManager }: Canvas3DContext, props: Partial<Canvas3DProps> = {}): Canvas3D {
         const p: Canvas3DProps = { ...DefaultCanvas3DParams, ...props };
-
+        console.trace();
+        console.log(props);
```
dsehnal, line 295: Remove logs or only show them in debug mode.

src/mol-gl/scene.ts
```diff
@@ -45,8 +45,8 @@ function calculateBoundingSphere(renderables: GraphicsRenderable[], boundingSphe
 }
 
 function renderableSort(a: GraphicsRenderable, b: GraphicsRenderable) {
-    const drawProgramIdA = (a.getProgram('colorBlended') || a.getProgram('colorWboit')).id;
-    const drawProgramIdB = (b.getProgram('colorBlended') || a.getProgram('colorWboit')).id;
+    const drawProgramIdA = (a.getProgram('colorBlended') || a.getProgram('colorWboit') || a.getProgram('colorDpoit')).id;
+    const drawProgramIdB = (b.getProgram('colorBlended') || b.getProgram('colorWboit') || b.getProgram('colorDpoit')).id;
```
dsehnal, line 49: Thanks for the bugfix here.

dsehnal (commented):
- The tests are still failing, not with shader compilation. Does ``npm run test`` work locally for you?
- A few more things I've noticed on 2nd read
- Can you please please add a param ``enable-dpoit`` support to ``src/apps/index.html`` and pass it to the ``Viewer`` in ``src/apps/app.ts`` 
   - Do it so that when ``enabled-dpoit === '1'`` it disables dpoit and enables this new method. This way we can start testing this feature in the main viewer.

With these change I think we can merge this.

giagitom:
> * The tests are still failing, not with shader compilation. Does `npm run test` work locally for you?

Hmm.. I'm having problems running `npm run test` 

I get this result 
image

dsehnal:
> Hmm.. I'm having problems running `npm run test`

It might be python version error. You might have to create an environment with conda/mamba or venv to fix this. I am sure there will be something about this error on stack overflow.

dsehnal:
@giagitom This is the error (from https://github.com/molstar/molstar/runs/7995859412?check_suite_focus=true#step:7:74)
```
       'frag' shader info log 'ERROR: 0:793: 'tDpoitDepth' : undeclared identifier 
      ERROR: 0:793: 'texture2D' : no matching overloaded function found 
      ERROR: 0:793: 'rg' :  field selection requires structure or vector on left hand side 
      ERROR: 0:793: '=' :  cannot convert from 'const float' to 'highp 2-component vector of float'
      ERROR: 0:794: 'tDpoitFrontColor' : undeclared identifier 
      ERROR: 0:794: 'texture2D' : no matching overloaded function found 
      ERROR: 0:794: '=' :  cannot convert from 'const float' to 'highp 4-component vector of float'
      ERROR: 0:796: 'MAX_DPOIT_DEPTH' : undeclared identifier 
```

Seems like a missed ``#ifdef`` or wrongly declared define.

giagitom:
@dsehnal One error has been solved. 
Have you any suggestions regarding the other two errors? 

https://github.com/molstar/molstar/runs/7995859412?check_suite_focus=true#step:7:42

https://github.com/molstar/molstar/runs/7995859412?check_suite_focus=true#step:7:1012

dsehnal:
> https://github.com/molstar/molstar/runs/7995859412?check_suite_focus=true#step:7:42

You've changed the renderer and added new resources to use there with your new pass. Also, you still have the depth peeling enabled by default in the current code, so these numbers changed

-Make sure the resource isn't created when the new transparency method isn't enabled 
- you still need to change it to enable the original method by default


> https://github.com/molstar/molstar/runs/7995859412?check_suite_focus=true#step:7:1012

The error is displayed further up the log https://github.com/molstar/molstar/runs/7995859412?check_suite_focus=true#step:7:74

giagitom:
> You've changed the renderer and added new resources to use there with your new pass. Also, you still have the depth peeling enabled by default in the current code, so these numbers changed
> 
> -Make sure the resource isn't created when the new transparency method isn't enabled
> 
> * you still need to change it to enable the original method by default

Now depth peeling shouldn't be enabled by default. Can you clarify what do you mean by resource?  I have quite a few shortcomings on the subject.

dsehnal:
> Now depth peeling shouldn't be enabled by default.

Hmm, I must have seen outdated code then.

> Can you clarify what do you mean by resource?

In your case, it creates a different number of attribute buffers. Maybe you added an extra attribute in a shader somewhere which is only used for the depth peeling?

giagitom:
> In your case, it creates a different number of attribute buffers. Maybe you added an extra attribute in a shader somewhere which is only used for the depth peeling?

For dual depth peeling pass I need 7 textures ( 5 more than wboit ) 
I also need two renderables ( 1 more than wboit )
Regarding additional attribues, I'm not aware of it. Could it be due to the additional renderable and the additional vertex shader? The quad vert uses `attribute vec2 aPosition;`

If this is the case should I simply increment the expected value throwing the error by 1?

arose:
Thanks @giagitom! I will need some time to look at this.

giagitom:
I have found an other blending issue on opaque objects (not related with the number of depth peeling passes used)
image

src/mol-gl/shader/chunks/common-frag-params.glsl.ts
```diff
@@ -39,6 +39,12 @@ uniform int uMarkingType;
     #endif
 #endif
 
+#if defined(dRenderVariant_colorDpoit)
+    #define MAX_DPOIT_DEPTH 99999.0
```
arose, line 43: note that this is defined in JS as well

src/mol-gl/shader/direct-volume.frag.ts
```diff
@@ -62,6 +62,12 @@ uniform int uGroupCount;
     uniform sampler2D tMarker;
 #endif
 
+#if defined(dRenderVariant_colorDpoit)
+    #define MAX_DPOIT_DEPTH 99999.0
```
arose, line 66: note that this is defined in JS as well

arose:
Hi @giagitom, this looks very nice, thanks for contributing this!

I had a first look at it and fixed the blending issue by blending to the existing color target, not a separate dpoit blend target.

Also cleaned up the drawbuffer code. You only need call drawbuffer once for each framebuffer at init time.

Fixed the resource counts in the spec.

Haven't looked at transparent back ground and backward compatibility not direct-volumes.

Could you look into using half float buffers for the color targets when available?

arose:
Another look
- Transparent background seems to fine as far as I can tell.
- Tweaked the volume rendering under dpoit a bit - just do standard blending at the end. It wasn't working before and very slow.

arose:
Made some more tweaks, now looking at compat issues. Getting "GL_INVALID_FRAMEBUFFER_OPERATION: Framebuffer is incomplete: Attachments have inconsistent bit plane counts." when using `prefer-webgl1=1`.

arose:
OK, fixed the webgl1 issue. In webgl1 drawbuffers must be in the same format for some reason. Saw the same when implementing the GPU isosurface extraction code.

arose:
@giagitom and @dsehnal please have another look. I think this is good to go. In v4 we can look into changing the transparency method dynamically.

dsehnal (approved):
I think it's ok to merge, thanks for the contibution @giagitom.

giagitom:
Thanks for the help @arose.
I confirm that blending issues on opaque background are now solved. I have found two different issues though. 
- When testing with ambient occlusion off seems that depth-test is disabled (I think this is happening only when postprocessing pass isn't used since If i enable outlines the issue disappears). 
image
- Regarding transaprent background seems that some blending issues are still present
image

arose:
Thanks for testing @giagitom 

> When testing with ambient occlusion off seems that depth-test is disabled

`depthMask` needs to be set to false for dpoit. Added it to the bind methods

> Regarding transaprent background seems that some blending issues are still present

Yeah, basically pre-multiplied alpha was done twice. Changed the logic for dpoit so it is done only in the dpoit code.

giagitom:
Awesome! Thanks. 
It looks good to me. I have noticed an other small issue with transparent background but it isn't related to dpoit. I'ill open an other issue for that. Merge when you want.

Issue #543, opened 2022-08-30
Which is the cheapest way to find the minimum distances between each element of a Structure and a reference Structure?


giagitom:
Hi, Is there a built in function for this?

dsehnal:
Haave a look at how bonds are computed ``https://github.com/molstar/molstar/tree/master/src/mol-model/structure/structure/unit/bonds``.

giagitom:
@dsehnal thanks for pointing to that. In my understanding bonds are computed by finding all elements within a max radius. But in my case I don't have a radius limit. I was thinking about iteratively execute `lookup3d.find(...)` increasing the radius until i find at least one element on the reference structure.. But I'm concerned about the performances of this method.. Maybe there is a better way?

dsehnal:
I see. I think the best course of action then would be adding a ``nearest(x, y, z): T | undefined`` function to ``Lookup3D`` (and ``GridLookup3DImpl``) in ``mol-math/geometry/lookup3d``. It should be relatively straightforward to implement this. Can you make a PR with this and we can iterate from there?

Being even fancier, adding ``distanceTo(other: Lookup3D<T>)`` would make it possible to make an even more optimized version of this.

arose:
@giagitom seems like this is done with your PR merged, right?

giagitom:
@arose yes, it is. I'm closing this

PR #545, opened 2022-09-01
adding nearest and distance to point methods to lookup3d

giagitom:
@dsehnal this is the initial PR addressing #543


src/mol-math/geometry/lookup3d/common.ts
```diff
@@ -41,8 +41,10 @@ export namespace Result {
 export interface Lookup3D<T = number> {
     // The result is mutated with each call to find.
     find(x: number, y: number, z: number, radius: number, result?: Result<T>): Result<T>,
+    nearest(x: number, y: number, z: number): { index: number, squaredDistance: number } | undefined,
+    distanceTo(x: number, y: number, z: number): number,
```
dsehnal, line 45: Instead of this, add function ``distanceToVec(sphere: Sphere3D, v: Vec3)`` to ``namespace Sphere3D``.

src/mol-math/geometry/lookup3d/grid.ts
```diff
@@ -40,6 +40,40 @@ class GridLookup3DImpl<T extends number = number> implements GridLookup3D<T> {
...
+                index = indices[i];
+            }
+        }
+        return { index: index, squaredDistance: nearestDist };
+    }
+
+    distanceTo(x: number, y: number, z: number): number {
+        // distance between a sphere and a point
+        const { center, radius } = this.boundary.sphere;
+        return Vec3.distance(Vec3.create(x, y, z), center) - radius; // if negative, point is inside sphere
```
dsehnal, line 74: Don't create temporary ``Vec3`` like this. This cause an allocation of an array on each function call, which is very expensive if you do that often. Instead, create a global ``tmpVec`` which you set the position to (this pattern is used often in the code).

src/mol-math/geometry/lookup3d/grid.ts
```diff
@@ -40,6 +40,40 @@ class GridLookup3DImpl<T extends number = number> implements GridLookup3D<T> {
         return ret;
     }
 
+    nearest(x: number, y: number, z: number): { index: number, squaredDistance: number } | undefined {
```
dsehnal, line 43: Maybe this function should be ``nearest(x: number, y: number, z: number, k: number): Result<T>`` instead and return ``k``-nearest points. This will make it possible to find a nearest atom to another one within a structure (by setting ``k=2`` and taking the further one).

src/mol-math/geometry/lookup3d/grid.ts
```diff
@@ -40,6 +40,40 @@ class GridLookup3DImpl<T extends number = number> implements GridLookup3D<T> {
...
+        this.ctx.radius = startingRadius < 0 ? radiusIncrement : startingRadius + radiusIncrement;
+        this.ctx.isCheck = false;
+        const result = this.result;
+        query(this.ctx, result);
+        while (!result.count) { // necessary to check if grid is empty to avoid infinite loop
+            this.ctx.radius += radiusIncrement;
+            query(this.ctx, result);
+        }
+        const { indices, squaredDistances } = result;
+        let index = indices[0], nearestDist = squaredDistances[0];
```
dsehnal, line 59: Instead of this, how about you find the voxel the point is in and then recursively visit neighboring cells to find the closest point? You always have to visit at least 1 layer of closest voxels. If you don't find the closest point there, you increase the threshold. There are some clever ways to do this too (like you can keep queue of voxels you need to visit represented as integer indices to avoid array allocations).

src/mol-model/structure/structure/util/lookup3d.ts
```diff
@@ -86,6 +86,53 @@ export class StructureLookup3D {
...
+                return { index: nearestIndex!, unit: nearestUnit!, squaredDistance: nearestDist };
+            }
+        }
+    }
+
+    distanceTo(x: number, y: number, z: number): number {
+        // distance between a sphere and a point
+        const { center, radius } = this.boundary.sphere;
+        return Vec3.distance(Vec3.create(x, y, z), center) - radius; // if negative, point is inside sphere
+    }
```
dsehnal, line 134: Remove this function too.

src/mol-model/structure/structure/util/lookup3d.ts
```diff
@@ -86,6 +86,53 @@ export class StructureLookup3D {
...
+
+        if (!elementCount) return undefined;
+
+        const radiusIncrement = 0.1; // how to choose a good increment?
+        const startingRadius = this.distanceTo(x, y, z);
+        let radius = startingRadius < 0 ? radiusIncrement : startingRadius + radiusIncrement;
+
+        while (true) {
+            const closeUnits = this.unitLookup.find(x, y, z, radius, ctx.closeUnitsResult);
+            radius += radiusIncrement;
```
dsehnal, line 105: I am not sure this is a good approach which will work in general. Maybe a better one would be to sort the units based on distance to the query point and then query the units in that order, keeping the distance to the closest point. Then once the distance starts to be greater than the current closest points, you can stop the iteration.

src/mol-model/structure/structure/util/lookup3d.ts
```diff
@@ -86,6 +86,53 @@ export class StructureLookup3D {
...
+            if (closeUnits.count) {
+                let nearestIndex: number, nearestUnit: Unit, nearestDist = Number.MAX_SAFE_INTEGER;
```
dsehnal, line 107: I think you want ``Number.MAX_VALUE`` here.

src/mol-model/structure/structure/util/lookup3d.ts
```diff
@@ -86,6 +86,53 @@ export class StructureLookup3D {
...
+                for (let t = 1, _t = closeUnits.count; t < _t; t++) {
+                    const unit = units[closeUnits.indices[t]];
+                    Vec3.set(this.pivot, x, y, z);
+                    if (!unit.conformation.operator.isIdentity) {
+                        Vec3.transformMat4(this.pivot, this.pivot, unit.conformation.operator.inverse);
+                    }
```
dsehnal, line 113: This can be outside of the loop.

src/mol-math/geometry/lookup3d/grid.ts
giagitom: > Instead of this, how about you find the voxel the point is in and then recursively visit neighboring cells to find the closest point? You always have to visit at least 1 layer of closest voxels. If you don't find the closest point there, you increase the threshold. 

Yes, I could do something like this. But in case the point is outside the grid this could lead to problems in my understanding. Expanding from the nearest voxel/s instead of the original center could exclude nearer points from the result. Should we use two different methods in case the point is inside or outside the grid?

dsehnal: Expanding from the nearest voxel should work as well as you can expand in a way similar to how you expand your radius here. You just keep the distance to the furthest current included point and based on the voxel position (e.g. center of the voxel, you can stop the search).

giagitom: @dsehnal i tried to draw what I mean. This is a 2D example but I guess it should be the same in 3D. Also please let me know if there is something conceptually wrong (for example I am unsure about the boundary sphere radius)

There are two expansions in gray, one from the external point origin and the other from the nearest voxel. In this configuration the nearest point respect to the nearest voxel isn't the same of the external point expansion.

image

dsehnal: The method I am suggesting will still work. You do not stop searching the 1st time you find a point, you stop searching when all voxels that have not been visited are further away then the closest point you've found:

image

You can do this using breath-first search (with termination condition also based on the on the currently found closest point) of the voxels starting from the closest one.

dsehnal: To clarify a bit more, you check against the distance to the query point, not distance to the closest voxel.

giagitom: Ok thanks, I think I understood now. 

Regarding the way to find the nearest voxel do you have any ideas? Should I evaluate the position of each voxel and keep the nearest one? Or maybe something else?

Instead regarding the layer expansion I was thinking to keep track of already visited voxels ( in a map maybe? ) and move the current layer voxels in each direction avoiding already visited voxels. Is this a good approach?

dsehnal: I think just doing ``min(max(0, floor(x / cell_size_x)), n_x)`` for each dimension for the initial cell should be a reasonable estimate.

For points outside the bounding sphere, doing intersection between the ray going from the point to the center of the bounding sphere and the sphere and "snap" that point to the closest cell from the step above might provide a better estimate.

Even a better estimate for outside points could be doing intersection between the faces of the grid box and the ray from B. Although this might be somewhat harder to compute.

giagitom: @dsehnal with the last commit I implemented a method nearestPointToVec in box3D. It should return the very nearest point in the box respect to a random vector. This is the concept:
- The method first searches for the point to be inside the box
- If not, it makes projections on all planes of the box. If projection is inside the box (on the surface of the box) the distance and the vector are kept to make comparisons.
- If at least one projection is on the surface of the box it returns the projection with minimum distance from the point.
- If no projections on planes are within the box boundaries, it makes projections on all 12 segments of the box and evaluate distance only if projection is inside the segment like before.
- If everything fails it looks at all vertices of the box and returns the vertex with minimum distance.

dsehnal: I think this is too complicated. All you need is an intersection between a ray from the query point and the center of the bounding sphere/box. Here is a simple method https://gamedev.stackexchange.com/a/18459. If this fails, you can simply use the closest vertex of the box as a starting point.

giagitom: >Even a better estimate for outside points could be doing intersection between the faces of the grid box and the ray from B. Although this might be somewhat harder to compute.

I was just trying to apply this method you mentioned. But searching online I realized it was necessary to search also for projections on edges and on vertices.


>Here is a simple method https://gamedev.stackexchange.com/a/18459. If this fails, you can simply use the closest vertex of the box as a starting point.

Sorry, this is not my field.. a few questions regarding this method.
- It isn't clear to me how to get the intersection point from this?
- I'm not sure how could it fail since the ray is directed towards the bounding box center?

giagitom: Also an other question, 
Does molstar always uses AABB boxes type? Even for instances with transformed coordinates for example?

dsehnal: > I'm not sure how could it fail since the ray is directed towards the bounding box center?

Yeah, this can't happen (but I am always paranoid about these things).

> Does molstar always uses AABB boxes type? Even for instances with transformed coordinates for example?

You query the transformed units by applying an inverse transform on the query point, so it's always AABB.

giagitom:
Hi @dsehnal ,
I added `nearest` method at grid and structure level. It finds the nearest k elements from a point. 
I also implemented ray/box intersection method on `Box3D` to find the nearest point of the grid.

giagitom:
@dsehnal have you been able to look into this by chance? I tested this a bit and it seems to work fine

dsehnal:
Sorry didnt have time yet, I will try to find some soon.

giagitom:
Ok no problem! Thanks

src/mol-math/geometry/lookup3d/grid.ts
```diff
@@ -277,4 +289,123 @@ function query<T extends number = number>(ctx: QueryContext, result: Result<T>):
...
+            for (let ix = -1; ix <= 1; ix++) {
+                const xPos = gX + ix;
+                if (xPos < 0 || xPos >= sX) continue;
+                for (let iy = -1; iy <= 1; iy++) {
+                    const yPos = gY + iy;
+                    if (yPos < 0 || yPos >= sY) continue;
+                    for (let iz = -1; iz <= 1; iz++) {
+                        const zPos = gZ + iz;
+                        if (zPos < 0 || zPos >= sZ) continue;
+                        nextArrG.push(xPos, yPos, zPos);
```
dsehnal, line 382: Check here also that the cell you are adding wasn't visited yet?

src/mol-model/structure/structure/util/lookup3d.ts
```diff
@@ -39,6 +41,25 @@ export namespace StructureResult {
...
+            units.length = count;
+        }
+        for (let i = 1, c = result.count; i < c; i++) {
+            if (arrayLess(squaredDistances, i - 1, i) > 0) {
+                arraySwap(squaredDistances, i - 1, i);
+                arraySwap(indices, i - 1, i);
+                arraySwap(units, i - 1, i);
+                i = Math.max(0, i - 2);
+            }
+        }
```
dsehnal, line 61: What kind of sort is this? What you are looking for is a structure called heap (binary could be enough, but perhaps fibonacci better) to keep the list of the nearest points.

src/mol-math/geometry/lookup3d/grid.ts
```diff
@@ -277,4 +289,123 @@ function query<T extends number = number>(ctx: QueryContext, result: Result<T>):
...
+    } else {
+        gX = Math.floor((x - minX) / delta[0]);
+        gY = Math.floor((y - minY) / delta[1]);
+        gZ = Math.floor((z - minZ) / delta[2]);
+    }
+    let gCount = 1, nextGCount = 0, arrG = tmpArrG1, nextArrG = tmpArrG2, prevFurthestDist = Number.MAX_VALUE, prevNearestDist = -Number.MAX_VALUE, nearestDist = Number.MAX_VALUE, findFurthest = true, furthestDist = - Number.MAX_VALUE, distSqG: number;
+    arrG.length = 0;
+    arrG.push(gX, gY, gZ);
+    tmpMapG.clear();
+    while (result.count < k && result.count < indicesCount) {
```
dsehnal, line 324: This might work, but the implementation feels somewhat weird. There should be a binary/fibonacci heap (or an equivalent priority queue) that keeps the k nearest items.

dsehnal (commented):
It's a nice progress, but I don't think it's "there" yet.

src/mol-model/structure/structure/util/lookup3d.ts
giagitom: I have found this project, https://github.com/gwtw/ts-fibonacci-heap do you think this could work for our use case?

dsehnal: It might and since it's MIT license we can also clone the code into the `mol-util` package (with credits in the file header) instead of relying on this as a dependency.

Tho I am a bit worried about their use of the iterator object as that is quite slow to do the allocations. But that could be improved.

dsehnal: Also, we need to be careful about the size of the heap as this implementation has O(n) complexity for this operation.

giagitom:
@dsehnal I implemented the nearest search with fibonacci heap on grid and lookup3d. There is a huge improvement when k is high. 

>Also, we need to be careful about the size of the heap as this implementation has O(n) complexity for this operation so we might have to track it separately.

I'm also avoiding using heap size in the implementation.

src/mol-util/fibonacci-heap.ts
```diff
@@ -0,0 +1,388 @@
+/**
+ * Copyright (c) 2018-2022 mol* contributors, licensed under MIT, See LICENSE file for more info.
```
dsehnal, line 2:
```suggestion
 * Copyright (c) 2022 mol* contributors, licensed under MIT, See LICENSE file for more info.
```

src/mol-util/fibonacci-heap.ts
```diff
@@ -0,0 +1,388 @@
...
+
+    constructor(key: K, value?: V) {
+        this.key = key;
+        this.value = value;
+        this.prev = this;
+        this.next = this;
+    }
+}
+
+class NodeListIterator<K, V> {
```
dsehnal, line 35: Can you add a `reset(node)` method here and use a static instance of this where possible?

dsehnal (commented):
Good job!

To wrap this up:
- Can you also port the tests for the fib heap to a `.spec.ts` file like in the rest of the code?
- Write at least some simple tests for the nearest functionality?
- See if the static iterator with reset helps with the fib heap?

src/mol-util/fibonacci-heap.ts
giagitom: >See if the static iterator with reset helps with the fib heap?

I have made some tests with and without static iterator and seems that there is no difference when making a search with k=100.000 (about **300ms**). 
Instead there is a worsening in performances when making 100.000 different random coordinate nearest searches with k=1 (**18.8 seconds** with static iterator vs **16.7 seconds** without). 

Do you think this is something strange? 
I implemented static iterator reset method i this way 
```ts
public reset(start: Node<K, V>) {
        this._index = -1;
        this._items = []; // using this._items.length = 0 leads to about 1 second performances worsening 
        let current = start;
        do {
            this._items.push(current);
            current = current.next;
        } while (start !== current);
    }
```
And initilaized it like this.
```ts
const tmpIt = new NodeListIterator<any, any>();
```

dsehnal: That seems not such a big improvement. The different would be likely that in your implementation with k=1, the iterator just gets created more often. But ~18s for k=1 100k queries seems to be quite slow. I reckon you can optimize it to be at least 10x faster.

giagitom:
Now 100.000 random coords nearest search with k=1 are 4x to 15x faster depending on structure dimensions and number of units.

arose:
Nice speedup!

What is the range of k? I wonder for what k using a fib heap is better than just an array of length k that is kept sorted.

How does the Bottom-Up list look like in the Chrome Performance tab, that is, where is the most time spent?

giagitom:
@dsehnal I further improved performances for k=1 situations where is unnecessary to use heap.extractMinimum(). I think as well this could be further improved but for now I'm happy as it is and it could be merged. I'll test this further in next days to be sure it works as expected. 

@arose k can range from 1 to >= structure.elementCount

here are the results from Performance tab:
- For 100.000 nearest searches with k=1 on entry 4V99 (300 units). Here `extractMinimum` happens only at units search time.  
image

- For a single nearest search with k = 100.000 on entry 4V99 (300 units). Pretty fast I think?
image

dsehnal:
@giagitom wanna merge this now and possible make further changes in future PRs?

giagitom:
@dsehnal yes thanks. I forgot to tell I already mentioned 
this under unreleased section on change log.

PR #548, opened 2022-09-03
add parentDisplay param to interactions repr

arose:
Only has an effect when "includeParent" is enabled. "Stub" shows just the child side of interactions to the parent. "Full" shows both sides of interactions to the parent. "Between" shows only interactions to the parent.

The way to use it is to create a structure selection that only has the ligand you want to see interactions to and then set the parentDisplay to "between".


dsehnal (approved):
LGTM

dsehnal (approved):
long live copy paste bug fixing :)

PR #552, opened 2022-09-09
add support for failIfMajorPerformanceCaveat

arose:
Support for ``failIfMajorPerformanceCaveat`` webgl attribute. Add ``PluginConfig.General.AllowMajorPerformanceCaveat`` and ``allow-major-performance-caveat`` Viewer GET param.

With this PR, the plugin (and Viewer) will fail with a webgl not available error if no hardware acceleration is available, i.e., GPU emulated with software renderer, which makes molstar unusable slow in almost all cases.

Chrome has a setting "Use hardware acceleration when available" which can be used to test this.

When using molstar to render images server-side a GPU is often not available and a software renderer is OK, which is still support with the new options.

### DIFF
--- a/src/mol-plugin-ui/controls/screenshot.tsx
+++ b/src/mol-plugin-ui/controls/screenshot.tsx
@@ -1,7 +1,8 @@
 /**
- * Copyright (c) 2020 mol* contributors, licensed under MIT, See LICENSE file for more info.
+ * Copyright (c) 2020-2022 mol* contributors, licensed under MIT, See LICENSE file for more info.
  *
  * @author David Sehnal <david.sehnal@gmail.com>
+ * @author Alexander Rose <alexander.rose@weirdbyte.de>
  */
 
 import * as React from 'react';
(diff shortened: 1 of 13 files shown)


src/mol-plugin-ui/controls/screenshot.tsx
```diff
@@ -151,18 +152,18 @@ function drawPreview(helper: ViewportScreenshotHelper, target: HTMLCanvasElement
...
 function ViewportFrame({ plugin, canvas, color = 'rgba(255, 87, 45, 0.75)' }: { plugin: PluginContext, canvas: HTMLCanvasElement | null, color?: string }) {
     const helper = plugin.helpers.viewportScreenshot;
-    const params = useBehavior(helper?.behaviors.values!);
-    const cropParams = useBehavior(helper?.behaviors.cropParams!);
-    const crop = useBehavior(helper?.behaviors.relativeCrop!);
+    if (!helper || !canvas) return null;
+
+    const params = useBehavior(helper.behaviors.values);
+    const cropParams = useBehavior(helper.behaviors.cropParams);
+    const crop = useBehavior(helper.behaviors.relativeCrop!);
```
dsehnal, line 159: This is not correct (in general). All hooks need to be executed before any branching. That's why the ``useBehavior`` supports undefined argument (otherwise you would be just able to do ``const x = beh ? useBehavior(beh) : undefined``.

PR #553, opened 2022-09-09
fix handling of PDB TER records

arose:


src/mol-model-formats/structure/pdb/atom-site.ts
```diff
@@ -91,7 +91,7 @@ export function getAtomSite(sites: AtomSiteTemplate, hasTer: boolean): { [K in k
         if (asymIdCounts.has(asymId)) {
             // only change the chains name if there are TER records
             // otherwise assume repeated chain name use is from interleaved chains
-            if (hasTer && asymIdChanged) {
+            if (terIndices.size > 0 && asymIdChanged || terIndices.has(i)) {
```
dsehnal, line 94:
```suggestion
            if ((terIndices.size > 0 && asymIdChanged) || terIndices.has(i)) {
```

Alternatively, would it now be enough to just check for ``terIndides.has(i)`` ?

arose: yeah, should be good

PR #563, opened 2022-09-21
Module Load Performance Markers

dsehnal:
Renders module load marker's in `embedded.html`:

```
npm run build
```
+ open ``build/viewer/embedded.html``

image


dsehnal:
Thanks a lot @awaterho !

According to the data, the entire evaluation of the scripts takes ~800ms and based on your other tests, creating instance of the plugin is also very fast & you can actually display zika without problems. So this seems the slowdown happens during parsing of the script... any ideas here?

dsehnal:
@awaterho I see, I will make another pr which prints performance markers to the web page directly for that load function so that we can identify the bottleneck.

awaterho:
I narrowed it down to this line:           
https://github.com/molstar/molstar/blob/2f86f5d758e4561b8a2319d199624ed5dd2417c3/src/mol-util/assets.ts#L106

The thing is , it is incredibly slow the first time (>1minute), but after that I can load  other models without problems????

dsehnal:
That is weird indeed. 

Can you try calling just this function without any boiler plate around and perhaps add some performance markers to the implementation of the ajaxGet function? I can help with that and create a PR if you want.

Issue #560, opened 2022-09-14
Error when canvas element is hidden entirely


russellp17:
This is a pretty minor issue but wanted to see if there's a quick fix. For users viewing a structure on a small window, they may cover the canvas element entirely, especially when opening Mol* sidebars which results in a `Error: empty textures are not allowed` error:

<img width="818" alt="Screen Shot 2022-08-10 at 5 30 33 PM" src="https://user-images.githubusercontent.com/71293199/190192524-a3a131d3-52a3-4ccf-ad97-58ec9cd35a2d.png">

dsehnal:
Thanks for the report, this might be an easy fix in the `canvas3d.ts:handleResize` function (or the functions it is calling) to make the minimum size say (2, 2). Would you create a PR for this?

russellp17:
Yep, I can put up a fix soon

PR #569, opened 2022-09-27
Fix "empty textures" error on empty canvas

russellp17:
Fixes an issue where hiding the canvas element would trigger an error. Used 2x2 as the minimum dimensions as discussed in the issue: https://github.com/molstar/molstar/issues/560


dsehnal (approved):
Thanks

Issue #577, opened 2022-10-05
Configurable carbon color based on structure


jpattle:
How might I implement a custom `ColorTheme` similar to Element Symbol that instead bases the carbon color on what structure the representation belongs to? That way everything showing for a single structure (polymer, ligand, carbohydrate, non-standard, focus target, focus surroundings, etc) would have the same carbon color. Ideally this would be configurable at the structure level, so that if you changed the color assigned to the structure all representations for that structure would automatically update.

Example:

- I load in 2 structures, 3FGR and 129L
- I want to set 3FGR's carbon color to blue for the whole structure (both chains)
- Similarly I want to set 129L's carbon color to purple
- Any new components I add to 3FGR or 129L should be have the same carbon color as the rest of the structure
- If I update the carbon color for 3FGR from blue to green then all the components with this new custom color theme should update

Is something like this possible? So far I've made a `ColorTheme` based on Element Symbol to set the carbon color to a single color, but I'm not sure how to tie that to the structure, or how to make it so that if the color for the structure updates it updates all representations to match.

Thanks in advance for your help!

dsehnal:
You can use a custom property which defines the carbon color at the model level and a theme which pulls the color from there.

arose:
There already is a `CustomModelProperty` called `Model.Index` that gets added by the `StructureInfo` behavior.

Maybe the model-index color theme could be extended to use the `Model.Index` in addition to `Model.TrajectoryInfo`. With that the model-index theme could become an option for the carbonColor in the element-symbol theme. I think that could be a nice contribution.

jpattle:
@arose happy to contribute. How would I update the `model-index` color theme to use `Model.Index` in addition to `Model.TrajectoryInfo`? As far as I can tell it is coloring based off trajectory index and does not appear as an option in the UI if there's only one trajectory for the model. Would it be better to create another color theme, like `structure-index` that works similarly to `model-index` but colors based on model index instead of trajectory index, then add that as a carbon color option?

In addition, I noticed that the carbon color in `element-symbol` is not adjusted by the `props.saturation` and `props.lightness` values whereas the other element colors are. Is this intentional, or should I update this as a part of this enhancement (should be a quick fix)?

arose:
[...]

PR #583, opened 2022-10-11
Model index updates & carbon color

jpattle:
Based off discussion for [molstar-577](https://github.com/molstar/molstar/issues/577)

* Renamed the existing model-index color theme to trajectory-index
* Created a new model-index color theme that uses `Model.Index` to determine color
* Added the new model-index color theme as an option in the element-symbol color theme for carbon color
* Added the option to alter the element-symbol's carbon color by the saturation/lightness parameters
    * If we feel this change is not needed then I can remove it

Contains a 'breaking' change in that the model-index theme no longer works at all like it used to.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ Note that since we don't clearly distinguish between a public and private interf
 
 ## [Unreleased]
 
+- [Breaking] Renamed the ``model-index`` color theme to ``trajectory-index``
+- Added a new ``model-index`` color theme that uses the `Model.Index` property
+- Added the new ``model-index`` color theme as an option for the carbon color in the ``element-symbol`` and ``ilustrative`` color themes
+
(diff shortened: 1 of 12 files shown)


arose:
Thanks @jpattle, a few requests

- please add a few lines to the changelog, including for the rename prefixed with [Breaking]
- add your name to the top of the file you added
- please add `model-index` as an option to the `illustrative` theme
- if you would actually use `adjustCarbonColor` I am ok with adding it, if not I suggest removing it

arose:
Thank you @jpattle, much appreciated. Made some small tweaks to support color palettes that require a count. I will go ahead and quickly add a `structure-index` theme as well, which can be useful if you create a structure from multiple models, e.g., a receptor and a ligand.

arose:
Added ``structure-index`` 1886d9d72f6d67464bcd621698efb4faadc2f32c

Issue #582, opened 2022-10-11
Component controls collapsing when any option is selected


arussell123:
Hi,

When components are edited using the `Add Representation`, `Set Coloring` or `Modify by Selection` options in the `Components` section of `Structure Tools` the component that is being edited automatically collapses after making a selection.
This is often confusing to some of our the users so would it be possible to keep the expanded component and option open. 

In the codebase in [`src/mol-plugin-ui/structure/components.tsx`](https://github.com/molstar/molstar/blob/6ed42e9521b60d25b4a02803ea821b5d40a43c67/src/mol-plugin-ui/structure/components.tsx#L283) it collapses because `action` in the state is changed from 'action' to undefined (line 283) when the action is selected so removing this line would fix my issue.
I can't find a way to override this programmatically from our wrapper.

I am happy to contribute to change this if needed.

Thank you for your time!

PR #584, opened 2022-10-12
Prevent component controls collapsing when option is selected

arussell123:
Stops the component controls from automatically collapsing whenever an option is selected, see issue #582.

### DIFF
--- a/src/mol-plugin-ui/structure/components.tsx
+++ b/src/mol-plugin-ui/structure/components.tsx
@@ -280,7 +280,6 @@ class StructureComponentGroup extends PurePluginUIComponent<{ group: StructureCo
 
     selectAction: ActionMenu.OnSelect = item => {
         if (!item) return;
-        this.setState({ action: void 0 });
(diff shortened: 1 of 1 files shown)


dsehnal:
I don't think we should make this change globally. This control is often used as a replacement for `<select />` control and it is expected that it immediate collapses.

I suggest adding `doNotCollapseControl` option to `ParamDefinition.(Multi)Select` and corresponding support to the control which renders the options as an alternative.

arussell123:
@dsehnal Sorry if I've misunderstood something here but wouldn't this change only effect the behaviour when you select an action in the `StructureComponentGroup`. Is this component used elsewhere (it is not exported so I assumed it was only used in this file). If I add the option to `Select` the wider group will collapse.

I can add a prop to control this behaviour if this behaviour isn't wanted by default in the `StructureComponentControls` component.

dsehnal:
@arussell123 you are right, my bad here.

Issue #597, opened 2022-10-24
QA metric coloring beyond pLDDT


gtauriello:
I was checking out the support of ModelCIF extensions like ma_qa_metric in molstar and was happy to see that there is some code aimed at also colouring based on QMEAN (additionally to pLDDT for AlphaFold models). Unfortunately, the code seems to rely on specific strings in `_ma_qa_metric.name` which is not very reliable since that is a free text field. I would suggest to use `_ma_qa_metric.type` ([link to doc](https://mmcif.wwpdb.org/dictionaries/mmcif_ma.dic/Items/_ma_qa_metric.type.html)) instead which is an enumeration and defines different types of lDDT predictors (e.g. "pLDDT" as used by AlphaFold or "pLDDT all-atom in [0,1]" as used by QMEAN/QMEANDisCo in SWISS-MODEL).

As an example, here is a ModelArchive model using QMEANDisCo for model confidence: https://modelarchive.org/doi/10.5452/ma-tbvar3d-19

Would be great to have such models displayed with colours according to their per-residue model confidence. The code should mostly be there already (e.g. if I rename `_ma_qa_metric.name` to "qmean" in the ma-tbvar3d-19 example, it seems to work), but one can generalize it to support any type of per-residue model confidence score which has a fixed range of values (as of now that would be the various "pLDDT" scores and "normalized score").

As a side note: as I am part of the ModelCIF working group and of the ModelArchive (MA) development team, I am happy to provide more input on ModelCIF and how we use it in MA if that is desired.

arose:
Hi @gtauriello, I agree that using the type enum is better. One challenge is how to make these scores available for use. For now I just map specific type to names that show up in the UI. This can obviously be improved, as needed, to handle more of he flexibility the `_ma_qa_metric` category allows for. Any suggestions and PRs are welcome.

gtauriello:
This seems to be well addressed by #1476 (I tested it on ModelArchive entries with accession codes "ma-denv-03" and "ma-tbvar3d-19" and also on a non-public entry with multiple pLDDT variants)

PR #602, opened 2022-10-29
use ma_qa_metric.type instead of .name

arose:
addressing #597

### DIFF
--- a/src/extensions/model-archive/quality-assessment/prop.ts
+++ b/src/extensions/model-archive/quality-assessment/prop.ts
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2021 mol* contributors, licensed under MIT, See LICENSE file for more info.
+ * Copyright (c) 2021-2022 mol* contributors, licensed under MIT, See LICENSE file for more info.
  *
  * @author Alexander Rose <alexander.rose@weirdbyte.de>
  */
@@ -14,14 +14,23 @@ import { CustomPropSymbol } from '../../../mol-script/language/symbol';
 import { Type } from '../../../mol-script/language/type';
 import { CustomPropertyDescriptor } from '../../../mol-model/custom-property';
 import { MmcifFormat } from '../../../mol-model-formats/structure/mmcif';
+import { mmCIF_Schema } from '../../../mol-io/reader/cif/schema/mmcif';
 
 export { QualityAssessment };
 
+type MetricType = mmCIF_Schema['ma_qa_metric']['type']['T']
+type MetricName = 'pLDDT' | 'qmean';
+
 interface QualityAssessment {
-    localMetrics: Map<string, Map<ResidueIndex, number>>
+    localMetrics: Map<MetricType, Map<ResidueIndex, number>>
     pLDDT?: Map<ResidueIndex, number>
     qmean?: Map<ResidueIndex, number>
 }
+
+const NameToType: { [k in MetricName]: MetricType } = {
+    pLDDT: 'pLDDT',
+    qmean: 'pLDDT all-atom in [0,1]',
+};
 
 namespace QualityAssessment {
     const Empty = {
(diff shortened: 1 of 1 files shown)


arose:
@gtauriello what do you think, does this fix it for you?

arose:
ping @gtauriello

gtauriello:
@arose Thanks for looking into this. And sorry for the delayed answer. I was travelling last week and could only look at it now.

I cannot fully judge if the code works as intended (I don't have a setup to test it), but from what I see in the proposed code changes, I would assume that it does the job for the currently existing ModelCIF files.

If it's ok, I comment here on how to make scores available for use (as commented on in #597).

I see 2 separate issues on that:

1. What colour scheme to use for display. In SWISS-MODEL we chose to offer both colour schemes to any type of model ("Confidence gradient" for our classic "QMEAN" scheme and "Confidence class" for the one used by AlphaFold DB; see [here for doc](https://swissmodel.expasy.org/docs/help#colour_schemes) and [here for an example with both SM and AFDB model](https://swissmodel.expasy.org/repository/uniprot/Q6DQX2)). The schemes are scaled to [0,1] or [0,100] depending on the type of score. If you want to use a heuristic to pick between "gradient" and "class" for the user (as done currently), that's of course also fine.
2. What label to set for the confidence score. Personally, I wouldn't call them "pLDDT Score" and "QMEAN Score" but rather label them as "X (Y)" with "X" being "Pred. lDDT" or "Pred. lDDT-CA" (depending on `_ma_qa_metric.type`) and "Y" being whatever is in `_ma_qa_metric.name`. This would be more robust for future scores. To clarify the used colour scheme one can additionally add "gradient" or "class" somewhere in the label.

Of course the suggestions above come at some amount of extra complexity for the code as one needs to keep track if we do "gradient" or "class" display, whether range is in [0,1] or [0,100] and what label to have for the score (as opposed to all of those being hard coded as it is now).

As a side note: it seems that ESMFold has been using AF-like pLDDT values but scaled to [0,1]. So if one ever sees ModelCIF files from them, those should have "pLDDT in [0,1]" as type.

dsehnal:
Is this something we still want to address?

dsehnal:
@gtauriello Thanks, can you please clarify this:

> Of course the suggestions above come at some amount of extra complexity for the code as one needs to keep track if we do "gradient" or "class" display, whether range is in [0,1] or [0,100] and what label to have for the score (as opposed to all of those being hard coded as it is now).

What criteria do we need to use to determine the gradient/class display and the ranges? 

Generally don't mind adding extra logic, just need to know what the criteria are.

gtauriello:
Sure. So here is the proposed logic:

- Pick range according to [`_ma_qa_metric.type`](https://mmcif.wwpdb.org/dictionaries/mmcif_ma.dic/Items/_ma_qa_metric.type.html). The values "pLDDT", "pLDDT all-atom" and "pLDDT to polymer" have a range in [0, 100] while "pLDDT all-atom in [0,1]" and "pLDDT in [0,1]" have it in [0, 1].
- Either allow user to switch between gradient and class display or pick according to `_ma_qa_metric.name`. This will never be fully robust since any text is valid as the name. In my experience, one can distinguish AlphaFold-variants which use "plddt" and QMEAN-variants which use "qmean" (both case-insensitive) somewhere in the name. So I would use that as logic to pick (the default) display: gradient for QMEAN and class for AlphaFold.
- Use label "[X] ([Y]; [Z])" for color theme name where [X] is `_ma_qa_metric.name`, [Y] is "gradient" or "class", and [Z] is `_ma_qa_metric.type` (one could also rename the types but not worth the effort I think)
- Use label "[X] ([Z])" (same [X] and [Z] as above) for the scores shown in mouseover in the 3D view

As a general note: it is possible to have multiple scores in `_ma_qa_metric` with `type` = "pLDDT" and `mode` = "local". So far I have only encountered one set of models in ModelArchive (or elsewhere) with 2 per-residue scores (not public yet but I can also generate a mock example if needed). Not sure which per-residue score would be picked by mol* but at least the label texts suggested above make it explicit as one can distinguish them by `_ma_qa_metric.name`. And being able to switch between them as color theme would of course be great.

dsehnal:
Replaced by  #1476

Issue #558, opened 2022-09-13
Transparent background blending issue. Pixels of wrong colors transparent objects


giagitom:
I have noticed this when enlarging the image. Could this be something related to antialiasing?

When exporting the image it looks like this
![image](https://user-images.githubusercontent.com/20627977/189885771-d38123ba-1359-44e3-96e8-691d97f7a74a.png)
![image](https://user-images.githubusercontent.com/20627977/189881410-259ced11-3996-4e25-9519-846f1e95e097.png)
![image](https://user-images.githubusercontent.com/20627977/189881667-17ff92d0-fec1-408f-82bc-a3b7cb8339fc.png)
![image](https://user-images.githubusercontent.com/20627977/189886335-2b3a1411-fe63-4287-bae7-e9476c19f052.png)
![image](https://user-images.githubusercontent.com/20627977/189887007-c365d9da-a65d-49a3-be91-7545a0519f7f.png)

giagitom:
Anyone else noticed this artifacts?

dsehnal:
At least the "inner ones" look like an overflow on near-white colors. Perhaps the edges too. @arose any idea where this could happen the shaders?

giagitom:
An other (maybe related) thing.. This is a screenshot taken from an external screen capture tool (top) and a normal screenshot with transparent background that I overlayed for comparison purposes with molstar empty canvas (bottom). I notice two things:
- a thin white halo on borders of both opaque and transparent objects (only on the top image)
![image](https://user-images.githubusercontent.com/20627977/191934798-9b61ddc9-760a-4b1f-a5b6-806b88a2fe95.png)

- and also a slight change in colors that is more visible on transparent objects (that can be seen better in the oversaturated image below)
![image](https://user-images.githubusercontent.com/20627977/191940337-17eba8c0-87e4-46b6-9eac-b98a770139ba.png)

dsehnal:
Maybe it's the ambient occlusion. If you turn that off, does this still happen? And if you turn of antialias?

giagitom:
How to disable antialias? 
Disabling ambient occlusion i get this (same configuration)

- The thin white outline is more marked than before 
- I start seeing strange colors on borders on top 
![image](https://user-images.githubusercontent.com/20627977/191945055-5dc607ee-240f-4cd4-8444-d16fda16cfa6.png)

- The change in colors is still present 
[...]

PR #603, opened 2022-10-30
Transparent bg fix

giagitom:
@arose here is the PR addressing #558. I had to clamp from 0.01 to 0.99 cause I just noticed that when exporting images larger than the viewport the issue was still visible


arose (commented):
thanks, looks good

PR #604, opened 2022-10-31
by-volume-value theme

dsehnal:
# Description

coloring of arbitrary (cpu generated) geometries by user-selected volume

image

## Actions


src/mol-theme/color/external-volume.ts
```diff
@@ -0,0 +1,159 @@
...
+
+export function ExternalVolumeColorTheme(ctx: ThemeDataContext, props: PD.Values<ExternalVolumeColorThemeParams>): ColorTheme<ExternalVolumeColorThemeParams> {
+    let volume: Volume | undefined;
+    try {
+        volume = props.volume.getValue();
+    } catch {
+        // .getValue() is resolved during state reconciliation => would throw from UI
+    }
+
+    // NOTE: this will currently not work with GPU iso-surfaces since it requires vertex coloring
```
arose, line 60: added a texture-mesh position iterator, so it will work by downloading the positions form the gpu

arose (commented):
Loading a state does not work for me. Otherwise, great!!!

dsehnal:
> Loading a state does not work for me.

@arose this will currently only work if the volume is loaded before the structure. But I have some ideas about updating `mol-state` to handle this case.

PR #613, opened 2022-11-09
Clamp cube

arose:
# Description

## Actions

### DIFF
--- a/src/mol-plugin-state/transforms/volume.ts
+++ b/src/mol-plugin-state/transforms/volume.ts
@@ -88,6 +88,8 @@ const VolumeFromCube = PluginStateTransform.BuiltIn({
         return {
             dataIndex,
             entryId: PD.Text(''),
+            clampMin: PD.Numeric(-1024),
+            clampMax: PD.Numeric(1024),
         };
     }
 })({
--- a/src/mol-plugin-state/transforms/volume.ts
+++ b/src/mol-plugin-state/transforms/volume.ts
@@ -88,14 +88,24 @@ const VolumeFromCube = PluginStateTransform.BuiltIn({
         return {
             dataIndex,
             entryId: PD.Text(''),
-            clampMin: PD.Numeric(-1024),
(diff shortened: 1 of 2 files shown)


arose:
Not too sure about this approach as I think it skews the relative isovalue

src/mol-plugin-state/transforms/volume.ts
```diff
@@ -88,12 +88,24 @@ const VolumeFromCube = PluginStateTransform.BuiltIn({
         return {
             dataIndex,
             entryId: PD.Text(''),
+            clamp: PD.MappedStatic('off', {
+                'off': PD.EmptyGroup(),
+                'on': PD.Group({
+                    min: PD.Numeric(-1024),
+                    max: PD.Numeric(1024),
```
dsehnal, line 95: Add `preserveRelativeValue: PD.Bool()` ?

src/mol-plugin-state/transforms/volume.ts
```diff
@@ -88,12 +88,24 @@ const VolumeFromCube = PluginStateTransform.BuiltIn({
         return {
             dataIndex,
             entryId: PD.Text(''),
+            clamp: PD.MappedStatic('off', {
```
luna215, line 91: Wouldn't this turn off the clamp feature automatically every time we try to load a cube?

luna215:
Thanks for this @arose! Looks like when I turn on the clamp feature, the cube file renders nicely:

https://user-images.githubusercontent.com/7971405/201384738-048c9655-49df-447a-b8b3-23cfbbed7a5d.mov

I left a comment about the clamp feature defaulting to off whenever we load a cube and not being able to toggle on/off programmatically. Unless there is a way to do this and I'm just not aware?

dsehnal:
@arose do we still need this?

andy197:
Just bumping this! We generated cube files that we can visualize on Pymol, but not in Mol*. We believe it's related to this clamping issue.

PR #615, opened 2022-11-10
keep asset url to detect compression

JonStargaryen:
# Description
Tiny quality-of-life improvement that extends the handling of zip/gzip assets to URLs.
https://files.rcsb.org/download/4HHB.cif.gz
https://files.rcsb.org/download/4HHB.pdb.gz
https://models.rcsb.org/4HHB.bcif.gz

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Note that since we don't clearly distinguish between a public and private interf
 ## [Unreleased]
 
 - Make `PluginContext.initContainer` checkered canvas background optional
+- Store URL of downloaded assets to detect zip/gzip based on extension
 
 ## [v3.23.0] - 2022-10-19
 
--- a/src/mol-util/assets.ts
+++ b/src/mol-util/assets.ts
@@ -100,7 +100,7 @@ class AssetManager {
                 }
 
                 const data = await ajaxGet({ ...asset, type: 'binary' }).runInContext(ctx);
-                const file = new File([data], 'raw-data');
+                const file = new File([data], asset.url);
                 this._assets.set(asset.id, { asset, file, refCount: 1 });
                 return Asset.Wrapper(await readFromFile(file, type).runInContext(ctx), asset, this);
             });


dsehnal (commented):
It looks fine, hopefully won't bite us... what if the URL will have the .gz there, but will have content-encoding set as well?

dsehnal:
Maybe we can check the response headers and only add the "filename" if it doesn't have the gzip content encoding?

JonStargaryen:
> what if the URL will have the .gz there, but will have content-encoding set as well?

That's actually the case for https://models.rcsb.org/4HHB.bcif.gz (should fix that though...) and it works.

> Maybe we can check the response headers and only add the "filename" if it doesn't have the gzip content encoding?

Hmm, I tried to check headers inside of ajaxGetInternal() but content-encoding isn't set at that point. Chrome at least seems to handle/hide all of that (content-encoding = gzip appears in the network tab).

Another, more resilient, solution would be to check the first few bytes of the file for zip/gzip magic numbers and decompress as needed. That would work independent of the URL (or file extension) but has drawbacks as well.

That said, this is also fine to keep as is. Just saw this during testing and thought it would be nice/convenient to have.

dsehnal:
> That's actually the case for https://models.rcsb.org/4HHB.bcif.gz (should fix that though...) and it works.

Are you double gzipping your responses?

JonStargaryen:
Apparently for that endpoint :(

dsehnal:
Ok, this should not break anything... hopefully :)

JonStargaryen:
@dsehnal This indeed breaks anything related to validation reports (https://files.rcsb.org/pub/pdb/validation_reports/hh/4hhb/4hhb_validation.xml.gz). The browser handles the encoding and now Mol* trips up when trying to "decompress" the XML. 

Should we just revert this PR?

dsehnal:
Yea. And need new release then. Can you make a pr which undoes this?

JonStargaryen:
Sure thing, will do later today.

PR #619, opened 2022-11-13
add key to symmetry operator

arose:
- molql operatorKey
- index-pair operator prop

# Description

## Actions


src/mol-model-formats/structure/property/bonds/index-pair.ts
```diff
@@ -14,6 +14,8 @@ import { ElementIndex } from '../../../../mol-model/structure';
 
 export type IndexPairsProps = {
     readonly key: ArrayLike<number>
+    /** Sorted contor-paired operator keys */
```
dsehnal, line 17:
```suggestion
    /** Sorted cantor-paired operator keys */
```

dsehnal (commented):
I think it would be better to add two string columns with `operator_a` and `operator_b` instead since I think the name is better defined whereas the key is only implicit. What do you think?

PR #637, opened 2022-11-27
Screen-space shadows

arose:
Simple screen-space shadows. Changes default light to be in top-left. Shadows are generally off by default but on for cellpack models. More useful for large scale shadows but with some scene-dependent tweaking OK for small-scale stuff too.


arose:
image

arose:
image

giagitom:
This is awesome!

src/mol-gl/shader/postprocessing.frag.ts
```diff
@@ -120,7 +121,20 @@ void main(void) {
...
+        if (!isBackground(opaqueDepth)) {
+            viewDist = abs(getViewZ(opaqueDepth));
+            fogFactor = smoothstep(uFogNear, uFogFar, viewDist);
+            vec4 shadow = texture2D(tShadows, coords);
+            if (!uTransparentBackground) {
+                color.rgb = mix(mix(vec3(0), uFogColor, fogFactor), color.rgb, shadow.a);
+            } else {
+                color.rgb = mix(vec3(0) * (1.0 - fogFactor), color.rgb, shadow.a);
+            }
+        }
```
dsehnal, line 134: Do you think we could get significant extra performance by optimizing the shaders? E.g. reducing the number of operations that happen across the board.

src/mol-gl/shader/shadows.frag.ts
```diff
@@ -0,0 +1,130 @@
...
+
+uniform mat4 uProjection;
+uniform mat4 uInvProjection;
+
+uniform float uMaxDistance;
+uniform float uTolerance;
+uniform float uBias;
+
+bool isBackground(const in float depth) {
+    return depth == 1.0;
```
dsehnal, line 35: Won't this  break with different precisions/drivers/etc?

dsehnal (approved):
Very cool!

src/mol-gl/shader/shadows.frag.ts
arose: We use it in a bunch of places and haven't seen issues. The 1.0 comes from clearing the texture to 1.0.

src/mol-gl/shader/postprocessing.frag.ts
arose: The whole shadow pp takes <0.2 ms, compared to 2ms for occlusion, which is much slower due to more and incoherent texture fetches. I don't think we are at the point to do sub-millisecond optimizations :)

PR #643, opened 2022-12-02
Add include transparent parameter for outlines

arussell123:
# Description
Add a parameter to the outline settings to choose whether to show the outlines for transparent components.

## Actions


arose:
Thank you for doing this! I was thinking more of a flag to toggle outlines for anything with opacity < 1. We may want to use the transparent depth buffer for other things than outlines.

I would add a `includeTransparent` param to `PostprocessingParams.outline` and then add a define to `getDepthTransparent` in `outlines.frag` and `postprocessing.frag` to return 1.0 when switched off

```
float getDepthTransparent(const in vec2 coords) {
    #ifdef dTransparentOutline
        return unpackRGBAToDepth(texture2D(tDepthTransparent, coords));
    #else
        return 1.0;
    #endif
}
```

You should also rename `PostprocessingPass.isOutlineEnabled` to `PostprocessingPass.isTransparentOutlineEnabled` and check if transparent outlines are on to avoid doing a depth pass.

arose:
Thank you for the changes @arussell123. Could you please also replace `PostprocessingPass.isOutlineEnabled` with

```
static isTransparentOutlineEnabled(props: PostprocessingProps) {
    return props.outline.name === 'on' && props.outline.params.includeTransparent;
}
```

this should avoid doing the extra depth pass for transparent outlines in `mol-canvas3d/passes/draw.ts`

arussell123:
Ah good point, I'll change it now!

arose:
splendid, thank you @arussell123

PR #644, opened 2022-12-03
support power-preference webgl attribute

arose:
Support for ``powerPreference`` webgl attribute. Add ``PluginConfig.General.PowerPreference`` and ``power-preference`` Viewer GET param.

At the moment the attribute is only honored on dual gpu macs but other oses will follow.


dsehnal (approved):
Nice.

PR #657, opened 2022-12-12
Fix QualityAssessment assignment

### LINKED ISSUE #656: Wrong pLDDT display if label_seq_id and auth_seq_id do not match
I have observed that the per-residue pLDDT values do not get displayed correctly if label_seq_id and auth_seq_id do not match in the file. The pLDDT values stored in the `_ma_qa_metric_local` table are mapped to residues using `label_seq_id` (see [here](https://mmcif.wwpdb.org/dictionaries/mmcif_ma.dic/Items/_ma_qa_metric_local.label_seq_id.html) for doc) and so it should be irrelevant whether or not there is a different `_atom_site.auth_seq_id` in the file. Yet this does not seem to be the case.

In the attached package ([molstar-issue-testfiles.zip](https://github.com/molstar/molstar/files/10210331/molstar-issue-testfiles.zip)), there are two example files which only differ by how `auth_seq_id` values were filled:

- `O00257-D2_Q6W2J9-D8_auth.cif` is the file I would like to be handled properly. The pLDDT values in the file correctly relate to the residue numbers given in `label_seq_id` but yet I only see correct pLDDT values for the few residues in chain A which happen to have identical numbers for `label_seq_id` and `auth_seq_id`
- `O00257-D2_Q6W2J9-D8_no_auth.cif` "fixes" the display problem by simply having all `auth_seq_id` numbers matching the `label_seq_id`

As I would like to use mol* to display `O00257-D2_Q6W2J9-D8_auth.cif` in ModelArchive (file will be deposited there in the next few days), I would be happy if this can be fixed soon. It sounds like a minor bug to me, but as I don't know the mol* code, I hope it's not a can of worms...

dsehnal:
# Description

Fixes #656

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Note that since we don't clearly distinguish between a public and private interf
     - Fix sphere near-clipping with orthographic projection
     - Fix cylinder near-clipping
     - Add interior cylinder caps
+- Fix `QualityAssessment` assignment bug for structures with different auth vs label sequence numbering
 
 ## [v3.26.0] - 2022-12-04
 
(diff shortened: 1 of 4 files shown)


arose (approved):
thanks, lgtm

Issue #635, opened 2022-11-25
Question about representation visuals


giagitom:
I'm trying to modify a representation visual to learn how it works. I have a few questions:
- Can I make a visual where some specific parts of its mesh can have a variable opacity? Or should I make two separate visuals to achieve this?
- Can i make a visual composed of mesh and impostors at the same time?
- How to construct a visual in order to be able to distinguish specific parts of it in the color theme?

arose:
> Can I make a visual where some specific parts of its mesh can have a variable opacity? Or should I make two separate visuals to achieve this?

You can use `Representation.setState({ transparency })` for per-group transparency.

> Can i make a visual composed of mesh and impostors at the same time?

No, you should create a representation with two visuals.

> How to construct a visual in order to be able to distinguish specific parts of it in the color theme?

When creating geometry (mesh, spheres, lines, ...) you assign a `group` to a vertex, sphere, line, ... When creating a visual you provide functions that map those groups to locations and loci.

Have a look at https://github.com/molstar/molstar/blob/0b2459f4e1f87d95b189aca851883f3bc2024e66/src/mol-repr/structure/visual/carbohydrate-symbol-mesh.ts for an example that does not use structure element index as the `group`.

giagitom:
Thanks @arose.

> When creating geometry (mesh, spheres, lines, ...) you assign a group to a vertex, sphere, line, ... When creating a visual you provide functions that map those groups to locations and **loci.**

so basically `builderState.currentGroup` is used to set the correct group for the next mesh to be added?

arose:
> so basically builderState.currentGroup is used to set the correct group for the next mesh to be added?

for meshes, yes

giagitom:
I'm searching for a way to alter the opacity of a specific visual without altering opacity of all other visuals in the representation. Is it possible in some way?

arose:
Not really, only via setTransparency and only if the visuals map to different loci. Probably best to create multiple representations.

giagitom:
Ok thanks, I'll work with two representations and show the second only when necessary. By the way, which is the correct way to hide/show a representation? Should I add/remove it from the canvas using ```canvas3d.remove(repr)``` or ```canvas3d.add(repr)```? Or there is a better way?

arose:
Use `repr.setState({ visible: false });`. Just improved the typing a bit in f84b5b633d36f32a0bc4c527dc743506a02e92b5.

giagitom:
A few other questions about this
[...]

PR #659, opened 2022-12-13
Impostor cylinders interior

giagitom:
# Description
Improving impostor cylinder interior when clipping to address #635
It is working well with my custom visual. Using default ball and sticks representation seems that there is no way to enable both caps to test..

image

### DIFF
--- a/src/mol-gl/shader/cylinders.frag.ts
+++ b/src/mol-gl/shader/cylinders.frag.ts
@@ -2,6 +2,7 @@
  * Copyright (c) 2020-2022 mol* contributors, licensed under MIT, See LICENSE file for more info.
  *
  * @author Alexander Rose <alexander.rose@weirdbyte.de>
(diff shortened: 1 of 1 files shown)


arose:
cool, thanks, maybe we can have it for default ball & sticks as well, trying a few things in https://github.com/molstar/molstar/tree/solid-impostor

giagitom:
It seems a very nice improvement! Beautiful.

 - I'm seeing some artifacts here (maybe some inverted normals?)
image
image

 - And some z fighting between the two central caps
image


 - Regarding transparency instead do you think there is a way to remove the center cap?
image


 - I was thinking about a way to handle cylinder sphere intersections at fragment level like the one below . 
 What do you think about discarding fragments using the same method that uses ```clip_pixel``` chunk using ```sphereSD```? 
 It would be necessary to know the sphere radius though.

image

arose:
closing in favor of #665

PR #665, opened 2022-12-15
Solid impostors

arose:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Note that since we don't clearly distinguish between a public and private interf
     - Fix cylinder near-clipping
     - Add interior cylinder caps
     - Add per-pixel object clipping
+    - Add `solidInterior` parameter
 - Fix `QualityAssessment` assignment bug for structures with different auth vs label sequence numbering
 - Refresh `ApplyActionControl`'s param definition when toggling expanded state
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ Note that since we don't clearly distinguish between a public and private interf
 ## [Unreleased]
 
 - Show histogram in direct volume control point settings
-
+- Add `solidInterior` parameter to sphere/cylinder impostors
 
 ## [v3.27.0] - 2022-12-15
 
@@ -19,7 +19,6 @@ Note that since we don't clearly distinguish between a public and private interf
     - Fix cylinder near-clipping
     - Add interior cylinder caps
     - Add per-pixel object clipping
-    - Add `solidInterior` parameter
 - Fix `QualityAssessment` assignment bug for structures with different auth vs label sequence numbering
 - Refresh `ApplyActionControl`'s param definition when toggling expanded state
 - Fix `struct_conn` bond assignment for ions
--- a/src/mol-geo/geometry/cylinders/cylinders.ts
+++ b/src/mol-geo/geometry/cylinders/cylinders.ts
@@ -158,6 +158,7 @@ export namespace Cylinders {
         ignoreLight: PD.Boolean(false, BaseGeometry.ShadingCategory),
         xrayShaded: PD.Boolean(false, BaseGeometry.ShadingCategory),
         transparentBackfaces: PD.Select('off', PD.arrayToOptions(['off', 'on', 'opaque']), BaseGeometry.ShadingCategory),
+        solidInterior: PD.Boolean(true, BaseGeometry.ShadingCategory),
         bumpFrequency: PD.Numeric(0, { min: 0, max: 10, step: 0.1 }, BaseGeometry.ShadingCategory),
         bumpAmplitude: PD.Numeric(1, { min: 0, max: 5, step: 0.1 }, BaseGeometry.ShadingCategory),
     };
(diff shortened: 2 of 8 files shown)


arose:
It seems a very nice improvement! Beautiful.

 - I'm seeing some artifacts here (maybe some inverted normals?)
image
image

 - And some z fighting between the two central caps
image


 - Regarding transparency instead do you think there is a way to remove the center cap?
image


 - I was thinking about a way to handle cylinder sphere intersections at fragment level like the one below . 
 What do you think about discarding fragments using the same method that uses ```clip_pixel``` chunk using ```sphereSD```? 
 It would be necessary to know the sphere radius though.

image

_Originally posted by @giagitom in https://github.com/molstar/molstar/issues/659#issuecomment-1351811135_

arose:
@giagitom, the issues with normals and z-fighting should be fixed with 86f7a8b273cdd72eda54164c2d8f5ce1e4ce1748.

> Regarding transparency instead do you think there is a way to remove the center cap?

It is not really a cap but two layers of transparency from two cylinders. I think you need to use a single cylinder to gt rid of it.

> I was thinking about a way to handle cylinder sphere intersections at fragment level like the one below.
> What do you think about discarding fragments using the same method that uses clip_pixel chunk using sphereSD?
> It would be necessary to know the sphere radius though.

Nice idea but it would only help with the parts where the cylinder penetrates the sphere, you will still have artifacts where the sphere penetrates the cylinder. I think if you want a really clean transparent look you will need to either ray trace the whole molecule as one object or create a solid mesh of the whole molecule using solid constructive geometry (CSG).

giagitom:
Awesome! I confirm it is working well now. The only issue I noticed is something broken when clipping spheres that suddenly disappear.

issue



> Nice idea but it would only help with the parts where the cylinder penetrates the sphere, you will still have artifacts where the sphere penetrates the cylinder

Hmm, I'm not sure I'm getting why there would be further intersections.. I noticed we also have cylinderSD by the way, maybe it could be used on sphere impostor to handle sphere cylinder intersections. Also isn't this already something similar to CSG?

arose:
> Awesome! I confirm it is working well now. The only issue I noticed is something broken when clipping spheres that suddenly disappear.

right, bad refactor, fixed in 7aae2d0616e670910bf1c57776d042b4299a1d57

arose:
> Hmm, I'm not sure I'm getting why there would be further intersections..

By artifact I mean more than one layer of transparency for bonded atoms, that is spheres connected with cylinders. Each sphere and cylinder are rendered separately and so whenever they intersect there will be multiple layers of transparency.

> I noticed we also have cylinderSD by the way, maybe it could be used on sphere impostor to handle sphere cylinder intersections.

Sure we can add a list of clip objects to each sphere and cylinder... basically for each neighboring sphere and cylinder. There are also cylinder-cylinder intersections that are not within a sphere if you look closely at your examples above. Technically I think it is possible but do we want to go there though?

> Also isn't this already something similar to CSG?

Yes  it is.

arose:
@giagitom also note that there is `adjustCylinderLength` param which can help with transparent ball & sticks.

image

giagitom:
I have seen something else maybe. In a situation like this
image

If I try to enable linkCap I obtain this
image

But I can't see any change enabling stubCap. I exected the interior to disappear in the center but it remains.

giagitom:
> Sure we can add a list of clip objects to each sphere and cylinder... basically for each neighboring sphere and cylinder. There are also cylinder-cylinder intersections that are not within a sphere if you look closely at your examples above. Technically I think it is possible but do we want to go there though?

@arose yes, you are right. I think that it is not necessary to to implement this at the moment. 



> also note that there is adjustCylinderLength param which can help with transparent ball & sticks.

yeah I was already aware of that, but there are a few problems when the aspect ratio is near or equal to 1, so I was searching for a way to better handle that particular situations.

image

giagitom:
Do you think it could be possible to extend the ```solidInterior``` functionality to meshes also? 
image

arose:
> But I can't see any change enabling stubCap. I exected the interior to disappear in the center but it remains.

The cap, or more precisely the back of cylinder B, in the center remains because it is the a second transparent layer, the first layer is the front of cylinder A.

> yeah I was already aware of that, but there are a few problems when the aspect ratio is near or equal to 1, so I was searching for a way to better handle that particular situations.

Maybe it is best to use high quality sphere/cylinder meshes if you want transparency together with double-sided geometry.

arose:
> Do you think it could be possible to extend the solidInterior functionality to meshes also? To achieve something like this for example

With some caveats, yes, I outlined an option in #667 if you want to give it a try. I think it is more work than the depth peeling support.

giagitom:
Thanks @arose 

> The cap, or more precisely the back of cylinder B, in the center remains because it is the a second transparent layer, the first layer is the front of cylinder A.

Sorry, I misspoke. I was just thinking to make only the interior disappear by enabling the cap of cylinder B. This is just because the interior is darker than the cap. But maybe there are drawbacks in doing this that I don't know. By the way what does the stubCap option should do normally?

arose:
> Sorry, I misspoke. I was just thinking to make only the interior disappear by enabling the cap of cylinder B. This is just because the interior is darker than the cap. But maybe there are drawbacks in doing this that I don't know.

It would be less dark (if you interior handling makes things darker) but you would still get a second transparency layer. You can emulate how it would look like by setting render interiorDarkening to 0.

> By the way what does the stubCap option should do normally?

The stubCap option has only an effect on half-bond, that is, bonds for which only one cylinder is shown.

PR #663, opened 2022-12-14
Ability to show only polar hydrogens

dsehnal:
# Description

## Actions

- [ ] Added description of changes to the `[Unreleased]` section of `CHANGELOG.md`
- [ ] Updated headers of modified files
- [ ] Added my name to `package.json`'s `contributors`

PR #669, opened 2022-12-19
tweak ignoreHydrogens non-polar handling

arose:
Follow up to #663.


arose:
@jpattle proposed tweak. In 4.0 we may make `inoreHydrogens` itself a union type, e.g., `'none' | 'all' | 'non-polar'` and potentially other options.

dsehnal (approved):
LGTM

PR #672, opened 2022-12-19
Volseg: Fixed not loading volume section from state

midlik:
# Description

## Actions

### DIFF
--- a/src/extensions/volumes-and-segmentations/entry-root.ts
+++ b/src/extensions/volumes-and-segmentations/entry-root.ts
@@ -151,6 +151,9 @@ export class VolsegEntryData extends PluginBehavior.WithSubscribers<VolsegEntryP
             // do nothing
         }
 
+        const volumeVisual = this.findNodesByTags(VOLUME_VISUAL_TAG)[0];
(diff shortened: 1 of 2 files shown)


src/extensions/volumes-and-segmentations/entry-root.ts
```diff
@@ -151,6 +151,9 @@ export class VolsegEntryData extends PluginBehavior.WithSubscribers<VolsegEntryP
             // do nothing
         }
 
+        const volumeVisual = this.findNodesByTags(VOLUME_VISUAL_TAG)[0];
+        if (volumeVisual) this.currentVolume.next(volumeVisual.transform);
```
dsehnal, line 155: This isn't correct if there is more than 1 segmentations loaded.

Need to use the `StateSelection.findAncestor(this.plugin.state.data.tree, this.plugin.state.data.cells, e.ref, a => a.transform.ref === ref)` as well here.

midlik: this.findNodesByTag only searches the subtree of current entry root.
So it should be OK.

PR #678, opened 2022-12-23
IsolatedModules tsconfig

dsehnal:
# Description

Add `isolatedModules: true` to tsconfig.json. Should help with integrating Mol* to 3rd party libraries.

## Actions


dsehnal:
Ok, this is actually more complicated and will require breaking changes so will have to wait for 4.0.

PR #680, opened 2022-12-26
Fix some cyclic imports & reduce const enum usage

dsehnal:
# Description

Fix some cyclic imports and reduce the use of const enums. This should make it easier to use the library with the `isolatedModules: true` TS config.

## Actions


src/mol-model/structure/structure/unit.ts
```diff
@@ -36,6 +36,9 @@ type Unit = Unit.Atomic | Unit.Spheres | Unit.Gaussians
 namespace Unit {
     export const enum Kind { Atomic, Spheres, Gaussians }
 
+    // To use with isolatedModules
```
arose, line 39: I like this pattern

src/mol-model-props/computed/interactions/common.ts
```diff
@@ -158,7 +158,7 @@ export function interactionTypeLabel(type: InteractionType): string {
     }
 }
 
-export const enum FeatureType {
+export enum FeatureType {
```
arose, line 161: These are accessed fairly often, maybe worthwhile to use the `const enum FeatureType` and `enum FeatureTypes` pattern?

src/mol-model-props/computed/accessible-surface-area/shrake-rupley.ts
```diff
@@ -89,12 +89,19 @@ namespace AccessibleSurfaceArea {
...
     }
 
-    export const enum Flag {
+    export const enum Flags {
         NA = 0x0,
         Buried = 0x1,
         Accessible = 0x2
     }
 
+    export const Flag = {
```
arose, line 98: For this pattern, shall we always define the `enum` in term of the `const enum`? Also we should be consistent with the use of the plural for either the `enum`or the `const enum`.

src/mol-data/db/column.ts
```diff
@@ -93,20 +93,35 @@ namespace Column {
...
         Present = 0,
-        /** Expressed in CIF as `.` */
+        /** Expressed in CIF as `.` (= 1) */
         NotPresent = 1,
-        /** Expressed in CIF as `?` */
+        /** Expressed in CIF as `?` (= 2) */
         Unknown = 2
     }
 
+    export const ValueKind = {
```
arose, line 108: Define in terms of the `const enum`?

arose (commented):
Nice, thanks for tackling this. Shall we add a lint rule to allow `const enums` only with a comment?

src/mol-theme/color/_categories.ts
```diff
@@ -0,0 +1,14 @@
+/**
```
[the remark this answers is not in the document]
dsehnal: Wanted it to be on top, but yeah, can rename it.

dsehnal:
> Shall we add a lint rule to allow const enums only with a comment?

Perhaps. Know how to do this?

PR #693, opened 2023-01-14
support dim unmarked groups & marker edge strength

arose:
image

### DIFF
--- a/src/mol-canvas3d/passes/marking.ts
+++ b/src/mol-canvas3d/passes/marking.ts
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2021 mol* contributors, licensed under MIT, See LICENSE file for more info.
+ * Copyright (c) 2021-2023 mol* contributors, licensed under MIT, See LICENSE file for more info.
  *
  * @author Alexander Rose <alexander.rose@weirdbyte.de>
  */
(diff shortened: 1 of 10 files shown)


src/mol-canvas3d/passes/marking.ts
```diff
@@ -113,8 +115,10 @@ export class MarkingPass {
         const { values: overlayValues } = this.overlay;
         ValueCell.update(overlayValues.uHighlightEdgeColor, Color.toVec3Normalized(overlayValues.uHighlightEdgeColor.ref.value, highlightEdgeColor));
         ValueCell.update(overlayValues.uSelectEdgeColor, Color.toVec3Normalized(overlayValues.uSelectEdgeColor.ref.value, selectEdgeColor));
```
dsehnal, line 117: Should these 2 get `updateIfChanged` as well? Perhaps add one more version of `updateIfChanged` which accepts a comparator function, e.g., ``ValueCell.updateIfChangedCmp(a, b, Vec3.equals)``?

dsehnal (approved):
Nice.

src/mol-canvas3d/passes/marking.ts
dsehnal: Could be used in other places with `ValueCell.updateIfChangedCmp(a, b, shallowEquals)` too.

arose: Sounds good. In the context of uniforms it is not that important as they get uploaded tot the GPU every time anyway.

PR #694, opened 2023-01-15
wip, lint const enums

arose:
Follow up to #680.

@dsehnal could you check how to handle the remaining lint errors?


dsehnal:
@arose I don't really know how to fix this without doing `// eslint-disable` or introducing breaking changes (although it would likely not affect anyone).

Make this part of v4 release instead?

arose:
Yeah, would be v4 (or beyond). I had a look yesterday but did not feel quite motivated (never used isolated modules) to make decisions. I will leave it up to you what and when to handle this.

PR #699, opened 2023-01-17
Fix NtC tube and Confal pyramids mesh when there are discontinuous chains

MadCatX:
I just realized that when there are discontinuities in chains the mesh building logic won't detect them and draw things wrong. See the example below:

Wrong:
bad

Fixed:
fixed

This can probably only happen if the user applies a selection on the structure. I used the following input to create the testing example:

- Structure: https://dnatco.datmos.org/v4.1/RCSB/cif_dnatco_updated/1bna_v41C35A23.cif.gz
- Selection script:
```
(sel.atom.res
    (or
        (in-range atom.resno 1 2)
        (or
            (in-range atom.resno 5 9)
            (or
                (in-range atom.resno 11 12)
                (or
                    (in-range atom.resno 13 14)
                    (in-range atom.resno 16 21)
                )
            )
        )
    )
)
```

I'm sorry for not catching this sooner.

### DIFF
--- a/src/extensions/dnatco/behavior.ts
+++ b/src/extensions/dnatco/behavior.ts
@@ -39,11 +39,15 @@ export const DnatcoNtCs = PluginBehavior.create<{ autoAttach: boolean, showToolT
         }
 
         unregister() {
+            this.ctx.customModelProperties.unregister(ConfalPyramidsProvider.descriptor.name);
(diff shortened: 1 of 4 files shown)


arose:
Looks good, can you please add an entry to the Changelog since this fixes a released version, thanks

arose:
> Looks good, can you please add an entry to the Changelog since this fixes a released version, thanks

All good actually, wasn't released yet.

PR #700, opened 2023-01-21
Added HeadlessPluginContext, Canvas3DRenderer, image-renderer example

midlik:
# Description

I created `HeadlessPluginContext`, which allows running Mol* Plugin outside a browser (in Node.js).
An example of its usage is shown in `src/examples/image-renderer/`.

## Actions


russellp17:
this is cool! I'd expect these sorts of dependencies to be under [optionalDependencies](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#optionaldependencies) in package.json instead of unlisted, so that library users can decide whether to skip them if necessary - is there a reason not to do that?

dsehnal:
That's a good point @russellp17, listing these in optional dependencies makes sense. @midlik  can you please update this?

midlik:
Thank you for your suggestion, @russellp17, I added them to optionalDependencies.

dsehnal (approved):

src/mol-util/nodejs-browser-io.ts
```diff
@@ -0,0 +1,60 @@
+/**
+ * Copyright (c) 2018-2020 mol* contributors, licensed under MIT, See LICENSE file for more info.
```
arose, line 2: date...

src/mol-util/nodejs-browser-io.ts
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * @author Adam Midlik <midlik@gmail.com>
+ *
+ * Implements some browser-only global variables for Node.js environment.
+ * These workarounds will also work in browsers as usual.
+ */
+
+
+/** Determines whether the current code is running in Node.js or in a browser */
+export const RUNNING_IN_NODEJS = typeof document === 'undefined';
```
arose, line 12: there other runtime times... this checks if it is running in a browser, to check for node do

```
const isNode =
  typeof process !== "undefined" &&
  process.versions != null &&
  process.versions.node != null;
```

src/mol-util/lazy-imports.ts
```diff
@@ -0,0 +1,47 @@
+/**
+ * Copyright (c) 2019-2022 mol* contributors, licensed under MIT, See LICENSE file for more info.
```
arose, line 2: date...

src/mol-util/data-source.ts
```diff
@@ -7,21 +7,14 @@
  * Adapted from LiteMol
  */
 
+import * as fs from 'fs';
```
arose, line 10: Importing a node module like this causes issues in the browser and forces you to handle it with a bundler like webpack. Please find another way.

src/mol-canvas3d/renderer.ts
```diff
@@ -0,0 +1,223 @@
+/**
+ * Copyright (c) 2019-2022 mol* contributors, licensed under MIT, See LICENSE file for more info.
```
arose, line 2: date

src/mol-canvas3d/renderer.ts
```diff
@@ -0,0 +1,223 @@
+/**
```
arose, line 1: I would move this to mol-plugin/util/headless-screenshot.ts

src/mol-canvas3d/renderer.ts
```diff
@@ -0,0 +1,223 @@
...
+
+
+const lazyImports = LazyImports.create('gl', 'jpeg-js', 'pngjs') as {
+    'gl': typeof import('gl'),
+    'jpeg-js': typeof import('jpeg-js'),
+    'pngjs': typeof import('pngjs'),
+};
+
+
+export type ImageRendererOptions = {
```
arose, line 37: should have same name as the class it is used with

src/mol-canvas3d/renderer.ts
```diff
@@ -0,0 +1,223 @@
...
+
+export type RawImageData = {
+    data: Uint8ClampedArray,
+    width: number,
+    height: number,
+}
+
+
+/** To render Canvas3D when running in Node.js (without DOM) */
+export class Canvas3DRenderer {
```
arose, line 51: I would call this HeadlessScreenshotHelper

src/mol-util/nodejs-browser-io.ts
```diff
@@ -0,0 +1,60 @@
+/**
```
arose, line 1: I would call this file `nodesjs-shims.ts`, more generic

arose (commented):
A bit late, but here is my review. Nice addition, thanks @midlik.

midlik:
Thank you @arose , I will address your feedback in another PR.

src/mol-util/data-source.ts
midlik: OK, now importing 'fs' and 'path' via LazyImports

PR #727, opened 2023-02-18
Cartoon nucleic with sugar visual

giagitom:
# Description

Add new cartoon visuals to support nucleotide base with sugar. Addressing #635 

## Actions


arose:
thanks, I will look at it in a bit

arose:
Hi @giagitom, finally reviewed it. I refactored it a bit to make sure nucleotides are treated the same in all visuals by using shared code. Also added a `thicknessFactor` that also works for the block and ring visuals. I renamed the visuals to atomic-* to distinguish them from the existing ring visual. It was easier to make the changes rightaway than to discribe what I wanted, hope that is okay for you. Please have a look if everything still works as you want, then it is good go. @dsehnal want to have a look as well?

arose (approved):

PR #740, opened 2023-03-08
Change nodejs-shim conditional to avoid checking `document` object existence

russellp17:
# Description

On updating molstar I noticed some runtime errors were being thrown because we have `document` polyfilled in a NodeJS env. This change updates the conditionals in nodejs-shims.ts to avoid checking `typeof document === 'undefined'`.

## Actions


dsehnal:
Thanks

PR #742, opened 2023-03-10
Selection and focus improvements

jpattle:
# Description
* Updated the selection toggle buttons in measurements and selection controls so that they are not shown if selection mode is turned off in the plugin configuration
* Extended the camera focus bindings to allow override of the 'click-away' behaviour that resets the camera (enabled by default)

## Actions


src/mol-plugin/behavior/dynamic/camera.ts
```diff
@@ -28,6 +29,15 @@ const DefaultFocusLociBindings = {
         Trigger(B.Flag.Secondary, M.create()),
         Trigger(B.Flag.Primary, M.create({ control: true }))
     ], 'Camera center and focus', 'Click element using ${triggers}'),
+    clickResetCameraOnEmpty: Binding([
```
jpattle, line 32: I am unsure if these constitute a breaking change. Technically if someone else has overridden the bindings in their implementation of the plugin they would need to add these new bindings or it wouldn't work.

src/mol-plugin/behavior/dynamic/camera.ts
```diff
@@ -18,7 +19,7 @@ const B = ButtonsType;
 const M = ModifiersKeys;
 const Trigger = Binding.Trigger;
 
-const DefaultFocusLociBindings = {
+export const DefaultFocusLociBindings = {
```
jpattle, line 22: Exporting these allows for easy override in the plugin config:
```
const pluginSpec: PluginUISpec = {
  ...
  behaviours: [
    PluginSpec.Behavior(PluginBehaviors.Camera.FocusLoci, {
      bindings: {
        ...DefaultFocusLociBindings, // Preserve default settings
        clickResetCameraOnEmpty: Binding.Empty,
        clickResetCameraOnEmptySelectMode: Binding.Empty,
      }
    }
  ],
  ...
}
```

src/mol-plugin-ui/structure/measurements.tsx
```diff
@@ -208,13 +210,16 @@ export class MeasurementControls extends PurePluginUIComponent<{}, { isBusy: boo
             entries.push(this.historyEntry(history[i], i + 1));
         }
 
+        const shouldShowToggleHint = this.plugin.config.get(PluginConfig.Viewport.ShowSelectionMode);
+        const toggleHint = shouldShowToggleHint ? (<> (toggle <ToggleSelectionModeButton inline /> mode)</>) : null;
```
dsehnal, line 214:
```suggestion
        const toggleHint = shouldShowToggleHint ? (<>{' '}(toggle <ToggleSelectionModeButton inline /> mode)</>) : null;
```

src/mol-plugin/behavior/dynamic/camera.ts
dsehnal: This might be a breaking change indeed. Can you maybe have a global `const` with the default value and fallback to it inside the implementation if it's undefined?

src/mol-plugin/behavior/dynamic/camera.ts
```diff
@@ -28,6 +45,8 @@ const DefaultFocusLociBindings = {
         Trigger(B.Flag.Secondary, M.create()),
         Trigger(B.Flag.Primary, M.create({ control: true }))
     ], 'Camera center and focus', 'Click element using ${triggers}'),
+    clickResetCameraOnEmpty: undefined,
```
arose, line 48: I would prefer `Binding.Empty` instead of making them optional

arose (changes requested):
Some linting issues and two comments, otherwise looks good to me, thanks.

- Please use `Bindings.Empty`
- `src\mol-plugin-ui\structure\superposition.tsx` also shows the selection mode icon, so it would be good to hide it there as well like in measurements

src/mol-plugin/behavior/dynamic/camera.ts
jpattle: Hi Alexander, I originally had `Binding.Empty` but it was discussed [here](https://github.com/molstar/molstar/pull/742#discussion_r1132139975) that it results in an API change meaning it would be a breaking update. I can revert to what I had originally if needed.

jpattle:
@dsehnal @arose are you happy for this work to be merged in?

@arose please see [this comment](https://github.com/molstar/molstar/pull/742#discussion_r1141747870) regarding your outstanding comment

src/mol-plugin/behavior/dynamic/camera.ts
dsehnal: What I've meant that is that the code that refrences this property should fall back to the default value, e.g.

```ts
clickResetCameraOnEmpty: Binding.Empty`,
...

const clickResetBinding = params.clickResetCameraOnEmpty ?? DefaultValueForClickResetCameraOnEmpty;
```

The params for behaviors are saved in the snapshots, so this might have caused the binding being `undefined` for older states and raise an exception.

dsehnal:
@jpattle I am fine with merging this. Just please add the `Binding.Empty` back.

PS: sorry for the delay in review, been quite busy recently.

dsehnal (approved):
LGTM

PR #750, opened 2023-03-16
Marking improvements

dsehnal:
# Description

Only trigger `identify` when event originated from the canvas

## Actions

### DIFF
--- a/src/mol-canvas3d/helper/interaction-events.ts
+++ b/src/mol-canvas3d/helper/interaction-events.ts
@@ -197,9 +197,9 @@ export class Canvas3dInteractionHelper {
             this.drag(x, y, buttons, button, modifiers);
         });
 
-        input.move.subscribe(({ x, y, inside, buttons, button, modifiers, onElement: onTargetElement }) => {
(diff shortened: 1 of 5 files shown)


src/mol-util/input/input-observer.ts
```diff
@@ -202,6 +205,7 @@ type PointerEvent = {
     clientY: number
     pageX: number
     pageY: number
+    target: any
```
arose, line 208: any reason this is not Element or so?

src/mol-canvas3d/helper/interaction-events.ts
```diff
@@ -197,8 +197,12 @@ export class Canvas3dInteractionHelper {
             this.drag(x, y, buttons, button, modifiers);
         });
 
-        input.move.subscribe(({ x, y, inside, buttons, button, modifiers }) => {
+        input.move.subscribe(({ x, y, inside, buttons, button, modifiers, onElement }) => {
```
arose, line 200: might as well call leave in input-observer

arose: scratch that, it is good to have it

arose (approved):
simplified our own ui code due this change

Issue #746, opened 2023-03-15
Cartoon Nucleic acid polymers colored with Element Symbol theme   


giagitom:
Hi, I'm wondering if it is possible to change the default element symbol (Oxygen) used for cartoon nucleic acid polymers when coloring using ```Element Symbol``` theme. Changing it to Carbon would allow to use the ```Carbon color``` option also for nucleotides.

![image](https://user-images.githubusercontent.com/20627977/225347369-f978678f-28d6-439e-9e00-9fb201c6d58f.png)

giagitom:
@arose, @dsehnal would it be acceptable to change the trace atom names on RNA, DNA and PNA types like this? 

```ts
export const PolymerTypeAtomRoleId: { [k in PolymerType]: { [k in AtomRole]: Set<string> } } = {
...
[PolymerType.RNA]: {
   trace: new Set(['C3\'', 'C3*']), //OLD new Set(['O3\'', 'O3*']),
...
```

I could make a PR if there is interest on this.

This is the comparison on pdb ID 1BNA (**original trace on left**). It also looks nicer (in my opinion), due to the shorter base length. 

![image](https://user-images.githubusercontent.com/20627977/226406898-8e7405f5-d2b0-4e8c-8dcc-b925f787b78d.png)

arose:
I agree, addressed in #746 with a small fix for PNA.

The DNA/RNA trace atoms where carbons before. If I remember correctly it was changed after discussion with @MadCatX. @MadCatX do you remember? Is it ok for you to change it back to carbons?

MadCatX:
Yes, @arose @giagitom I recall. I proposed the change form C3' to O3' because I was argued that it provides a more accurate depiction of what the NA structure actually looks like. I don't believe that there is any universal consensus about this though. I can always change it back in my plugins if I need to.

giagitom:
> I don't believe that there is any universal consensus about this though.
@MadCatX I agree there is no consensus on this. For example PyMol has P as the default option but allows also for C3'.

@arose an other thing to consider is that in the visual with sugar ring that will be introduced (#727) the trace will touch directly the sugar ring, giving a not entirely perfect result.

Probably the best possible way would be to use C3' as trace and use O3' as trace position during the mesh construction (don't know if this is something easy to achieve though)?

MadCatX:
Just FTR, I have no objections against changing this back to C3' if it makes things easier in the broader context. I can always revert this change in my code if I need to.

[...]

PR #757, opened 2023-03-21
Trace tweaks

arose:
- Fix polymer trace shift for PNA
- Use carbon for all trace atoms (to play nicely with color themes)


arose:
Going to think about this some more wrt to discussion in #746 and #727

arose:
Closing in favor of #838

PR #763, opened 2023-03-27
Handle resizing viewer element when window remains the same size

### LINKED ISSUE #749: Resize not detected when container element changes size
Molstar handles resizing the window, but there can be cases where the container element is resized but the window is not resized (e.g. another element is added/removed that causes the viewer to shrink in size). I think this can be handled with https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver - I'll try to put up a PR soon

russellp17:
# Description
Use ResizeObserver to trigger `handleResize` when the canvas's parent element changes size. This handles changes that affect the viewer dimensions even if the window itself does not emit a resize event.

closes #749 

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Note that since we don't clearly distinguish between a public and private interf
 
 
 ## [Unreleased]
+- Handle resizes of viewer element even when window remains the same size
 
 - Input/controls improvements
     - Move or fly around the scene using keys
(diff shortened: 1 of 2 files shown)


dsehnal (commented):
Looks good overall, but I think we should still keep a fallback for the old method in case the resize observer is not defined, which might be the case for example for older tablets. Can you please include that?

russellp17:
> I think we should still keep a fallback for the old method in case the resize observer is not defined, which might be the case for example for older tablets

👍 added

PR #772, opened 2023-03-31
Forsake lazy imports

midlik:
# Description
- Removed LazyImports (introduced in PR #700) as it was generally problematic. Instead, the caller code must import necessary module add pass them to the constructor of HeadlessPluginContext/HeadlessScreenshotHelper.

## Actions


package.json
```diff
@@ -110,7 +110,6 @@
     "@graphql-codegen/typescript-operations": "^3.0.2",
     "@types/cors": "^2.8.13",
     "@types/gl": "^6.0.2",
-    "@types/jpeg-js": "^0.3.7",
```
dsehnal, line 113: Would this break the build if the optional dependencies are not installed? (I think it would)

package-lock.json
dsehnal, line 1: Did you have the latest version of the repo forked before you made the PR? This diff doesn't make much sense for removing one typing extension.

dsehnal (commented):
Please fix the linting before merging and I think we might have to undo the removal of the package?

package.json
arose: also there is a looming issue that the packages might provide their own typing and @types would not be available/outdated

midlik: I apologize. It will not work.
Previously, `npm run build` was telling me this: 
`npm WARN deprecated @types/jpeg-js@0.3.7: This is a stub types definition. jpeg-js provides its own type definitions, so you do not need this installed.` 
So I tried removing `@types/jpeg-js`, running `npm install --omit=optional`, and building and it worked. But perhaps I just forgot to delete something. Now I tried again and it doesn't work.
I rolled this change back.

package-lock.json
midlik: I put `@types/jpeg-js` back so this is not relevant anymore.

midlik:
Thanks for feedback, it should now be all sorted out.

package.json
dsehnal: You prolly had the older version of `node_modules`

dsehnal (approved):

PR #779, opened 2023-04-06
Fix broken rendering caused by changes in 291d7abb7804d58e2ed4a35601715d50e6d4b8b8

MadCatX:
Reuse of vertex arrays among GL programs seems to break rendering. With commit 291d7abb7804d58e2ed4a35601715d50e6d4b8b8 applied, cartoon visuals look like this

paint_it_black
and it's not easy to face up when the whole structure is black. Suggested partial revert fixes the problem for me.

This issue appears to be specific to only some OpenGL driver because amongst the computers I have at hand ATM I can only reproduce it on a laptop with AMD Vega 8 and Mesa 23.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ Note that since we don't clearly distinguish between a public and private interf
 - Avoid `renderMarkingDepth` for fully transparent renderables
 - Remove `camera.far` doubeling workaround
 - Add `ModifiersKeys.areNone` helper function
-- Fix broken rendering caused by vertex array reuse
+- Fix rendering issues caused by VAO reuse
 
 ## [v3.33.0] - 2023-04-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Note that since we don't clearly distinguish between a public and private interf
 - Avoid `renderMarkingDepth` for fully transparent renderables
 - Remove `camera.far` doubeling workaround
 - Add `ModifiersKeys.areNone` helper function
+- Fix broken rendering caused by vertex array reuse
 
 ## [v3.33.0] - 2023-04-02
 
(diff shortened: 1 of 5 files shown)


arose:
Thanks for investigating this. @dsehnal reported (second hand) the same but I don't have a machine on which I can reproduce this. Maybe some driver optimize an attribute away or change the order so that normals are zero. Would have been nice to reduce VAO count. Does picking still work? What happens if you enable the FlipSided parameter? Is only cartoon broken or any mesh geometry?

MadCatX:
After some more testing I think we can "narrow" the problem down a bit. I can only reproduce it with Firefox and Mesa OpenGL. On all computers that I could test with, I get black structure with the following visuals:

- Cartoon
- Molecular surface
- Orientation
- Putty
- Non-covalent interactions
- Validation clashes
- Membrane orientation
- Confal pyramids
- NtC tube

With FlipSide enabled I get some color back but it looks like a blend between black and the expected color. The structure also looks the same both from inside and outside. Picking cues seem to be rendered fine.

Additionally, on an older laptop with Intel HD3000 GPU I also get the "Gaussian surface" visual rendered all black.

The list HW/SW configs I tested with so far:
| GPU | Driver | Browser | Results
|-------|-----------|------------|-----------
| AMD Vega 8 | Mesa 23.0.1 (radeonsi) | Firefox 111 | ❌
| AMD Vega 8 | Mesa 23.0.1 (radeonsi) | Chromium 111(?) | ✔️
| AMD RX580 | Mesa 22.3.7 (radeonsi)  | Firefox 111 | ❌
| AMD RX580 | Mesa 22.3.7 (radeonsi)  | Chromium 111 | ✔️
| Intel HD3000 | Mesa 23.0.1  (i915)  | Firefox 111/112b | ❌
| Intel HD3000 | Mesa 23.0.1  (i915)  | Chromium 111 | ✔️
| nVidia GTX 1050 | nVidia Windows (don't recall version)  | Firefox 111 | ✔️

Not reusing VAOs fixes the problem everywhere AFAICT.

arose:
Thanks @MadCatX, merging this

PR #782, opened 2023-04-08
add eachLocation to representation/visual interface

arose:
Add `.eachLocation` to representation/visual interface for when `.getAllLoci` is too coarse.

### DIFF
--- a/src/mol-model/structure/structure/element/loci.ts
+++ b/src/mol-model/structure/structure/element/loci.ts
@@ -148,7 +148,7 @@ export namespace Loci {
      * The loc argument of the callback is mutable, use Location.clone() if you intend to keep
      * the value around.
      */
-    export function forEachLocation(loci: Loci, f: (loc: Location) => any) {
+    export function forEachLocation(loci: Loci, f: (loc: Location) => void) {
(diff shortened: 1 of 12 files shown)


src/mol-model/structure/structure/element/loci.ts
```diff
@@ -148,7 +148,7 @@ export namespace Loci {
      * The loc argument of the callback is mutable, use Location.clone() if you intend to keep
      * the value around.
      */
-    export function forEachLocation(loci: Loci, f: (loc: Location) => any) {
+    export function forEachLocation(loci: Loci, f: (loc: Location) => void) {
```
dsehnal, line 151: Why the `void` here? I think it's ok if the lambda returns a value if we just discard it.

src/mol-repr/util.ts
```diff
@@ -45,6 +46,8 @@ export namespace VisualUpdateState {
     }
 }
 
+export type LocationCallback = (loc: Location, isSecondary: boolean) => void
```
dsehnal, line 49: `=> any`?

dsehnal (approved):

PR #787, opened 2023-04-13
Address Node incompatibility in `mol-util/file-info`

russellp17:
# Description
Changes `mol-util/file-info` to avoid `instanceof File` checks so that it runs on Node. I also added test cases to document current behavior, renamed the helper function + interface for clarity, and removed the unused `src` property from the return value - there seem to be several other unused properties but I'm hesitant to remove them in case I'm missing something.

Also changed a `window.location.hostname` comparison check that happens at the top-level so that it doesn't error on Node if it's not polyfilled.

Verified by downloading a file from URL and uploading a file on the example viewer; some of the types work out to `any` so I'm not 100% confident that it's safe in all codepaths.

I think ideally there wouldn't be any `new File` or `new Blob` calls for compatibility and those objects could be replaced by interfaces that are implemented by web-compatible versions and Node-compatible versions (like `FileHandle`).

## Actions

### DIFF
--- a/src/extensions/volumes-and-segmentations/index.ts
+++ b/src/extensions/volumes-and-segmentations/index.ts
@@ -19,7 +19,7 @@ import { VolsegEntryFromRoot, VolsegGlobalStateFromRoot, VolsegStateFromEntry } 
 import { VolsegUI } from './ui';
 
 
-const DEBUGGING = window.location.hostname === 'localhost';
+const DEBUGGING = window?.location?.hostname === 'localhost';
 
 export const VolsegVolumeServerConfig = {
     // DefaultServer: new PluginConfigItem('volseg-volume-server', DEFAULT_VOLUME_SERVER_V2),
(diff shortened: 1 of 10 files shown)


src/extensions/volumes-and-segmentations/index.ts
```diff
@@ -19,7 +19,7 @@ import { VolsegEntryFromRoot, VolsegGlobalStateFromRoot, VolsegStateFromEntry }
 import { VolsegUI } from './ui';
 
 
-const DEBUGGING = window.location.hostname === 'localhost';
+const DEBUGGING = window?.location?.hostname === 'localhost';
```
dsehnal, line 22: This one might not be enough. Maybe `type window !== 'undefined' && window !== null ? window.location?.hostname === 'localhost' : false` ?

dsehnal (approved):
Looks good, perhaps just that one tweak?

src/extensions/volumes-and-segmentations/index.ts
russellp17: yeah was just going to add a note/update this - `window` will be undefined if it's not polyfilled

PR #796, opened 2023-04-20
Add a uniform color theme for NtC tube that still paints the residue and segment dividers in a different color

MadCatX:
With some practical experience we found out that having the NtC tube painted accordingly to NtCs could make the structure look "busy" and difficult to read in some situations. This patch adds an alternative color theme where everything but the residue and step boundaries are painted in uniform color.

image
vs.
image


arose:
lgtm

PR #797, opened 2023-04-22
support ignoreHydrogens for interactions

arose:
Hydrogens: show all, only polar, hide all

image

image

image


arose:
@dsehnal good to go?

src/mol-model-props/computed/representations/interactions-intra-unit-cylinder.ts
```diff
@@ -39,22 +41,64 @@ async function createIntraUnitInteractionsCylinderMesh(ctx: VisualContext, unit:
...
+                        pos(elements[idx], p);
+                        const dist = Vec3.distance(p, pB);
+                        if (dist < minDistA) {
+                            minDistA = dist;
+                            Vec3.copy(posA, p);
+                        }
+                    }
+                });
+
+                eachIntraBondedAtom(unit, idxB, (_, idx) => {
```
dsehnal, line 82: Creating two closures inside a loop isn't exactly cheap. Tho I guess it shouldn't be a problem in this particular case.

If you feel like it in the future, there is a way to avoid creating the closure with something like this:

```ts
function eachIntraBondedAtom(unit, idx, cb, thisArg?: State) {
   for (const b of ...) {
     cb.call(thisArg, x, y);
   }
}
```

and then the callback can be a top-level function like this:

```ts
function eachBond(this: State, x, y) {
  this.whatever = x;
}
```

dsehnal (approved):

src/mol-model-props/computed/representations/interactions-intra-unit-cylinder.ts
arose: good point, opened #813

PR #800, opened 2023-04-23
Implementing 2 colors interpolation on impostor cylinders

giagitom:
# Description

- Implement 2 colors interpolation on impostor cylinders.
- Allow for single aromatic dash count

image

- exposed setting under shading category to toggle interpolation
image
 

## Actions


arose:
Interesting, thanks @giagitom I will have a look. Could you please make "Allow for single aromatic dash count" a separate PR?

giagitom:
> Interesting, thanks @giagitom I will have a look. Could you please make "Allow for single aromatic dash count" a separate PR?

Hi @arose, done.

src/mol-geo/geometry/color-data.ts
```diff
@@ -156,11 +157,19 @@ function createInstanceColor(locationIt: LocationIterator, color: LocationColor,
 /** Creates color texture with color for each group (i.e. shared across instances) */
 function createGroupColor(locationIt: LocationIterator, color: LocationColor, colorData?: ColorData): ColorData {
     const { groupCount } = locationIt;
-    const colors = createTextureImage(Math.max(1, groupCount), 3, Uint8Array, colorData && colorData.tColor.ref.value.array);
+    const colors = createTextureImage(Math.max(1, groupCount * 2), 3, Uint8Array, colorData && colorData.tColor.ref.value.array);
```
arose, line 160: We need to find a way to only double the size when using. Also I don't like haveing a specific type of location exposed here. One idea is to add an optional getLocation2 to the LocationIterator and a property hasLocation2 which you can use here.

src/mol-geo/geometry/color-data.ts
```diff
@@ -156,11 +157,19 @@ function createInstanceColor(locationIt: LocationIterator, color: LocationColor,
 /** Creates color texture with color for each group (i.e. shared across instances) */
 function createGroupColor(locationIt: LocationIterator, color: LocationColor, colorData?: ColorData): ColorData {
```
arose, line 158: would need support in `createGroupInstanceColor` as well

src/mol-gl/shader/cylinders.vert.ts
```diff
@@ -50,6 +54,7 @@ void main() {
     vEnd = (modelTransform * vec4(aEnd, 1.0)).xyz;
     vSize = size * aScale;
     vCap = aCap;
+    vColorMode = aColorMode;
```
arose, line 57: I think you can pre-mix vColor and vColor2 here already instead of in the frag shader, i.e., set vColor and vColor2 to values here in a way you don't need a branch when mixing/interpolating them in the frag shader.

src/mol-gl/shader/cylinders.frag.ts
```diff
@@ -9,6 +9,7 @@ precision highp float;
 precision highp int;
 
 #define bumpEnabled
+#define multiColor
```
arose, line 12: Could this be a define that is only set for visuals that need it? Also, how about calling it `dualColor`?

src/mol-geo/geometry/cylinders/cylinders-builder.ts
```diff
@@ -58,17 +61,20 @@ export namespace CylindersBuilder {
                 } else {
                     Vec3.add(tmpVecB, tmpVecA, tmpDir);
                 }
-                add(tmpVecA[0], tmpVecA[1], tmpVecA[2], tmpVecB[0], tmpVecB[1], tmpVecB[2], radiusScale, topCap, bottomCap, group);
+                if (interpolate) {
+                    colorMode = Vec3.distance(start, tmpVecB) / (d * 2);
```
arose, line 65: Can you make it so that colorMode is strictly an integer (as union type)? At some point it would be good to pack colorMode and cap into a single bit field and integers would help.

arose (changes requested):

arose (changes requested):
This looks nice, thanks @giagitom, a few comments inline. The main aspect to solve is avoiding doubeling of memory for the colors when not used. I am ok to have this only support for impostor cylinders. Support for mesh cylinders can be implemented later as needed.

giagitom:
Hi @arose, thanks for the review, I'll start working on this as soon as possible

src/mol-geo/geometry/color-data.ts
giagitom: @arose I'm trying to implement this logic, I need a way to update ```createLocationIterator``` or pass an additional variable to it in order to use single or double location iterator depending on the ```colorMode``` property. Do you think it is possible in some way?

giagitom: @arose, this should be done with last commit.

src/mol-gl/shader/cylinders.frag.ts
giagitom: Done with last commit

src/mol-gl/shader/cylinders.vert.ts
giagitom: Hmm.. I'm not sure what you mean here.

src/mol-geo/geometry/cylinders/cylinders-builder.ts
giagitom: @arose The problem here is that I am using colorMode in two ways. 

1. Dashed cylinders are colored with an uniform color using colorMode from 0.0 - 1.0 to know the blending factor to be used for each dashed cylinder. I need this because at this point I have no way to know the full cylinder (the non-dashed one) length necessary for a correct interpolation.
2. Full length cylinders instead are interpolated and use colorMode = 3.0

For this reason, in my understanding I can't avoid using the float value. 
How about retaining colorMode as a float, utilizing the integer component to encode caps, and assigning the fractional part for the percentage?

src/mol-geo/geometry/color-data.ts
arose: thanks, looks good, just did a few tweaks to the locate-iterator to consolidate the code

src/mol-gl/shader/cylinders.vert.ts
arose: went ahead and did it in da034d950269f28748a76b1d40f23fcd3c8fe7c9

src/mol-geo/geometry/cylinders/cylinders-builder.ts
arose: OK, makes sense, thanks for the explanation

arose (approved):
Hi @giagitom, this is now good from my side. Please have a look if the changes I made are OK for you.

giagitom:
Hi @arose, thanks for the changes. Everything looks good to me. 

Just a curiosity, I'm very surprised about da034d9. I wasn't aware this could be done at vertex level. 

```vColor.rgb = mix(vColor.rgb, color2.rgb, mix(-0.25, 1.25, f / 1.5));```

In my understanding `f` can be 0.0 or 1.0; and `vColor` mix factor, depending from `f` only, could have a maximum of 2 different values. So how the gradient could be generated in this way at fragment level?

File: src/mol-gl/shader/cylinders.vert.ts [hunk 6]
```diff
@@ -79,6 +79,18 @@ void main() {
         gl_Position.z = (uProjection * mvPosition).z;
     }
 
+    #if defined(dDualColor) && defined(dRenderVariant_color) && (defined(dColorType_group) || defined(dColorType_groupInstance))
+        // dual-color mixing
+        // - for aColorMode between 0 and 1 use aColorMode to interpolate
+        // - for aColorMode == 2 do nothing, i.e., use vColor
+        // - for aColorMode == 3 use position on cylinder axis to interpolate
+        if (aColorMode <= 1.0){
```
giagitom, line 87: What about testing here for special case in which vColor === color2 ? Could this improve performances?

arose: maybe, maybe not, it would be very small and hard to measure since generally the bottleneck is in the fragment shader

arose:
> In my understanding `f` can be 0.0 or 1.0; and `vColor` mix factor, depending from `f` only, could have a maximum of 2 different values. So how the gradient could be generated in this way at fragment level?

vColor is a varying, so its values in the fragment shader come from linear interpolation of the values at the three triangle vertices.

giagitom:
> vColor is a varying, so its values in the fragment shader come from linear interpolation of the values at the three triangle vertices.

Now I understand, thanks for the explanation.

PR #802, opened 2023-04-24
struct-conn extension

midlik:
# Description

## Actions


src/extensions/struct-conn/index.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Copyright (c) 2018-2023 mol* contributors, licensed under MIT, See LICENSE file for more info.
+ *
+ * @author Adam Midlik <midlik@gmail.com>
+ */
+
+import { PluginContext } from '../../mol-plugin/context';
+
+
+export namespace StructConnExtensionFunctions {
```
dsehnal, line 10: We are trying to move away from `namespace` usage. Please export this for example as a `export const StructConnExtensions`.

src/extensions/struct-conn/index.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Copyright (c) 2018-2023 mol* contributors, licensed under MIT, See LICENSE file for more info.
```
dsehnal, line 2:
```suggestion
 * Copyright (c) 2023 mol* contributors, licensed under MIT, See LICENSE file for more info.
```

src/apps/viewer/app.ts
```diff
@@ -508,3 +509,7 @@ export const ViewerAutoPreset = StructureRepresentationPresetProvider({
         }
     }
 });
+
+export const ExtensionData = {
+    'struct-conn': StructConnExtensionFunctions,
+};
```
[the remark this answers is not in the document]
midlik: OK.

src/extensions/struct-conn/index.ts
midlik: OK.

src/apps/viewer/app.ts
```diff
@@ -48,6 +48,7 @@ import '../../mol-util/polyfill';
 import { ObjectKeys } from '../../mol-util/type-helpers';
 import { SaccharideCompIdMapType } from '../../mol-model/structure/structure/carbohydrates/constants';
 import { Backgrounds } from '../../extensions/backgrounds';
+import { StructConnExtensionFunctions } from '../../extensions/struct-conn';
```
dsehnal, line 51: Can you please move this to the `extensions/wwpdb/struct-conn` folder and rename it to `wwPDBStructConnExtensionFunctions`.

src/apps/viewer/app.ts
```diff
@@ -508,3 +509,7 @@ export const ViewerAutoPreset = StructureRepresentationPresetProvider({
         }
     }
 });
+
+export const ExtensionData = {
+    StructConn: StructConnExtensionFunctions,
+};
```
dsehnal, line 515:
```suggestion
export const PluginExtensions = {
    wwPDBStructConn: wwPDBStructConnExtensionFunctions,
};
```

src/extensions/struct-conn/index.ts
```diff
@@ -0,0 +1,365 @@
...
+            RESIDUES_VISUAL_PARAMS,
+            { tags: [TAGS.RESIDUE_REPR] }
+        ).commit();
+    }
+    if (selectionByAtom.isOk) {
+        const atomBalls = await plugin.build().to(selectionByAtom).apply(
+            StructureRepresentation3D,
+            ATOMS_VISUAL_PARAMS,
+            { tags: [TAGS.ATOM_REPR] }
+        ).commit();
```
dsehnal, line 269: You should be able to do all of this in a single commit. If a selection is empty, the visual will be skipped (hopefully :)).

At the very least, the selections can be committed together and so can the the visuals.

src/extensions/struct-conn/index.ts
```diff
@@ -0,0 +1,365 @@
...
+}
+function getExtensionState(plugin: PluginContext): ExtensionState {
+    return (plugin.customState as any).StructConn ??= createExtensionState(plugin);
+}
+
+
+/** Map in which keys are refs of plugin state tree nodes,
+ * and they get automatically deleted from the map
+ * when the node is deleted from the tree. */
+class NodeMap<T> extends PluginBehavior.WithSubscribers<undefined> {
```
dsehnal, line 329: Why do you need this? It seems to me you can always query the current state tree to get all the info you need to change it. Extract struct conn can be cached directly at the structure level.

src/extensions/struct-conn/index.ts
```diff
@@ -0,0 +1,365 @@
...
+    distance: number,
+    partner1: StructConnPartner,
+    partner2: StructConnPartner,
+}
+
+
+/** Return an object with all struct_conn records read from mmCIF.
+ * Return {} if structure comes from another format than mmCIF.
+ */
+function extractStructConns(structure: Structure): { [id: string]: StructConnRecord } {
```
dsehnal, line 155:
```suggestion
function extractStructConns(model: Model): { [id: string]: StructConnRecord } {
```

and cache the result in `model._staticPropertyData['wwpdb-struct-conn-extension-data']`. Then you don't need the plugin behavior.

Or just add `function getStructConn(model: Model, id: string)?: StructConn` and query that each time you need it. No need for cache then.

src/extensions/struct-conn/index.ts
```diff
@@ -0,0 +1,365 @@
...
+
+type StructNode = ReturnType<typeof selectStructureNode>
+
+/** Return the first structure node belonging to `entry` (e.g. '1tqn'),
+ * or to the first loaded structure node overall if `entry` is `undefined`.
+ * Includes only "root" structures, not structure components. */
+function selectStructureNode(plugin: PluginContext, entry: string | undefined) {
+    const structNodes = plugin.state.data
+        .selectQ(q => q.ofType(PluginStateObject.Molecule.Structure))
+        .filter(node => node.obj && !node.obj.data.parent && !node.transform.transformer.definition.isDecorator);
```
dsehnal, line 112:
```suggestion
        .selectQ(q => q.ofType(PluginStateObject.Molecule.Structure))
        .filter(node => node.obj && !node.obj.data.parent && !node.transform.transformer.definition.isDecorator);
```

You should filter on `PluginStateObject.Molecule.Model` directly as that is where you are looking for the id anyway. Then you can find a suitable child structure node to create your representation on. `plugin.managers.structure.hierarchy` should help there quite a bit.

src/extensions/struct-conn/index.ts
```diff
@@ -0,0 +1,365 @@
...
+    const hiddenNodes = getExtensionState(plugin).hiddenNodes;
+    for (const nodeRef of Array.from(hiddenNodes.keys())) {
+        try {
+            setSubtreeVisibility(plugin.state.data, nodeRef, false); // false means visible
+        } catch {
+            // this is OK, the node has been removed
+        }
+    }
+    hiddenNodes.clear();
+}
```
dsehnal, line 308: You are only hiding or showing all SFNG nodes, no need for the `getExtensionState` here at all.

src/extensions/struct-conn/index.ts
```diff
@@ -0,0 +1,365 @@
...
+/** Remove anything created by `addStructConnInspection` */
+async function removeAllStructConnInspections(plugin: PluginContext) {
+    const selNodes = [
+        ...plugin.state.data.selectQ(q => q.root.subtree().withTag(TAGS.RESIDUE_SEL)),
+        ...plugin.state.data.selectQ(q => q.root.subtree().withTag(TAGS.ATOM_SEL)),
+    ];
+    for (const node of selNodes) {
+        await plugin.build().delete(node).commit();
+    }
+}
```
dsehnal, line 283:
```suggestion
function removeAllStructConnInspections(plugin: PluginContext) {
    const selNodes = [
        ...plugin.state.data.selectQ(q => q.root.subtree().withTag(TAGS.RESIDUE_SEL)),
        ...plugin.state.data.selectQ(q => q.root.subtree().withTag(TAGS.ATOM_SEL)),
    ];
    const update = plugin.build();
    for (const node of selNodes) {
        update.delete(node);
    }
    return update.commit();
}
```

dsehnal (changes requested):

src/apps/viewer/app.ts
midlik: OK.

src/apps/viewer/app.ts
midlik: OK.

src/extensions/struct-conn/index.ts
midlik: > If a selection is empty, the visual will be skipped (hopefully)
You're right, it will be skipped if it's in one commit (but it it's two commits and the selection is empty, the visual would apply to the whole structure!).
I was adding more visuals in the older version, that's why I split the commits.

midlik: I will do this:
```
    const residueVis = await plugin.build().to(structNode).apply(StructureComponent,...).apply(StructureRepresentation3D, ...).commit();
    const atomVis = await plugin.build().to(structNode).apply(StructureComponent,...).apply(StructureRepresentation3D, ...).commit();
```

dsehnal: I think that's even worse. If you are doing both, you can do:

```ts
const update = plugin.build()
const residueVis = update.to(structNode).apply(StructureComponent,...).apply(StructureRepresentation3D, ...).selector;
const atomVis = ....
await update.commit();
```

dsehnal: > (but it it's two commits and the selection is empty, the visual would apply to the whole structure!).

This is actually a bug then.

src/extensions/struct-conn/index.ts
midlik: Good point. I'll change that.

src/extensions/struct-conn/index.ts
midlik: I don't see how it would make things easier. It would add extra work of finding the suitable child structure node.

src/extensions/struct-conn/index.ts
midlik: > cache the result in model._staticPropertyData['wwpdb-struct-conn-extension-data']
I like this idea, but the `_staticPropertyData` doc says "Not to be accessed directly, each custom property descriptor defines property accessors that use this field to store the data." Should I just ignore it?

dsehnal: Reality is not perfect. You can ignore this. The only issue this could cause if some other extension called the property `'wwpdb-struct-conn-extension-data'` too. See for example ``https://github.com/molstar/molstar/blob/master/src/extensions/pdbe/preferred-assembly.ts#L23` for how we use it.

src/extensions/struct-conn/index.ts
dsehnal: And now you are doing an extra work to incorrectly filter out the decorators. You should actually be applying your selection to the lowest decorator of the root, and not the root itself. There could be things like transforms in the decorators.

src/extensions/struct-conn/index.ts
dsehnal: Add `entry: string` parameter to the hide function and use it to find the SFNG only in that subtree.

src/extensions/struct-conn/index.ts
midlik: This was to ensure that cached data that relate to a node are removed when that node is deleted.
I use in 2 places: for caching the parsed struct_conns, and for keeping track of the hidden SNFG nodes. The first could be probably moved to `model._staticPropertyData`. The latter I'm not sure about - it could cause weird behavior if removed (https://github.com/molstar/molstar/pull/802#discussion_r1179131514).

src/extensions/struct-conn/index.ts
midlik: OK, somehow I was struggling with getting the atomVis node then (to zoom on it), but now I found a way though refs.

midlik: Should be fixed now.

src/extensions/struct-conn/index.ts
midlik: I managed to get rid of the need for NodeMap and removed it. Thanks for help.

src/extensions/struct-conn/index.ts
midlik: Should be fixed now.

src/extensions/struct-conn/index.ts
midlik: Should be fixed now.

src/extensions/struct-conn/index.ts
midlik: I checked it now, and even though I apply the selections to the root structure, in reality they get applied to the decorator. Is there maybe some smart magic going on?

You can see, that the selections are hanging on `0LcJJj-WDrnJZbPNBuKi-A`, not `eYTXqUzrVd9dEiAWUgQA0g`:
```
              {
                "parent": "tA8TVGNV9FUPCqdX-KZY4Q",
                "transformer": "ms-plugin.structure-from-model",
                "ref": "eYTXqUzrVd9dEiAWUgQA0g",
              },
              {
                "parent": "eYTXqUzrVd9dEiAWUgQA0g",
                "transformer": "ms-plugin.custom-structure-properties",
                "ref": "0LcJJj-WDrnJZbPNBuKi-A",
              },
              {
                "parent": "0LcJJj-WDrnJZbPNBuKi-A",
                "transformer": "ms-plugin.structure-component",
                "tags": [
                  "structconn-focus-residue-sel"
                ],
                "ref": "AWuKbfKGb0LAkHLoEAN_ZQ",
              },
              {...},
              {
                "parent": "0LcJJj-WDrnJZbPNBuKi-A",
                "transformer": "ms-plugin.structure-component",
                "tags": [
                  "structconn-focus-atom-sel"
                ],
                "ref": "YdB3_WI7VXbrfRsd2GaAeQ",
              },
              {...}
```

src/apps/viewer/index.html
```diff
@@ -38,6 +38,19 @@
...
+            <button onclick="molstarViewer?.plugin && molstar.ExtensionData.wwPDBStructConn.inspectStructConn(molstarViewer.plugin, undefined, 'covale6');">covale6</button>
+            <button onclick="molstarViewer?.plugin && molstar.ExtensionData.wwPDBStructConn.inspectStructConn(molstarViewer.plugin, undefined, 'disulf1');">disulf1</button>
+            <button onclick="molstarViewer?.plugin && molstar.ExtensionData.wwPDBStructConn.inspectStructConn(molstarViewer.plugin, undefined, 'disulf4');">disulf4</button>
+            <button onclick="molstarViewer?.plugin && molstar.ExtensionData.wwPDBStructConn.inspectStructConn(molstarViewer.plugin, undefined, 'disulf6');">disulf6</button>
+            <button onclick="molstarViewer?.plugin && molstar.ExtensionData.wwPDBStructConn.inspectStructConn(molstarViewer.plugin, undefined, 'disulf9');">disulf9</button>
+            <button onclick="molstarViewer?.plugin && molstar.ExtensionData.wwPDBStructConn.clearStructConnInspections(molstarViewer.plugin, undefined);">Clear</button>
+        </div>
```
dsehnal, line 53: Please remove this. The default viewer instance is no place to showcase extensions like this.

The example in the .md file will have to be enough for now.

src/extensions/wwpdb/struct-conn/index.ts
```diff
@@ -0,0 +1,295 @@
...
+
+    const atomsVisRef = update.to(structNode).apply(
+        StructureComponent,
+        { label: `${conn.id} (atoms)`, type: { name: 'expression', params: expressionByAtoms } },
+        { tags: [TAGS.ATOM_SEL] }
+    ).apply(
+        StructureRepresentation3D,
+        ATOMS_VISUAL_PARAMS,
+        { tags: [TAGS.ATOM_REPR] }
+    ).ref;
```
dsehnal, line 257: You can use `.selector` and then you dont have to do the `.select(atomsVisRef)[0]` below.

docs/extensions/struct-conn.md
```diff
@@ -0,0 +1,51 @@
...
+        </div>
+        <div id="app" style="position: absolute; top: 10%; width: 100%; height: 90%;"></div>
+        <script type="text/javascript" src="./molstar.js"></script>
+        <script type="text/javascript">
+            var molstarViewer;
+            molstar.Viewer.create('app', { layoutIsExpanded: false }).then(viewer => {
+                molstarViewer = viewer;
+                viewer.loadPdb('5elb');
+            });
+        </script>
```
dsehnal, line 46: This example isn't very useful for the usecase they want this in. I suggest something like this:

```html
<div style="position: absolute; width: 100%; height: 10%; padding-block: 10px;" id="struct_conns">
   <button onclick="clear">CLEAR</button>
</div>
```

```js
var pdbId = '5elb';
function clear() {
  molstar.PluginExtensions....
}
var molstarViewer;
molstar.Viewer.create('app', { layoutIsExpanded: false }).then(viewer => {
  viewer.loadPdb(pdbId).then(() => {
     var structConns = inspect(..., pdbId);
     var controls = document.getElementById('struct_conns');
     structConns.forEach(conn => {
       var button = document.createElement('button');
       button.addEventListener('click', ...);
       button.innerText = ...;
       controls.appendChild(button);
    });
  });
});
```

src/apps/viewer/app.ts
```diff
@@ -508,3 +509,7 @@ export const ViewerAutoPreset = StructureRepresentationPresetProvider({
         }
     }
 });
+
+export const ExtensionData = {
```
dsehnal, line 513: For the 3rd time:

```suggestion
export const PluginExtensions = {
```

dsehnal (changes requested):

src/apps/viewer/app.ts
midlik: OK.

src/apps/viewer/index.html
midlik: Oh, sorry, I forgot to remove it.

src/extensions/wwpdb/struct-conn/index.ts
midlik: Cool, thanks.

dsehnal (approved):

PR #804, opened 2023-04-26
Enable odd dash count (1,3,5)

giagitom:
# Description

Allows for odd aromatic dash count (1,3,5)

## Actions


arose:
Haven't looked in detail but in general this should also be supported for lines and mesh-based cylinders

giagitom:
Yes, I'll do that, but prefer to wait for your feedback before, in case there is something to be changed

arose:
If you want to be fancy you can hide the caps of the dashes that meet for odd counts (unless they are stubs).

image

Other wise looks good, apart from that it should also be supported for lines and mesh-based cylinders as already mentioned.

arose (changes requested):
If you want to be fancy you can hide the caps of the dashes that meet for odd counts (unless they are stubs).

image

Other wise looks good, apart from that it should also be supported for lines and mesh-based cylinders as already mentioned.

Would be nice to allow odd dash counts for non-aromatic dashes, too, for example, used in interactions.

giagitom:
Should have done everything with last commit. Also supporting non aromatic odd dash count.

image

image

arose (changes requested):
nice changes, thank you!

dashes lines are not looking quite right

image

please add a changelog entry

giagitom:
Should be fixed now for meshes, lines and impostors.

image

arose:
Thanks for the changes. The aromatic dashes look different for impostor and non-imposter variant. They should look like the non-imposter variant where the dashes do not overlap.

image

image

giagitom:
It should be equal now.

arose (approved):
thanks @giagitom, quite nice, looks all good to me now

PR #807, opened 2023-04-27
State.updateNode fix for Null parents

dsehnal:
# Description

Fixes and edge case when trying to apply transforms for null nodes. Fixes:

```ts
const sel = await createSelection(...); // returns StateObject.Null since the selection was empty
// ... later
const repr = plugin.build().to(sel).apply(Representation3D, ...); // creates representation for the whole structure
```

## Actions


arose (approved):
lgtm

PR #808, opened 2023-04-28
Partial atomic charges extension

MergunFrimen:
# Description

This is an extension for coloring and labeling atoms and residues according to partial atomic charges that are read from a mmCIF file (added example mmCIF file with the custom categories to /examples).

## Actions


src/extensions/sb-ncbr/partial-charges/labels.ts
```diff
@@ -0,0 +1,39 @@
...
+                return;
+            }
+
+            const atomId = StructureProperties.atom.id(loc);
+            const model = loci.structure.model;
+            const data = SbNcbrPartialChargesPropertyProvider.get(model).value?.data;
+            if (data === undefined) return;
+            const { typeIdToAtomIdToCharge, typeIdToResidueToCharge } = data;
+
+            const typeId = SbNcbrPartialChargesPropertyProvider.getParams(model).typeId.defaultValue;
```
dsehnal, line 27: This looks really weird. Why are you getting the current id from the default value? I don't think this will show the proper tooltip.

src/extensions/sb-ncbr/partial-charges/labels.ts
```diff
@@ -0,0 +1,39 @@
...
+            const showResidueCharge = granularity === 'residue';
+            const charge = showResidueCharge
+                ? typeIdToResidueToCharge.get(typeId)?.get(atomId)
+                : typeIdToAtomIdToCharge.get(typeId)?.get(atomId);
+            const label = granularity === 'residue' ? 'Residue charge' : 'Atom charge';
+
+            return `<strong>${label}: ${charge?.toFixed(4) || 'undefined'}</strong>`;
```
dsehnal, line 34: Do you really always want to show 4 decimals? E.g. `0.0000` ?

src/extensions/sb-ncbr/partial-charges/labels.ts
```diff
@@ -0,0 +1,39 @@
...
+        },
+        group: (label: LociLabel): string => (label as string).toString().replace(/Model [0-9]+/g, 'Models'),
+        priority: 0,
```
dsehnal, line 37: Why priority 0? I don't think this should be shown 1st even above the standard description.

src/extensions/sb-ncbr/partial-charges/preset.ts
```diff
@@ -0,0 +1,32 @@
...
+        return !!a.data.models.some((m) => SbNcbrPartialChargesPropertyProvider.isApplicable(m));
+    },
+    params: () => StructureRepresentationPresetProvider.CommonParams,
+    async apply(ref, params, plugin) {
+        const structureCell = StateObjectRef.resolveAndCheck(plugin.state.data, ref);
+        const structure = structureCell?.obj?.data;
+        if (!structureCell || !structure) return {};
+
+        const colorTheme = SbNcbrPartialChargesColorThemeProvider.name as any;
+        return await PresetStructureRepresentations.auto.apply(
```
dsehnal, line 26: The `await` here is redundant.

src/extensions/sb-ncbr/partial-charges/property.ts
```diff
@@ -0,0 +1,193 @@
...
+import { CustomProperty } from '../../../mol-model-props/common/custom-property';
+import { PropertyWrapper } from '../../../mol-model-props/common/wrapper';
+import { ParamDefinition as PD } from '../../../mol-util/param-definition';
+import { MmcifFormat } from '../../../mol-model-formats/structure/mmcif';
+import { CustomPropertyDescriptor } from '../../../mol-model/custom-property';
+import { CustomModelProperty } from '../../../mol-model-props/common/custom-model-property';
+
+type TypeId = number;
+type IdToCharge = Map<number, number>;
+type ChargesData = {
```
dsehnal, line 11:
```suggestion
interface ChargesData {
```

src/extensions/sb-ncbr/partial-charges/property.ts
```diff
@@ -0,0 +1,193 @@
...
+            ([typeId, method]) => [typeId, method] as [number, string]
+        );
+        PartialChargesPropertyParams = {
+            typeId: PD.Select<number>(1, options),
+        };
+        defaultPartialChargesPropertyParams = PD.clone(PartialChargesPropertyParams);
+        updatedParams = true;
+    }
+
+    return PartialChargesPropertyParams;
```
dsehnal, line 42: This will only work for 1 model. Remove the `updatedParams` and just return a copy of the `PartialChargesPropertyParams` with the `typeId` as select.

If you wanted to optimize it, you can use `model._staticPropertyData` to cache the parsed structure. See for example https://github.com/molstar/molstar/blob/master/src/extensions/pdbe/struct-ref-domain.ts#L18 how we use this.

src/extensions/sb-ncbr/partial-charges/property.ts
```diff
@@ -0,0 +1,193 @@
...
+}
+
+async function getData(model: Model): Promise<CustomProperty.Data<PartialCharges>> {
+    await Promise.resolve();
```
dsehnal, line 46: Why is this async and why are you awaiting the promise here?

src/extensions/sb-ncbr/partial-charges/property.ts
```diff
@@ -0,0 +1,193 @@
...
+    };
+}
+
+function getTypeIdToMethod(model: Model) {
+    const typeIdToMethod: ChargesData['typeIdToMethod'] = new Map();
+
+    const sourceData = model.sourceData as MmcifFormat;
+    const rowCount = sourceData.data.frame.categories.partial_atomic_charges_meta.rowCount;
+    const typeIds = sourceData.data.frame.categories.partial_atomic_charges_meta.getField('id')?.toIntArray();
+    const methods = sourceData.data.frame.categories.partial_atomic_charges_meta.getField('method')?.toStringArray();
```
dsehnal, line 79: You don't need to turn this into an array. You can access the values via `.int` or `.str` directly since you only convert this to a map immediately anyway.

src/extensions/sb-ncbr/partial-charges/property.ts
```diff
@@ -0,0 +1,193 @@
...
+            },
+        },
+    };
+}
+
+function getTypeIdToMethod(model: Model) {
+    const typeIdToMethod: ChargesData['typeIdToMethod'] = new Map();
+
+    const sourceData = model.sourceData as MmcifFormat;
+    const rowCount = sourceData.data.frame.categories.partial_atomic_charges_meta.rowCount;
```
dsehnal, line 77: `partial_atomic_charges_meta` is not a good name for the category. At the very least, it should be prefixed with `sb_ncbr_` namespace (`sb_ncbr_partial_atomic_charges_meta`) since it is specific to that use case. Same applies to the other category.

src/extensions/sb-ncbr/partial-charges/property.ts
```diff
@@ -0,0 +1,193 @@
...
+        for (let residueId = 1; residueId < offsets.length; ++residueId) {
+            let charge = 0;
+            for (let atomId = offsets[residueId - 1] + 1; atomId <= offsets[residueId]; ++atomId) {
+                charge += atomIdToCharge?.get(atomId) || 0;
+            }
+            for (let atomId = offsets[residueId - 1] + 1; atomId <= offsets[residueId]; ++atomId) {
+                if (!residueToCharge.has(typeId)) residueToCharge.set(typeId, new Map());
+                residueToCharge.get(typeId)?.set(atomId, Number(charge.toFixed(4)));
+            }
+        }
```
dsehnal, line 130: How did you arrive at this usage of the offsets? This is not how they work. Also, I assume your `partial_atomic_charges.atom_id` should map to `atom_site.label_atom_id` and in that case, you mapping doesn't always have to be correct. I assume `offsets[residueId - 1] + 1` was a hack to get the 1-based indexing you see in most mmCIF files.

I am writing this from the top of my head, but it should be something like this:

```ts
const { offsets, count } = model.atomicHierarchy.residueAtomSegments;
const atomId = StructureProperties.atom.label_atom_id;
const label_atom_id = model.atomicHierarchy.atoms.label_atom_id;
for (let rI = 0; rI < count; rI++) {
  for (let aI = offsets[rI], _aI = offsets[rI+1]; aI < _aI; aI++) {
    const atom_id = label_atom_id.value(aI);
     // ...
  }
}
```

src/extensions/sb-ncbr/partial-charges/property.ts
```diff
@@ -0,0 +1,193 @@
...
+    const residueToCharge: ChargesData['typeIdToResidueToCharge'] = new Map();
+
+    typeIdToAtomIdToCharge.forEach((atomIdToCharge, typeId: number) => {
+        for (let residueId = 1; residueId < offsets.length; ++residueId) {
+            let charge = 0;
+            for (let atomId = offsets[residueId - 1] + 1; atomId <= offsets[residueId]; ++atomId) {
+                charge += atomIdToCharge?.get(atomId) || 0;
+            }
+            for (let atomId = offsets[residueId - 1] + 1; atomId <= offsets[residueId]; ++atomId) {
+                if (!residueToCharge.has(typeId)) residueToCharge.set(typeId, new Map());
```
dsehnal, line 127: You don't need to do this inside a nested loop. Only need to do this once at the top of this for each.

src/extensions/sb-ncbr/partial-charges/property.ts
```diff
@@ -0,0 +1,193 @@
...
+
+function getTypeIdToResidueIdToCharge(model: Model, typeIdToAtomIdToCharge: ChargesData['typeIdToAtomIdToCharge']) {
+    const { offsets } = model.atomicHierarchy.residueAtomSegments;
+    const residueToCharge: ChargesData['typeIdToResidueToCharge'] = new Map();
+
+    typeIdToAtomIdToCharge.forEach((atomIdToCharge, typeId: number) => {
+        for (let residueId = 1; residueId < offsets.length; ++residueId) {
+            let charge = 0;
+            for (let atomId = offsets[residueId - 1] + 1; atomId <= offsets[residueId]; ++atomId) {
+                charge += atomIdToCharge?.get(atomId) || 0;
```
dsehnal, line 124: Why the `?.` here?

src/extensions/sb-ncbr/partial-charges/property.ts
```diff
@@ -0,0 +1,193 @@
...
+): ChargesData['maxAbsoluteAtomCharges'];
+function getMaxAbsoluteCharges(
+    typeIdToCharge: ChargesData['typeIdToResidueToCharge']
+): ChargesData['maxAbsoluteResidueCharges'] {
+    const maxAbsoluteCharges: Map<number, number> = new Map();
+
+    typeIdToCharge.forEach((idToCharge, typeId) => {
+        const charges = Array.from(idToCharge.values());
+        const min = Math.min(...charges);
+        const max = Math.max(...charges);
```
dsehnal, line 147: We have `arrayMinMax` function in `mol-util/array`

src/extensions/sb-ncbr/partial-charges/property.ts
```diff
@@ -0,0 +1,193 @@
...
+    const names = model.sourceData.data.frame.categoryNames;
+    return (
+        names.includes('atom_site') &&
+        names.includes('partial_atomic_charges') &&
+        names.includes('partial_atomic_charges_meta')
+    );
+}
+
+export function isApplicable(model?: Model): boolean {
+    return !!model && model.sourceData.kind === 'mmCIF' && hasPartialChargesCategories(model);
```
dsehnal, line 177: The `model.sourceData.kind === 'mmCIF'` is already part of the `hasPartialChargesCategories`.

src/extensions/sb-ncbr/partial-charges/property.ts
```diff
@@ -0,0 +1,193 @@
...
+    if (!MmcifFormat.is(model.sourceData)) return false;
+    const names = model.sourceData.data.frame.categoryNames;
+    return (
+        names.includes('atom_site') &&
+        names.includes('partial_atomic_charges') &&
+        names.includes('partial_atomic_charges_meta')
+    );
+}
+
+export function isApplicable(model?: Model): boolean {
```
dsehnal, line 176: Can you please give this a more specific name?

src/extensions/sb-ncbr/partial-charges/property.ts
```diff
@@ -0,0 +1,193 @@
...
+}
+
+function getTypeIdToAtomIdToCharge(model: Model): ChargesData['typeIdToAtomIdToCharge'] {
+    const atomIdToCharge: ChargesData['typeIdToAtomIdToCharge'] = new Map();
+
+    const sourceData = model.sourceData as MmcifFormat;
+    const rowCount = sourceData.data.frame.categories.partial_atomic_charges.rowCount;
+    const typeIds = sourceData.data.frame.categories.partial_atomic_charges.getField('type_id')?.toIntArray();
+    const atomIds = sourceData.data.frame.categories.partial_atomic_charges.getField('atom_id')?.toIntArray();
+    const charges = sourceData.data.frame.categories.partial_atomic_charges.getField('charge')?.toFloatArray();
```
dsehnal, line 101: No need to convert to the arrays either here.

src/extensions/sb-ncbr/partial-charges/behavior.ts
```diff
@@ -0,0 +1,38 @@
+import { LociLabelProvider } from '../../../mol-plugin-state/manager/loci-label';
```
dsehnal, line 1: Can you please add `readme.md` file here with some basic documentation about what this extension does and how it works?

dsehnal (commented):
Thanks for the contribution Dominik. It's gonna need some more work before we merge this.

src/extensions/sb-ncbr/partial-charges/color.ts
```diff
@@ -0,0 +1,135 @@
...
+): ColorTheme<PartialChargesThemeParams> {
+    const model = ctx.structure?.models[0];
+    if (!model) {
+        throw new Error('No model found');
+    }
+    const data = SbNcbrPartialChargesPropertyProvider.get(model).value?.data;
+    const typeId = SbNcbrPartialChargesPropertyProvider.getParams(model).typeId.defaultValue;
+
+    // forces coloring updates
+    props.typeId = typeId;
```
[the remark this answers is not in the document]
MergunFrimen: I couldn't figure out how to update the coloring when switching to a different charge set. I noticed that the coloring changes when a parameter is changed so that's why I used this.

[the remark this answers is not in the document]
arose: Color themes support a `contextHash` for that purpose, see https://github.com/molstar/molstar/blob/804117475b651da9dd4f4be86bc3a6d9738d2837/src/mol-theme/color/secondary-structure.ts#L114

src/extensions/sb-ncbr/partial-charges/property.ts
MergunFrimen: I'm getting the residue charge by summing all charges of atoms in the residue. The reason why I'm then writing the resulting residue charge to each atomId in the residue is because I don't know which atomId Molstar will use when coloring residues in the ColorTheme.

The `partial_atomic_charges.atom_id` item should map to `atom_site.id`, however I wasn't able to find it in model.atomicHierarchy.

dsehnal: Right, I made a mistake with the field name. You can still access the id column via the raw "atom_site" and use `sourceIndex` from the hierarchy to get the correct index into that category.

In general your current approach wont work because mol* also sorts the inputs (hence the existence of the `sourceIndex`).

src/extensions/sb-ncbr/partial-charges/labels.ts
MergunFrimen: yes

src/extensions/sb-ncbr/partial-charges/property.ts
```diff
@@ -0,0 +1,195 @@
...
+                charge += atomIdToCharge.get(atom_id) || 0;
+            }
+            for (let aI = offsets[rI], _aI = offsets[rI + 1]; aI < _aI; aI++) {
+                const atom_id = atomIds.int(atomSourceIndex.value(aI));
+                residueToCharge.get(typeId)?.set(atom_id, Number(charge.toFixed(4)));
+            }
+        }
+    });
+
+    return residueToCharge;
```
MergunFrimen, line 133: Updated the use of offsets. Hopefully this is better?

src/extensions/sb-ncbr/partial-charges/property.ts
```diff
@@ -0,0 +1,195 @@
...
+    const residueToCharge: ChargesData['typeIdToResidueToCharge'] = new Map();
+
+    if (!atomIds) return residueToCharge;
+
+    typeIdToAtomIdToCharge.forEach((atomIdToCharge, typeId: number) => {
+        if (!residueToCharge.has(typeId)) residueToCharge.set(typeId, new Map());
+        for (let rI = 0; rI < count; rI++) {
+            let charge = 0;
+            for (let aI = offsets[rI], _aI = offsets[rI + 1]; aI < _aI; aI++) {
+                const atom_id = atomIds.int(atomSourceIndex.value(aI));
```
dsehnal, line 123: Perhaps a better way to get the atom id is from `model.atomicConformation.atomId`. You won't need to do the `atomSourceIndex` mapping then. (Forgot about this one before).

src/extensions/sb-ncbr/partial-charges/property.ts
dsehnal: This does indeed look better. Have you tested it behaves as expected too?

src/extensions/sb-ncbr/partial-charges/color.ts
```diff
@@ -0,0 +1,142 @@
...
+        throw new Error('No model found');
+    }
+    const data = SbNcbrPartialChargesPropertyProvider.get(model).value?.data;
+    if (!data) {
+        throw new Error('No partial charges data found');
+    }
+
+    const { absolute, chargeType } = props;
+    const { typeIdToAtomIdToCharge, typeIdToResidueToCharge, maxAbsoluteAtomCharges, maxAbsoluteResidueCharges } = data;
+    const typeId = SbNcbrPartialChargesPropertyProvider.props(model).typeId;
```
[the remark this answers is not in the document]
MergunFrimen: I describe that in the `README.txt` file. The user has an option to select the `typeId` in Custom Model Properties settings of Model in the State Tree UI.

src/extensions/sb-ncbr/partial-charges/property.ts
MergunFrimen: Yes, I have tested it on multiple mmcif files and it appears to work as intended.

src/extensions/sb-ncbr/partial-charges/property.ts
```diff
@@ -0,0 +1,192 @@
...
+        const residueCharges = residueToCharge.get(typeId)!;
+        for (let rI = 0; rI < count; rI++) {
+            let charge = 0;
+            for (let aI = offsets[rI], _aI = offsets[rI + 1]; aI < _aI; aI++) {
+                const atom_id = atomIds.value(aI);
+                charge += atomIdToCharge.get(atom_id) || 0;
+            }
+            for (let aI = offsets[rI], _aI = offsets[rI + 1]; aI < _aI; aI++) {
+                const atom_id = atomIds.value(aI);
+                residueCharges.set(atom_id, Number(charge.toFixed(4)));
```
dsehnal, line 125: - Rounding number to 4 decimal places: `Math.round(v * 1e4) / 1e4`
- Tho I don't really understand why you do that here at all since you call the `.toFixed` in the label provider anyway.

src/extensions/sb-ncbr/partial-charges/property.ts
```diff
@@ -0,0 +1,192 @@
...
+                typeIdToMethod,
+                typeIdToAtomIdToCharge,
+                typeIdToResidueToCharge,
+                maxAbsoluteAtomCharges,
+                maxAbsoluteResidueCharges,
+                maxAbsoluteAtomChargeAll,
+            },
+        },
+    };
+}
```
dsehnal, line 63: I think the `PropertyWrapper` around `ChargesData` is redundant and you should be able to return just `CustomProperty.Data<ChargesData | undefined>` here and modify the rest of the code accordingly.

I know it's used in the other extension which I pointed you to, but no need to use that here as the `PropertyWrapper.createInfo();` isn't used anywhere relevant to this extension (I hope).

-----

Can you try something like this and modifying the usage of this code?

```suggestion
async function getData(model: Model): Promise<CustomProperty.Data<ChargesData | undefined>> {
    if (!SbNcbrPartialChargesPropertyProvider.isApplicable(model)) return undefined;

    const typeIdToMethod = getTypeIdToMethod(model);
    const typeIdToAtomIdToCharge = getTypeIdToAtomIdToCharge(model);
    const typeIdToResidueToCharge = getTypeIdToResidueIdToCharge(model, typeIdToAtomIdToCharge);
    const maxAbsoluteAtomCharges = getMaxAbsoluteCharges(typeIdToAtomIdToCharge);
    const maxAbsoluteResidueCharges = getMaxAbsoluteCharges(typeIdToResidueToCharge);
    const maxAbsoluteAtomChargeAll = getMaxAbsoluteAtomChargeAll(maxAbsoluteAtomCharges, maxAbsoluteResidueCharges);

    return {
        typeIdToMethod,
        typeIdToAtomIdToCharge,
        typeIdToResidueToCharge,
        maxAbsoluteAtomCharges,
        maxAbsoluteResidueCharges,
        maxAbsoluteAtomChargeAll,
    };
}
```

src/extensions/sb-ncbr/partial-charges/property.ts
```diff
@@ -0,0 +1,192 @@
...
+import { PropertyWrapper } from '../../../mol-model-props/common/wrapper';
+import { ParamDefinition as PD } from '../../../mol-util/param-definition';
+import { MmcifFormat } from '../../../mol-model-formats/structure/mmcif';
+import { CustomPropertyDescriptor } from '../../../mol-model/custom-property';
+import { CustomModelProperty } from '../../../mol-model-props/common/custom-model-property';
+import { arrayMinMax } from '../../../mol-util/array';
+
+type TypeId = number;
+type IdToCharge = Map<number, number>;
+interface ChargesData {
```
dsehnal, line 12:
```suggestion
export interface SBNcbrPartialChargeData {
```

since this is used outside this file.

dsehnal (commented):
A few more cosmetic changed and then should be ok to merge.

src/extensions/sb-ncbr/partial-charges/property.ts
MergunFrimen: Must be a remnant from testing. Will remove it.

src/extensions/sb-ncbr/partial-charges/property.ts
MergunFrimen: Yes, the `info` property is never used.

MergunFrimen:
Awesome! Thank you very much for the reviews. :)

src/extensions/sb-ncbr/partial-charges/color.ts
```diff
@@ -0,0 +1,150 @@
...
+        };
+    }
+
+    return {
+        factory: PartialChargesColorTheme,
+        granularity: 'group',
+        color,
+        props,
+        description: 'Color atoms and residues based on their partial charge.',
+        preferSmoothing: false,
```
arose, line 125: Hey @MergunFrimen, could we set `preferSmoothing: true`? Or was there a specific reason to set it to false? It looks much nicer set to true for surfaces by default in my opinion. (Set aside that it is debatable to even use partial charges to color surfaces directly)

PR #809, opened 2023-04-30
Adding principal axes spec

giagitom:
# Description

as discussed with @arose this adds a spec for principal axes evaluation. It presents a corner case where points are placed in the same cartesian plane. This special case is not handled and ```origin``` vector has NaN and Infinity values. 

## Actions


arose:
The principal axes can be just a plane or a vector (even a point). To still get a correct `origin` in `calculateBoxAxes` you need to check for each of the three axes vector if they are valid and ignore them for the origin calculation.

giagitom:
Should be done now.

arose (approved):
thanks, lgtm

PR #810, opened 2023-04-30
Geo export2

arose:
# Description
[support points & lines in glTF export](https://github.com/molstar/molstar/commit/03dba0bde7e3a282b60122a92713187d74b625cc)

## Actions

- [X] Added description of changes to the `[Unreleased]` section of `CHANGELOG.md`
- [X] Updated headers of modified files

### DIFF
--- a/src/extensions/geo-export/mesh-exporter.ts
+++ b/src/extensions/geo-export/mesh-exporter.ts
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2021 mol* contributors, licensed under MIT, See LICENSE file for more info.
+ * Copyright (c) 2021-2023 mol* contributors, licensed under MIT, See LICENSE file for more info.
  *
  * @author Sukolsak Sakshuwong <sukolsak@stanford.edu>
  * @author Alexander Rose <alexander.rose@weirdbyte.de>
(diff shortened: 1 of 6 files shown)


src/extensions/geo-export/mesh-exporter.ts
```diff
@@ -349,15 +365,38 @@ export abstract class MeshExporter<D extends RenderObjectExportData> implements
...
+        const aStart = values.aStart.ref.value;
+        const aEnd = values.aEnd.ref.value;
+        const aGroup = values.aGroup.ref.value;
+        const vertexCount = (values.uVertexCount.ref.value / 4) * 2;
+        const drawCount = values.drawCount.ref.value / (2 * 3);
+
+        const n = (vertexCount / 2);
+        const vertices = new Float32Array(n * 2 * 3);
+        for (let i = 0; i < n; ++i) {
+            vertices[i * 6] = aStart[i * 4 * 3];
```
dsehnal, line 381: Why the `4 *`?

dsehnal (approved):

src/extensions/geo-export/mesh-exporter.ts
arose: we use two triangles made with 4 vertices to draw wide (more than a pixel) lines because webgl implementers don't support wide lines using the lines primitive

dsehnal: Could this be achieved by a single zero area triangle as well? ...two vertices are the endpoints with the last point being in the middle and having a shader similar to how we do outlines?

PR #825, opened 2023-05-21
Immer, es6 & beyond

arose:
Update immer to v10 and change build target to es6.

Mainly means dropping support for IE11. Maybe we should call it 4.0 and include other breaking changes.

Edit: I suggest to change the target to ES2018 (widely available and doesn't need many polyfills). Custom build can still be done with ES6 (or even lower if you pin immer to v9) if need be.


src/mol-canvas3d/canvas3d.ts
```diff
@@ -315,6 +315,10 @@ namespace Canvas3D {
         let y = 0;
         let width = 128;
         let height = 128;
+
+        let forceNextRender = false;
+        let currentTime = 0;
```
dsehnal, line 320: This was required by the ES6 target saying the variable is not declared yet?

dsehnal (commented):
- I don't have a problem with going to ES6, but I think we should test all functionality before doing so.
- Also, since this is a breaking change, should we make this a 4.0 release, perhaps together with https://github.com/molstar/molstar/pull/790 when it's ready?

src/mol-canvas3d/canvas3d.ts
arose: yes, because with es6 you have let/const

PR #826, opened 2023-05-21
Graph & bonds calc tweaks & fixes

arose:
# Description

## Actions


src/mol-math/graph/int-adjacency-graph.ts
```diff
@@ -176,6 +176,13 @@ export namespace IntAdjacencyGraph {
             prop[this.curB] = value;
         }
 
+        assignDirectedProperty<T>(propA: { [i: number]: T }, valueA: T, propB: { [i: number]: T }, valueB: T) {
+            propA[this.curA] = valueA;
+            propA[this.curB] = valueB;
+            propB[this.curB] = valueA;
+            propB[this.curA] = valueB;
```
dsehnal, line 183: Would it be enough to actually assign the value only in the desired direction and use a placeholder (say -1) for the other?

arose: still want both directions but the props need to be swapped because they are basically vertex properties

Issue #831, opened 2023-05-22
Impostor cylinders issue


giagitom:
I noticed an issue on impostor cylinders, here is a state to reproduce it [test.molj](https://github.com/molstar/molstar/files/11533051/test.molj.txt)

![image](https://github.com/molstar/molstar/assets/20627977/665227e2-7132-45a7-9a4c-e0a635edb0e9)

arose:
That is quite an edge case. As far as I can tell it only happens when the distance between near and far plane is about the size of a cylinder. Could you just disallow bringing near and far plane that close together?

giagitom:
I first noticed this while zooming in using middle mouse click+drag default event. Maybe I could try to use a custom event for this. For longer cylinders the effect seems to be more visible though. Here an example with HEME [test-heme.molj](https://github.com/molstar/molstar/files/11544176/test.molj.txt)
![image](https://github.com/molstar/molstar/assets/20627977/a0fff0da-3069-4ebf-903a-849a783b20fe)

giagitom:
@arose I think I have found the cause in this line in vertex shader

https://github.com/molstar/molstar/blob/3bf4a8f8e692dd0f2159e9734e922d334aed1bdc/src/mol-gl/shader/cylinders.vert.ts#L75

Modifying it in the way below it seems to work fine. I don't know if there are drawbacks on doing so though. Do you think this would be an acceptable fix?

```mvPosition.z -= length(vEnd - vStart) + vSize; // avoid clipping```

![image](https://github.com/molstar/molstar/assets/20627977/077529d5-04a2-4571-867c-607ae99233a7)

arose:
Right, thanks, that should be `mvPosition.z -= length(vEnd - vStart) + 2.0 * vSize; // avoid clipping` since vSize is the cylinder radius.

PR #833, opened 2023-05-24
fix cylinder impostor clipping near camera

arose:
Fix for #831

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Note that since we don't clearly distinguish between a public and private interf
 ## [Unreleased]
 
 - Optimize inter unit bond compute
+- Fix cylinder impostor clipping near camera (#831)
 
 ## [v3.35.0] - 2023-05-14
 
--- a/src/mol-gl/shader/cylinders.vert.ts
+++ b/src/mol-gl/shader/cylinders.vert.ts
@@ -72,7 +72,7 @@ void main() {
     gl_Position = uProjection * mvPosition;
 
     if (gl_Position.z < -gl_Position.w) {
-        mvPosition.z -= 2.0 * (length(vEnd - vStart) + vSize); // avoid clipping
+        mvPosition.z -= length(vEnd - vStart) + 2.0 * vSize; // avoid clipping
         gl_Position.z = (uProjection * mvPosition).z;
     }
 


giagitom:
I think I have found also the same issue in sphere vertex shader.
Changing it to ```mvPosition.z -= vRadius; // avoid clipping``` avoids clipping issues where the sphere suddenly disappears.

Also it seems that some artifacts are still present on cylinders in any case.

https://github.com/molstar/molstar/assets/20627977/84fe31f2-1442-4960-b8f2-bb7e470351af

The best I could do was using this.. ```mvPosition.z -= length(vEnd - vStart) / 2.0 + vSize;``` that seems to avoid the artifacts on the video. But other kind of artifacts remain as you can see

image

arose:
I'd like to better understand what is actually happening here before committing to changes. This seems to cause problems only when both a) the camera is closer then minNear and b) the far plane is close the the near plane.

giagitom:
@arose I have found a simple way that sims to work in any case for both spheres and cylinders. 

```ts
if (gl_Position.z < -gl_Position.w) {
    gl_Position.z = -gl_Position.w;
}
```
or just this.. don't know which method is faster

```ts
gl_Position.z = max(gl_Position.z, -gl_Position.w); // avoid clipping
```

arose:
@giagitom setting `gl_Position.z = -gl_Position.w;` means that we never clip and only discard in the fragment shader. This is way too expensive when you zoom in and spheres/cylinders should get clipped in the vertex shader. You essentially evaluate ginormous spheres that are behind the camera. I did a quick test with a ribosome (4v5a), all spheres, zoomed all the way in, and it is 4 times slower.

giagitom:
@arose I have an updated version that tries to avoid this performance issue. It also checks for the back plane because I noticed it prevents elements to disappear from behind. But now elements should never disappear on small front to back plane distances. 
Performances on spheres are still a bit worse than before, but I noticed that just increasing Clipping Min distance by 1.0 leads to a tremendous improvement in performances on my system.. Maybe we could update the default clip min distance to 6.0?

Here are the updated versions.
For spheres:

```ts
    if (gl_Position.z < -gl_Position.w) { //avoid front clipping
        mvPosition.z -= vRadius;
        mvPosition = (uProjection * vec4(mvPosition.xyz, 1.0));
        if (mvPosition.z >= -mvPosition.w) {
            gl_Position.z = -gl_Position.w;
        }
    } else if (gl_Position.z > gl_Position.w){ //avoid back clipping
          mvPosition.z += vRadius;
          mvPosition = (uProjection * vec4(mvPosition.xyz, 1.0));
          if (mvPosition.z <= mvPosition.w) {
              gl_Position.z = gl_Position.w;
          }
      }
```

For cylinders:

```ts
    if (gl_Position.z < -gl_Position.w) { //avoid front clipping
        mvPosition.z -= abs(vEnd.z - vStart.z) + vSize * 3.5;
        mvPosition = uProjection * mvPosition;
        if (mvPosition.z >= -mvPosition.w) {
            gl_Position.z = -gl_Position.w;
        }
    } else if (gl_Position.z > gl_Position.w) { //avoid back clipping
        mvPosition.z += abs(vEnd.z - vStart.z) + vSize * 3.5;
        mvPosition = uProjection * mvPosition;
        if (mvPosition.z <= mvPosition.w) {
            gl_Position.z = gl_Position.w;
        }
    }
```

https://github.com/molstar/molstar/assets/20627977/bc2ca22b-758c-4ecc-bf52-1f82ce312056

giagitom:
@arose I was testing a new configuration for cylinders and found out that the current configuration seems to be the best. It avoids any kind of artifacts and keeps performances the same as before on all my tests.
```ts
if (gl_Position.z < -gl_Position.w) { //avoid front clipping
        mvPosition.z -= 2.0 * (distance(vEnd , vStart) + vSize);
        mvPosition = uProjection * mvPosition;
        if (mvPosition.z >= -mvPosition.w) {
            gl_Position.z = -gl_Position.w;
        }
    } else if (gl_Position.z > gl_Position.w) { //avoid back clipping
        mvPosition.z += 2.0 * (distance(vEnd , vStart) + vSize);
        mvPosition = uProjection * mvPosition;
        if (mvPosition.z <= mvPosition.w) {
            gl_Position.z = gl_Position.w;
        }
    }
```

PR #836, opened 2023-05-27
add snapshot support for structure selections

arose:
# Description

Add snapshot support for structure selections. Off by default.

## Actions

### DIFF
--- a/src/mol-plugin-state/manager/structure/selection.ts
+++ b/src/mol-plugin-state/manager/structure/selection.ts
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2019-2021 mol* contributors, licensed under MIT, See LICENSE file for more info.
+ * Copyright (c) 2019-2023 mol* contributors, licensed under MIT, See LICENSE file for more info.
  *
  * @author David Sehnal <david.sehnal@gmail.com>
  * @author Alexander Rose <alexander.rose@weirdbyte.de>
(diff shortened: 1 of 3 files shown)


src/mol-plugin-state/manager/structure/selection.ts
```diff
@@ -484,6 +491,31 @@ export class StructureSelectionManager extends StatefulPluginComponent<Structure
         }
     }
 
+    getSnapshot(): StructureSelectionSnapshot {
+        const entries: StructureSelectionSnapshot['entries'] = [];
+
+        this.entries.forEach((entry, ref) => {
+            entries.push({
+                ref,
+                bundle: StructureElement.Bundle.fromLoci(entry.selection)
```
dsehnal, line 500: Could it happen that the bundle is using a typed array? In that case should convert it to native array.

src/mol-plugin/state.ts
```diff
@@ -89,6 +92,9 @@ class PluginState extends PluginComponent {
         if (snapshot.structureFocus) {
             this.plugin.managers.structure.focus.setSnapshot(snapshot.structureFocus);
         }
+        if (snapshot.structureSelection) {
+            this.plugin.managers.structure.selection.setSnapshot(snapshot.structureSelection);
```
dsehnal, line 96: Will this work on fresh snapshots where the visual isn't created yet?

src/mol-plugin-state/manager/structure/selection.ts
arose: We don't prevent it with type checking but the `Bundle.from*` methods don't use typed arrays. Got an idea to enforce that easily? We already include `Bundle`s in the state in `StructureSelectionFromBundle`, `OverpaintStructureRepresentation3DFromBundle` and other transforms.

src/mol-plugin/state.ts
arose: It needs the structures the bundles are refering to in the state tree which is the case here with the data snapshot applied before. Visuals can be created later and get the selection mark applied like normally. Also it worked for me using it :)

src/mol-plugin-state/manager/structure/selection.ts
dsehnal: Only to traverse the tree. But yeah, been using it in the state tree already, so should be fine.

dsehnal (approved):

PR #838, opened 2023-06-03
Cartoon improvements

arose:
# Description

Support more flexible coloring of cartoon visuals.

- Add `nucleicProfile` parameter to cartoon representation
- Add `cartoon` theme with separate colorings for for mainchain and sidechain visuals

image

## Actions

### DIFF
--- a/src/mol-theme/color/cartoon.ts
+++ b/src/mol-theme/color/cartoon.ts
@@ -48,23 +48,34 @@ export function getCartoonColorThemeParams(ctx: ThemeDataContext) {
     return params;
 }
 
-export function CartoonColorTheme(ctx: ThemeDataContext, props: PD.Values<CartoonColorThemeParams>): ColorTheme<CartoonColorThemeParams> {
-    const mainchain =
-        props.mainchain.name === 'uniform' ? UniformColorTheme(ctx, props.mainchain.params) :
-            props.mainchain.name === 'chain-id' ? ChainIdColorTheme(ctx, props.mainchain.params) :
-                props.mainchain.name === 'entity-id' ? EntityIdColorTheme(ctx, props.mainchain.params) :
-                    props.mainchain.name === 'entity-source' ? EntitySourceColorTheme(ctx, props.mainchain.params) :
-                        props.mainchain.name === 'molecule-type' ? MoleculeTypeColorTheme(ctx, props.mainchain.params) :
-                            props.mainchain.name === 'model-index' ? ModelIndexColorTheme(ctx, props.mainchain.params) :
-                                props.mainchain.name === 'structure-index' ? StructureIndexColorTheme(ctx, props.mainchain.params) :
-                                    props.mainchain.name === 'secondary-structure' ? SecondaryStructureColorTheme(ctx, props.mainchain.params) :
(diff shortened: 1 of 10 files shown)


src/mol-theme/color/cartoon.ts
```diff
@@ -0,0 +1,100 @@
...
+    }),
+};
+export type CartoonColorThemeParams = typeof CartoonColorThemeParams
+export function getCartoonColorThemeParams(ctx: ThemeDataContext) {
+    const params = PD.clone(CartoonColorThemeParams);
+    return params;
+}
+
+export function CartoonColorTheme(ctx: ThemeDataContext, props: PD.Values<CartoonColorThemeParams>): ColorTheme<CartoonColorThemeParams> {
+    const mainchain =
```
dsehnal, line 52: Use a switch statement or a dict map to a lambda provider?

src/mol-theme/color/cartoon.ts
```diff
@@ -0,0 +1,100 @@
...
+                                    props.mainchain.name === 'secondary-structure' ? SecondaryStructureColorTheme(ctx, props.mainchain.params) :
+                                        assertUnreachable(props.mainchain);
+
+    const sidechain =
+        props.sidechain.name === 'uniform' ? UniformColorTheme(ctx, props.sidechain.params) :
+            props.sidechain.name === 'residue-name' ? ResidueNameColorTheme(ctx, props.sidechain.params) :
+                props.sidechain.name === 'element-symbol' ? ElementSymbolColorTheme(ctx, props.sidechain.params) :
+                    assertUnreachable(props.sidechain);
+
+    function color(location: Location, isSecondary: boolean): Color {
```
dsehnal, line 69: Do all the color callbacks have two params now everywhere? 

I think passing the wrong number of args (eg different from the number in the declaration) can cause a deoptimization.

dsehnal (approved):

src/mol-theme/color/cartoon.ts
arose: we don't like indents :)

src/mol-theme/color/cartoon.ts
arose: The color function has always been called with two parameters. In most themes the color functions only use (defines?) one argument.

src/mol-theme/color/cartoon.ts
dsehnal:
```ts
const ThemeMap = {
  uniform: UniformColorTheme,
  'chain-id': ChainIdColorTheme,
  // ...
};

const theme = ThemeMap[props.mainchain.name as keyof typeof ThemeMap]?.(ctx, props.mainchain.params);
if (!theme) throw new Error(`Unknown mainchain theme: ${props.mainchain.name}`);
```

src/mol-theme/color/cartoon.ts
dsehnal: Somehow, I've always missed that :)

src/mol-theme/color/cartoon.ts
arose: went with the switch statement suggestion

PR #858, opened 2023-07-08
update getStateSnapshot behavior

dsehnal:
# Description

## Actions


arose (commented):
This is certainly better but does not fully solve the issue. Maybe we need to distinguish between an "auto save" snapshot that gets added automatically on save and snapshots that the user explicitly adds?

dsehnal:
> This is certainly better but does not fully solve the issue. Maybe we need to distinguish between an "auto save" snapshot that gets added automatically on save and snapshots that the user explicitly adds?

Yeah, that sounds good. Will do it.

arose (approved):
tweak changlog message?

PR #864, opened 2023-07-09
reduce spheres geo memory usage

arose:
# Description

- derive mapping from VertexID
- pull position and group from texture

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ Note that since we don't clearly distinguish between a public and private interf
 
 - Fix display issue with SIFTS mapping
 - Properly switch-off fog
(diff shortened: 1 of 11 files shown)


dsehnal (commented):
Cool!

Should there be a check somewhere that tests if the number of spheres fits into the texture (based on the max texture size) and throw and error otherwise?

arose:
Good point. Almost all desktops support 8k, most 16k. On mobile (android) the min is 4k, which is 16M atoms. See https://web3dsurvey.com/webgl2/parameters/MAX_TEXTURE_SIZE. I don't think it is a problem in practice, but if it turns out is is we can add a check and a fallback.

PR #878, opened 2023-07-24
Fix logic for trajectoryFromCCD

### LINKED ISSUE #877: Cannot load 1fqy
MolStar cannot load PDB structure 1fqy, showing "Error: No models found."

I tried it with older versions and it seems that it broke in this PR:
https://github.com/molstar/molstar/pull/731

(I suspect the problem is somewhere in the `trajectoryFromCCD` function because when I remove `_chem_comp_atom` category from the mmCIF, it loads it without error.)
https://github.com/JonStargaryen/molstar-proto/blob/2204e4e0d020bd59c86a26dd9b91b5f427f2661b/src/mol-plugin-state/transforms/model.ts#L306

JonStargaryen:
# Description
Proposal to fix #877 by ensuring that the CCD route only triggers if `chem_comp_atom` is defined and, simultaneously, no conventional coordinates are present.

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ Note that since we don't clearly distinguish between a public and private interf
 
 ## [Unreleased]
 
+- Don't rely solely on `chem_comp_atom` when detecting CCD files (#877)
+
 ## [v3.38.1] - 2023-07-22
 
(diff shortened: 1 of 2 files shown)


arose:
lgtm, thanks

PR #929, opened 2023-09-28
fix bounding sphere calculations for "element-like" visuals

dsehnal:
# Description

Previous version could result in the bounding sphere not updating even for a large change in conformation.

## Actions


arose:
lgtm, thanks

PR #932, opened 2023-09-30
Sharpening

arose:
# Description

- Add sharpening postprocessing option
- Take pixel-ratio into account for outline scale
- Gracefully handle missing HTMLImageElement
- Fix pixel-ratio changes not applied to all render passes

## Actions


src/mol-canvas3d/passes/postprocessing.ts
```diff
@@ -953,17 +958,23 @@ export class AntialiasingPass {
...
     private readonly smaa: SmaaPass;
+    private readonly cas: CasPass;
 
     constructor(webgl: WebGLContext, private drawPass: DrawPass) {
         const { colorTarget } = drawPass;
         const width = colorTarget.getWidth();
         const height = colorTarget.getHeight();
 
         this.target = webgl.createRenderTarget(width, height, false);
+        this.internalTarget = webgl.createRenderTarget(width, height, false);
```
dsehnal, line 973: Should the render targets be created lazily only if this is actually used?

dsehnal (approved):
Tested locally and code LGTM.

src/mol-canvas3d/passes/postprocessing.ts
arose: yeah, can safe some resources, there are quite a few places and not only render targets

PR #936, opened 2023-10-02
Sequence with secondary structure labels above

pechersky:
# Description

Add a new `Sequence` subclass: `SecondaryStructureSequence` to show DSSP values above the sequence.

Required modifying the SCSS for the existing spans.
Modified how spans are selected for marker updates.

## Actions

### DIFF
--- a/src/mol-plugin-ui/sequence.tsx
+++ b/src/mol-plugin-ui/sequence.tsx
@@ -419,7 +419,7 @@ function NonEmptySequenceWrapper({ children }: { children: React.ReactNode }) {
         {children}
     </div>;
 }
-export class SeconarySequenceView extends SequenceView {
+export class SecondaryStructureSequenceView extends SequenceView {
 
     getSequenceChild(key: number, sequenceWrapper: SequenceWrapper.Any) {
         return <SecondaryStructureSequence key={key} sequenceWrapper={sequenceWrapper} />;
--- a/src/mol-plugin-ui/sequence.tsx
+++ b/src/mol-plugin-ui/sequence.tsx
@@ -13,7 +13,7 @@ import { Structure, StructureElement, StructureProperties as SP, Unit } from '..
 import { SequenceWrapper } from './sequence/wrapper';
 import { PolymerSequenceWrapper } from './sequence/polymer';
 import { MarkerAction } from '../mol-util/marker-action';
-import { BoolControl, ParamOnChange, PureSelectControl } from './controls/parameters';
+import { PureSelectControl } from './controls/parameters';
 import { ParamDefinition as PD } from '../mol-util/param-definition';
 import { HeteroSequenceWrapper } from './sequence/hetero';
 import { State, StateSelection } from '../mol-state';
@@ -208,7 +208,8 @@ export function getStructureOptions(state: State) {
 
 export type SequenceViewMode = 'single' | 'polymers' | 'all'
 const SequenceViewModeParam = PD.Select<SequenceViewMode>('single', [['single', 'Chain'], ['polymers', 'Polymers'], ['all', 'Everything']]);
-const SecondaryStructureViewModeParam = PD.Boolean(false, { label: 'Show Secondary Structure' });
+type SecondaryStructureViewMode = 'off' | 'on'
+const SecondaryStructureViewModeSelectParam = PD.Select<SecondaryStructureViewMode>('off', [['off', 'Off'], ['on', 'On']], { label: 'Show Secondary Structure' });
 
 type SequenceViewState = {
     structureOptions: { options: [string, string][], all: Structure[] },
@@ -324,7 +325,6 @@ export class SequenceView extends PluginUIComponent<{ defaultMode?: SequenceView
             chain: PD.Select(chainOptions[0][0], chainOptions, { shortLabel: true, twoColumns: true, label: 'Chain' }),
             operator: PD.Select(operatorOptions[0][0], operatorOptions, { shortLabel: true, twoColumns: true }),
             mode: SequenceViewModeParam,
-            secondaryStructure: PD.Boolean(false)
         };
     }
 
@@ -335,7 +335,6 @@ export class SequenceView extends PluginUIComponent<{ defaultMode?: SequenceView
             chain: this.state.chainGroupId,
             operator: this.state.operatorKey,
             mode: this.state.mode,
-            secondaryStructure: this.state.secondaryStructure,
         };
     }
 
@@ -401,7 +400,7 @@ export class SequenceView extends PluginUIComponent<{ defaultMode?: SequenceView
                 {params.operator.options.length > 1 && <>
                     <PureSelectControl title={`[Instance] ${PD.optionLabel(params.operator, values.operator)}`} param={params.operator} name='operator' value={values.operator} onChange={this.setParamProps} />
                 </>}
-                {<BoolControl name={`[Secondary Structure]`} param={SecondaryStructureViewModeParam} value={values.secondaryStructure} onChange={({ value }) => { this.state.secondaryStructure = value; }}/>}
+                <PureSelectControl title={`[Secondary Structure]`} name='secondary-structure' param={SecondaryStructureViewModeSelectParam} value={this.state.secondaryStructure ? 'on' : 'off'} onChange={({ value }) => this.setState({ secondaryStructure: value === 'on' }) }/>
             </div>
 
             <NonEmptySequenceWrapper>
--- a/src/mol-plugin-ui/sequence.tsx
+++ b/src/mol-plugin-ui/sequence.tsx
@@ -13,7 +13,7 @@ import { Structure, StructureElement, StructureProperties as SP, Unit } from '..
 import { SequenceWrapper } from './sequence/wrapper';
 import { PolymerSequenceWrapper } from './sequence/polymer';
 import { MarkerAction } from '../mol-util/marker-action';
-import { PureSelectControl } from './controls/parameters';
+import { BoolControl, ParamOnChange, PureSelectControl } from './controls/parameters';
 import { ParamDefinition as PD } from '../mol-util/param-definition';
 import { HeteroSequenceWrapper } from './sequence/hetero';
 import { State, StateSelection } from '../mol-state';
@@ -208,6 +208,7 @@ export function getStructureOptions(state: State) {
 
 export type SequenceViewMode = 'single' | 'polymers' | 'all'
 const SequenceViewModeParam = PD.Select<SequenceViewMode>('single', [['single', 'Chain'], ['polymers', 'Polymers'], ['all', 'Everything']]);
+const SecondaryStructureViewModeParam = PD.Boolean(false, { label: 'Show Secondary Structure' });
 
 type SequenceViewState = {
     structureOptions: { options: [string, string][], all: Structure[] },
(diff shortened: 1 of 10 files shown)


pechersky:
I would love guidance if this is how you'd like contributions, and if there are better patterns in the code you'd prefer.

src/mol-plugin-ui/sequence/secondary-structure-sequence.tsx
```diff
@@ -0,0 +1,100 @@
...
+        const suffix = this.props.hideSecondaryStructure ? '' : ' msp-sequence-number-above';
+        return super.getSequenceNumberClass(seqIdx, seqNum, label) + suffix;
+    }
+
+    protected getSequenceSecondaryStructureSpan(seqIdx: number) {
+        const loci = this.props.sequenceWrapper.getLoci(seqIdx);
+        const location = StructureElement.Loci.getFirstLocation(loci, this.location);
+        if (!location) return;
+        const structure: Structure = this.props.sequenceWrapper.data.structure;
+        const secondaryStructure = ModelSecondaryStructure.Provider.get(structure.model);
```
dsehnal, line 46: This could be called just once outside the loop for better performance.

src/mol-plugin-ui/sequence/secondary-structure-sequence.tsx
```diff
@@ -0,0 +1,100 @@
...
+        if (!secondaryStructure) return;
+        const { kind } =
+        secondaryStructure.elements[
+            secondaryStructure.key[
+                secondaryStructure.getIndex(
+                    StructureProperties.residue.key(location)
+                )
+            ]
+        ];
```
dsehnal, line 55:
```suggestion
        const { kind } = secondaryStructure.elements[
            secondaryStructure.key[
                secondaryStructure.getIndex(
                    StructureProperties.residue.key(location)
                )
            ]
        ];
```

dsehnal (commented):
Thanks for the contribution. It looks good overall. Before we merge this, a few quetions:

- What would be the intended use case of this component?
- Do you plan to extend the `SequenceView` to also support using `SecondaryStructureSequence`? 
- Maybe add a [`PluginUISpec.components`](https://github.com/molstar/molstar/blob/fc44e66b26d798d09fa681643e65e3d73241838e/src/mol-plugin-ui/spec.ts#L19) optional property that allows using the new component?
  - Alternative option would be to define a new [PluginConfig](https://github.com/molstar/molstar/blob/fc44e66b26d798d09fa681643e65e3d73241838e/src/mol-plugin/config.ts#L25) item.

pechersky:
I implemented the code style changes.

We have an internal modification of the sequence view that is able to display pLDDT values on top of the primary sequence as background color. Having the secondary structure as an annotation helps with identifying interesting regions of the protein without having to cross-reference with the 3D view -- thus being able to quickly select and zoom to regions of interest.

Let me know if I extended `SequenceView` and the plugin config definition in a way that makes sense, please.

src/mol-plugin/layout.ts
```diff
@@ -23,6 +23,7 @@ export type PluginLayoutControlsDisplay = 'outside' | 'portrait' | 'landscape' |
 export const PluginLayoutStateParams = {
     isExpanded: PD.Boolean(false),
     showControls: PD.Boolean(true),
+    showSecondaryStructureSequence: PD.Boolean(false),
```
dsehnal, line 26: Please move this to `mol-plugin-ui`'s  `PluginUISpec.components.sequenceViewer` as a `view?: React.ComponentClass`.

src/mol-plugin-ui/viewport.tsx
```diff
@@ -126,6 +130,7 @@ export class ViewportControls extends PluginUIComponent<ViewportControlsProps, V
                 <div>
                     <div className='msp-semi-transparent-background' />
                     {this.plugin.config.get(PluginConfig.Viewport.ShowControls) && this.icon(BuildOutlinedSvg, this.toggleControls, 'Toggle Controls Panel', this.plugin.layout.state.showControls)}
+                    {this.plugin.config.get(PluginConfig.Viewport.ShowSecondaryStructureSequence) && this.icon(BuildOutlinedSvg, this.toggleSSSequence, 'Toggle Secondary Structure Sequence', this.plugin.layout.state.showSecondaryStructureSequence)}
```
dsehnal, line 133: How about moving the toggle here? I think it makes more sense to show it directly in the control.

image

src/mol-plugin/config.ts
```diff
@@ -55,6 +55,7 @@ export const PluginConfig = {
     Viewport: {
         ShowExpand: item('viewer.show-expand-button', true),
         ShowControls: item('viewer.show-controls-button', true),
+        ShowSecondaryStructureSequence: item('viere.show-ss-sequence-button', true),
```
dsehnal, line 58: - there is a typo: `viere`
- Also, with moving the toggle to a different place, I don't think we need this config anymore?

dsehnal (commented):
The code itself is fine. I am just not sure the best place to put the toggle is the icon in the viewport controls. 

Can you please consider moving the toggle directly to the sequence viewer component?

pechersky:
I tried to move the config/toggle into the component, not sure if I'm using the parameter components correctly.

src/mol-plugin-ui/sequence.tsx
```diff
@@ -389,13 +401,14 @@ export class SequenceView extends PluginUIComponent<{ defaultMode?: SequenceView
                 {params.operator.options.length > 1 && <>
                     <PureSelectControl title={`[Instance] ${PD.optionLabel(params.operator, values.operator)}`} param={params.operator} name='operator' value={values.operator} onChange={this.setParamProps} />
                 </>}
+                {<BoolControl name={`[Secondary Structure]`} param={SecondaryStructureViewModeParam} value={values.secondaryStructure} onChange={({ value }) => { this.state.secondaryStructure = value; }}/>}
```
dsehnal, line 404: Need to call `setState` instead of just re-assigning it. The original assigment is just to provide defaults.

```suggestion
                {<BoolControl name={`[Secondary Structure]`} param={SecondaryStructureViewModeParam} value={values.secondaryStructure} onChange={({ value }) => this.setState({ secondaryStructure: value }) }/>}
```

dsehnal (commented):
Good progress, almost there :)

> I tried to move the config/toggle into the component, not sure if I'm using the parameter components correctly.

Can you please share a screenshot with what it looks like?

src/mol-plugin-ui/sequence.tsx
```diff
@@ -313,7 +323,8 @@ export class SequenceView extends PluginUIComponent<{ defaultMode?: SequenceView
             entity: PD.Select(entityOptions[0][0], entityOptions, { shortLabel: true }),
             chain: PD.Select(chainOptions[0][0], chainOptions, { shortLabel: true, twoColumns: true, label: 'Chain' }),
             operator: PD.Select(operatorOptions[0][0], operatorOptions, { shortLabel: true, twoColumns: true }),
-            mode: SequenceViewModeParam
+            mode: SequenceViewModeParam,
+            secondaryStructure: PD.Boolean(false)
```
dsehnal, line 327: With how you use it, no need to include this here or the "values".
```suggestion
```

pechersky:
Thanks, will incorporate the comments. Screenshot -- can't get the BoolControl to appear! Will debug.

pechersky:
Aha, it was on the next line...
image

Is there an inline small boolean control?

pechersky:
> Can you please consider moving the toggle directly to the sequence viewer component?

Ah, I remember now why I had this separate -- the Sequence component needs to change as well, since it handles how the `updateMarker` count works, and how the `render` works, and has the right css classes.

Do you have a suggestion on how to incorporate the changes in this file back into the main `Sequence` component?
`src/mol-plugin-ui/sequence/secondary-structure-sequence.tsx`

pechersky:
Here's the screenshot on 1tqn. What's not clear to me is why only the "none" secondary structure kinds are found, and why the helixes aren't identified. For my internal pdbs, the model secondary structure is inferred correctly.
image

dsehnal:
- I think the on/off could also use a label in front of it. 
- Or perhaps you can also use `MappedStatic` parameter type where you name the options an use the `cycle: true`. [example](https://github.com/molstar/molstar/blob/492494033f1b117cc2da02f577b747fc1ba1cc03/src/mol-canvas3d/helper/handle-helper.ts#L45) for better UX.

pechersky:
I'm not sure how to use the MappedStatic parameter type in a PureSelectControl.

src/mol-plugin-ui/sequence.tsx
```diff
@@ -207,6 +208,8 @@ export function getStructureOptions(state: State) {
 
 export type SequenceViewMode = 'single' | 'polymers' | 'all'
 const SequenceViewModeParam = PD.Select<SequenceViewMode>('single', [['single', 'Chain'], ['polymers', 'Polymers'], ['all', 'Everything']]);
+type SecondaryStructureViewMode = 'off' | 'on'
+const SecondaryStructureViewModeSelectParam = PD.Select<SecondaryStructureViewMode>('off', [['off', 'Hidden Secondary Structure'], ['on', 'Show Secondary Structure']], { label: 'Show Secondary Structure above sequence' });
```
dsehnal, line 212:
```suggestion
const SecondaryStructureViewModeSelectParam = PD.Select<SecondaryStructureViewMode>('off', [['off', 'No Annotation'], ['on', 'Secondary Structure Annotation']], { label: 'Select residue annotation type' });
```

dsehnal (commented):
If you are ok with the current change in the label, I think we can just merge it and possibly do any tweaks separately if needed.

pechersky:
Update below. I switched to a single span and positioning it absolutely -- this is future-proofing for another PR I might make, which allows for setting the sequence background to match the color of residue in the protein. If I keep it the way it is, the secondary structure spans break up the background-color, introducing vertical breaks that aren't pretty.
image

arose:
Hi @pechersky, thank you for this contribution. I need to look at this closer but I have some comments for now

- the secondary structure labels get out of sync with the sequence
- the secondary structure provider can changed (from file/model to DSSP with different parameters), so we need to trigger a redraw, probably handle that in the `updated` event (https://github.com/molstar/molstar/blob/745746f2431314ac8c61930a6641469702e88f6d/src/mol-plugin-ui/sequence.tsx#L227-L231)
- given the extra vertical space used by the secondary structure labels I feel we may need to make the top panel resizable (should be separate PR)

pechersky:
@arose thanks for the comments!
> the secondary structure labels get out of sync with the sequence
the secondary structure provider can changed (from file/model to DSSP with different parameters), so we need to trigger a redraw, probably handle that in the updated event (

Can you please explain more about how the subscription and redraws currently happen? The sequence (on master) gets updated if the model is updated -- why does that not happen here any more? I would think the `sequenceWrapper` is able to listen to the state because of
```
            wrapper: getSequenceWrapper(this.state, this.plugin.managers.structure.selection),
```

Regarding subscribing to a change in the secondary structure provider -- wouldn't that be caught by a change in the Structure (because that would change the `structure.model`)? I don't know much about which refs/objects can be updated, and how much state to put into this `SequenceView` -- should it hold the ModelSecondaryStructure.Provider as state to listen for updates?

Thanks for bearing with me on all these questions.

arose:
@pechersky How about first doing a PR that has only the support for custom sequence view, i.e., what is in plgin.tsx and spec.ts. That way you can build you own viewer until we have time to look more into this PR.

rdirisio:
> @pechersky How about first doing a PR that has only the support for custom sequence view, i.e., what is in plgin.tsx and spec.ts. That way you can build you own viewer until we have time to look more into this PR.

One of our goals with these code changes is to set background colors for the sequence viewer based on some model property. 

Ideally, we could pass in a custom `Sequence` to the built-in `SequenceView`, since right now it just renders the default `Sequence` ([here](https://github.com/molstar/molstar/blob/6edbae80db340134341631f669eec86543a0f1a8/src/mol-plugin-ui/sequence.tsx#L398)).

What I suggest is adding a `sequence` param to the `{components: {sequenceViewer: ...}` that was just added, or making it a separate param like `components: {sequence: ...}}` because then we could use a `CustomSequence` component that renders colors however a user would like, and one wouldn't have to copy SequenceView wholesale to accomplish that.

arose:
Closing for now since one can now provide a custom sequence panel

PR #966, opened 2023-10-29
add clipPrimitive option to spheres geometry

arose:
- Add `clipPrimitive` option to spheres geometry, clipping whole spheres instead of cutting them

1TQN (28)
1TQN (29)


arose:
ping @dsehnal does it look OK to you?

dsehnal (approved):
Pong

PR #968, opened 2023-10-31
Add DragAndDropManager

dsehnal:
# Description

- [x] Adds the ability to include custom drag and drop handlers.

## Actions


src/mol-plugin-state/helpers/drag-and-drop.ts
```diff
@@ -0,0 +1,39 @@
...
+            format: { name: 'auto', params: {} },
+            visuals: true
+        }));
+    }
+}
+
+export async function handleDragAndDrop(plugin: PluginContext, files: File[]) {
+    const handlers = Array.from(plugin.customDragAndDropHandlers.values());
+
+    for (const handler of handlers) {
```
midlik, line 33: I would iterate in reverse order. So that newer handlers can simply override older handlers without having to remove them (use case: there is already a handler handling .pdb and .cif, I want to change behavior for .cif while keeping .pdb as is)

midlik (changes requested):
Looks good. I successfully tested it on a toy example (load a list of PDB IDs from a file).
The only suggestion I have is the order of applying handlers.

Also, it might get quite unpredictable when users drop more files with various extensions. But I don't think there's a simple way around that.

src/mol-plugin-state/helpers/drag-and-drop.ts
dsehnal: Well, that's not so simple since the registration order isn't fully guaranteed. I will think about it a bit more.

dsehnal: Refactored the code to a manager and applied your suggestion.

midlik: Btw, I thought Map keeps the insertion order (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map).

dsehnal: Apparently it does, but doesn't matter now.

arose (approved):
lgtm

PR #977, opened 2023-11-12
Snapshot improvements

dsehnal:
# Description

- Add `key` property
- Ability to existing snapshot name, key, and description
- Separate UI control showing description of the current snapshot

Will add option to trigger snapshots from descriptions & labels in a followup PR.

## Actions


arose:
thanks, will look at it in a bit

arose (commented):
Looks generally good, thank you.

I am wondering about the `key` property. Do we really need it? It can allow more freedom  when implementing triggering a snapshot by a key but it also allows having the same key for multiple snapshots which can be confusing.

Probably should not allow html in descriptions, just XSS or so waiting to happen.

dsehnal:
- I think the key is needed to allow the "on-click" functionality in a somewhat robust/editable manner. Will add uniqueness validation.
- You are right about the XSS... I did think about and just went with it.. but since you mention it as well: How about supporting markdown with special hooks for activating snapshots from the descriptions?

dsehnal:
https://github.com/molstar/molstar/assets/21145776/82e60bef-587d-4d4f-a6f4-21fa8dc7d911

arose (approved):
awesome

PR #981, opened 2023-11-15
Allow sequence (top control) resizing

pechersky:
Spun out from comment
https://github.com/molstar/molstar/pull/936#issuecomment-1784479431

# Description

## Actions


arose:
Thank you @pechersky. I think resizing should also change the viewport, otherwise the sequence panel just overlaps or reveals empty space. Would be nice to allow resizing along the entire pan not just in the right corner.

arose:
Closing for now since one can now provide a custom sequence panel

PR #986, opened 2023-11-20
reduce memory usage of SymmetryOperator.ArrayMapping

arose:
Example heap profile with ~500k units. Removing the closures and individual objects in favor of classes helps a lot for that number of units. `system / Context`, `(closure)`, `(number)`, and others are much smaller.

Unoptimized
memunopt

Optimized
memopt


src/mol-math/geometry/symmetry-operator.ts
```diff
@@ -167,127 +168,111 @@ namespace SymmetryOperator {
...
-}
+        x(i: T): number {
+            const m = this.operator.matrix;
+            const { x: xs, y: ys, z: zs } = this.coordinates;
+            const xx = m[0], yy = m[4], zz = m[8], tx = m[12];
 
-export { SymmetryOperator };
+            const x = xs[i], y = ys[i], z = zs[i], w = (m[3] * x + m[7] * y + m[11] * z + m[15]) || 1.0;
+            return (xx * x + yy * y + zz * z + tx) / w;
+        }
```
dsehnal, line 204: Have you tested the performance difference as well? The object deconstruction here and `this.` accesses were not here before.

This one I am not sure about, but if we go to ES6 and the deconstruction (`{ x: xs, y: ys, z: zs }`) isn't transpiled to assignment, that could be an issue too (actually, not sure that is happening now either and their isn't just some helper method used to unpack the object).

arose: did some perf tests, see other comment

arose:
Closures are faster (I think after the JIT kicks in) for mappers with 100000 positions. I'd say the memory savings are still worth it.

I haven't tested the effect of having millions of mappers.

```
10000 class testXYZ x 48,383 ops/sec ±0.39% (94 runs sampled)
10000 class testPos x 56,421 ops/sec ±0.50% (98 runs sampled)
10000 class testInvPos x 59,019 ops/sec ±0.15% (100 runs sampled)
10000 closures testXYZ x 7,788 ops/sec ±6.39% (91 runs sampled)
10000 closures testPos x 35,736 ops/sec ±11.59% (98 runs sampled)
10000 closures testInvPos x 35,501 ops/sec ±11.78% (95 runs sampled)
100000 class testXYZ x 603 ops/sec ±0.40% (95 runs sampled)
100000 class testPos x 1,391 ops/sec ±0.45% (98 runs sampled)
100000 class testInvPos x 1,650 ops/sec ±0.73% (92 runs sampled)
100000 closures testXYZ x 724 ops/sec ±0.39% (95 runs sampled)
100000 closures testPos x 1,612 ops/sec ±0.47% (94 runs sampled)
100000 closures testInvPos x 1,626 ops/sec ±0.32% (95 runs sampled)
1000000 class testXYZ x 55.15 ops/sec ±0.49% (59 runs sampled)
1000000 class testPos x 132 ops/sec ±0.52% (77 runs sampled)
1000000 class testInvPos x 181 ops/sec ±0.85% (85 runs sampled)
1000000 closures testXYZ x 70.03 ops/sec ±0.54% (73 runs sampled)
1000000 closures testPos x 163 ops/sec ±0.53% (84 runs sampled)
1000000 closures testInvPos x 156 ops/sec ±0.42% (81 runs sampled)
10000000 class testXYZ x 6.09 ops/sec ±1.32% (20 runs sampled)
10000000 class testPos x 14.15 ops/sec ±0.49% (40 runs sampled)
10000000 class testInvPos x 18.02 ops/sec ±0.59% (49 runs sampled)
10000000 closures testXYZ x 7.30 ops/sec ±0.69% (23 runs sampled)
10000000 closures testPos x 16.39 ops/sec ±0.57% (45 runs sampled)
10000000 closures testInvPos x 15.23 ops/sec ±0.53% (42 runs sampled)
```

src/mol-math/geometry/symmetry-operator.ts
```diff
@@ -167,127 +168,123 @@ namespace SymmetryOperator {
...
 
-    if (isW1(m)) {
-        // this should always be the case.
-        return function projectX_W1(i: number) {
-            return xx * xs[i] + yy * ys[i] + zz * zs[i] + tx;
-        };
-    }
+        position(i: T, s: Vec3): Vec3 {
+            const m = this._m;
+            const x = this._x[i], y = this._y[i], z = this._z[i];
```
dsehnal, line 254: In this particular case, I think actually doing a local copy isn't beneficial compared to using `this.` everywhere. The only difference would if it was caching nested prop (`this.x.y`).

It should be fine to drop the `_` and just use `this.x` etc everywhere.

src/mol-math/geometry/symmetry-operator.ts
```diff
@@ -167,127 +168,123 @@ namespace SymmetryOperator {
...
-}
+            const x = this._x[i], y = this._y[i], z = this._z[i], w = (m[3] * x + m[7] * y + m[11] * z + m[15]) || 1.0;
+            return (xx * x + yy * y + zz * z + tz) / w;
+        }
+    }
 
-function isW1(m: Mat4) {
-    return m[3] === 0 && m[7] === 0 && m[11] === 0 && m[15] === 1;
-}
+    class _ArrayMappingW1<T extends number> implements ArrayMapping<T> {
```
dsehnal, line 232: Add `_ArrayMappingIdentity` as well that skips the matrix transforms altogether?

dsehnal (approved):
I agree that in the grander scheme of things, the performance difference doesn't matter and the memory savings are more important.

I am not happy about loosing the `const { x, y, z } = conformation` syntax, but lesson learned about overhead of closures.

----------------

Knowing this now, I guess I would make a "mapper" part of the symmetry operator interface that just accepts the coordinates and index, instead of always wrapping the coordinates, e.g.:

```ts
interface CoordinateMapper<T extends number> { (coords: Coordinates, index: T, slot: Vec3): Vec3 }

interface SymmetryOperatorMapper<T> {
  position: CoordinateMapper<T>
  x(coords: Coordinates, index: T): number
  y(coords: Coordinates, index: T): number
  // ...
}
```

src/mol-math/geometry/symmetry-operator.ts
arose: nice, that helps quite some

src/mol-math/geometry/symmetry-operator.ts
arose: good idea, done

PR #976, opened 2023-11-10
MVS extension

midlik:
# Description
MolViewSpec (MVS) extension allows loading MVS view descriptions into Mol*. The new functionality is available:
- in the viewer via URL parameters `mvs-url`, `mvs-data`, `mvs-format`,
- in the viewer via drag&drop for `.mvsj` files,
- programmatically via `MVSData` (including data builder and validator) and `loadMVS`.

## Actions

- [x] Added description of changes to the `[Unreleased]` section of `CHANGELOG.md`
- [x] Updated headers of modified files
- [x] Added my name to `package.json`'s `contributors`

PR #991, opened 2023-11-24
MVS extension - additional work

midlik:
# Description

Adding and fixing stuff in the MVS extension + addressing the rest of feedback from the previous PR (https://github.com/molstar/molstar/pull/976).

## Actions


src/perf-tests/array.ts
```diff
@@ -0,0 +1,39 @@
+/**
```
arose, line 1: Thanks! Could you please wrap it in a Benchmark.Suite like in src\perf-tests\sort.ts and others.

arose: Would be curious to see worst case for sortIfNeeded where only the last two items are unsorted.

PR #999, opened 2023-12-05
Expose explicit bond orders from struct_conn in mmcif

pechersky:
# Description

StructConn was referencing the wrong column name, it was using auth_seq_id instead of label_seq_id
The latter is mandatory by https://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v50.dic/Categories/struct_conn.html
This was causing no matches found during `getEntriesFromStructConn`
for building the bond lookup.

## Actions


dsehnal:
This is by design and it is because `label_seq_id` is undefined for HETATM records.

If you want to have the `label_seq_id` present as well to query it using your own code, can just add it on top of the existing entries without replacing it.

arose:
`_struct_conn.ptnr1_label_seq_id` [1] is a pointer to `_atom_site.label_seq_id` [2] which is a pointer to `_entity_poly_seq.num` [3] and only given for polymers. So for non-polymers (aka HETATM for the old PDB format) the `_struct_conn.ptnr1_label_seq_id` is always a `.` (dot) meaning "not present".

Regarding 
>  `auth_seq_id` is not necessarily a number

Yes, we are consciously (see [4]) in violation of the schema there for efficiency reasons. Entries from the  PDB always use int. If I remember correctly there where some historical reasons for it to be a string in the schema.

[1] https://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v50.dic/Items/_struct_conn.ptnr1_label_seq_id.html
[2] https://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v50.dic/Items/_struct_conn.ptnr1_label_seq_id.html
[3] https://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v50.dic/Items/_entity_poly_seq.num.html
[4] https://github.com/molstar/molstar/blob/6edbae80db340134341631f669eec86543a0f1a8/src/cli/cifschema/util/cif-dic.ts#L232-L243

pechersky:
Okay, this is great to know. I'll have to make an issue on OpenMM because right now for `struct_conn`, they serialize solely `label_seq_id`. I'll switch the implementation back to `auth_seq_id` with a fallback to `label_seq_id`.

pechersky:
There are good reasons to have non-number `seq_id` -- if looking at an MD-modeled system, one often wants to cap residues that are before/after unresolved regions, so that the resolved residues on the edge of the break don't have a naked `NH3+` or `COO-`. Those cap residues (ACE and NMA/NME, respectively) need to get a sequence id. But it would be incorrect to give them the sequential integer before/after. So for NMA/NME, which comes on the C term end, are often "numbered" like `198A` when they cap residue `198`.

pechersky:
And thanks for the clarification on looking at `_entity_poly_seq.num` with the reference to polymer -- I'm learning a lot about the PDBx semantics! Yes, `ins_code` is the better solution for the NMA issue. I was just thinking how before `ins_code`, some might have injected it directly into `auth_seq_id`.

src/mol-model-formats/structure/property/bonds/struct_conn.ts
```diff
@@ -117,13 +120,16 @@ export namespace StructConn {
             // turns out "mismat" records might not have atom name value
             if (!atomName) return undefined;
 
+            // prefer auth_seq_id, but if it is 0, then fall back to label_seq_id
+            const resId = ps.auth_seq_id.value(row) ? ps.auth_seq_id.value(row) : ps.label_seq_id.value(row);
```
dsehnal, line 124: I still don't know how I feel about this. `model.atomicHierarchy.index.findResidue` explicitly expects `auth_seq_id`, so this will only work if the file you are loading has either undefined `atom_site.auth_seq_id` (in which case the `label_` version will be used automatically for this) or if `auth_seq_id` <=> `label_seq_id` everywhere.

---------

Also, I can see the edge case where someone complains that the struct_conn doesn't work for auth seq id 0 which I am sure someone somewhere used.

dsehnal: Use `.valueKind` on the column. https://github.com/molstar/molstar/blob/master/src/mol-data/db/column.ts#L108

pechersky: Ok, I'll switch to check on valueKind. Btw, in the OpenMM files, atom site auth and label seq ids are the same.

dsehnal (commented):
I am fine with merging this. @arose any objections?

arose:
will review later today

arose (approved):
fine with me; only thing is the changelog entry, I think it should be about all struct_conn bonds, not just bond-order, right?

PR #1011, opened 2024-01-02
Fixed drag and drop overlay on WebKit and Safari

chdominguez:
# Description
On Safari, `dataTransfer` is not populated during the drag operation. Thus, the drag overlay of Mol* was never getting active.

This change checks if we are on a Safari browser by checking the `userAgent`, if so, then forces the `showDragOverlay` variable to true.

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ Note that since we don't clearly distinguish between a public and private interf
   - Labels applied in one node
 - ModelServer SDF/MOL2 ligand export: fix atom indices when additional atoms are present
 - Avoid showing (and calculating) inter-unit bonds for huge structures
+- Fixed `DragOverlay` on WebKit/Safari browsers
 
 ## [v3.43.1] - 2023-12-04
 
--- a/package.json
+++ b/package.json
@@ -103,7 +103,8 @@
     "Russell Parker <russell@benchling.com>",
     "Dominik Tichy <tichydominik451@gmail.com>",
     "Yana Rose <yana.v.rose@gmail.com>",
-    "Yakov Pechersky <ffxen158@gmail.com>"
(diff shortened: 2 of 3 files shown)


dsehnal (commented):
Thank you for the fix. Before we merge this hack, could you please have a look into using [dataTransfer.types](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/types) instead as a more robust solution?

chdominguez:
I have noticed that in Firefox non-files were also triggering the drag overlay (for example if you drag the URL from the Firefox url bar into Mol*).

I fixed that and also the Safari bug using the `dataTransfer.types` as you suggested because for both browsers when a file is inside the dragging operation the `types` property contains the string "Files". The code looks like this now:

```
onDragEnter = (ev: React.DragEvent<HTMLDivElement>) => {
        let hasFile = false;
        if (ev.dataTransfer.items && ev.dataTransfer.items.length > 0) {
            for (let i = 0; i < ev.dataTransfer.items.length; i++) {
                if (ev.dataTransfer.items[i].kind !== 'file') continue;
                hasFile = true;
                break;
            }
        } else {
            for (let i = 0; i < ev.dataTransfer.types.length; i++) {
                if (ev.dataTransfer.types[i] !== 'Files') continue;
                hasFile = true;
                break;
            }
        }

        if (hasFile) {
            this.showDragOverlay.next(true);
        }
    };
```

Notice that I have changed the first `if` condition to account for the items length in order to correctly enter into the `else` clause. I replaced the else code to check only for "Files", as it is always that either on Chrome, Firefox or Safari.

dsehnal:
Looks good, did you push your changes?

chdominguez:
Pushed them right now!

dsehnal (approved):

PR #1012, opened 2024-01-02
check applicability of mvs repr/color

arose:
# Description

## Actions

### DIFF
--- a/src/extensions/mvs/components/annotation-color-theme.ts
+++ b/src/extensions/mvs/components/annotation-color-theme.ts
@@ -11,7 +11,7 @@ import { ThemeDataContext } from '../../../mol-theme/theme';
 import { ColorNames } from '../../../mol-util/color/names';
 import { ParamDefinition as PD } from '../../../mol-util/param-definition';
 import { decodeColor } from '../helpers/utils';
-import { getMVSAnnotationForStructure } from './annotation-prop';
+import { MVSAnnotationsProvider, getMVSAnnotationForStructure } from './annotation-prop';
 
 
 /** Parameter definition for color theme "MVS Annotation" */
@@ -76,5 +76,5 @@ export const MVSAnnotationColorThemeProvider: ColorTheme.Provider<MVSAnnotationC
     factory: MVSAnnotationColorTheme,
     getParams: ctx => MVSAnnotationColorThemeParams,
     defaultValues: PD.getDefaultValues(MVSAnnotationColorThemeParams),
-    isApplicable: (ctx: ThemeDataContext) => true,
+    isApplicable: (ctx: ThemeDataContext) => !!ctx.structure?.models.some(m => m.customProperties.has(MVSAnnotationsProvider.descriptor)),
 };
--- a/src/extensions/mvs/components/annotation-label/representation.ts
+++ b/src/extensions/mvs/components/annotation-label/representation.ts
@@ -9,6 +9,7 @@ import { Representation, RepresentationContext, RepresentationParamsGetter } fro
 import { ComplexRepresentation, StructureRepresentation, StructureRepresentationProvider, StructureRepresentationStateBuilder } from '../../../../mol-repr/structure/representation';
 import { MarkerAction } from '../../../../mol-util/marker-action';
 import { ParamDefinition as PD } from '../../../../mol-util/param-definition';
+import { MVSAnnotationsProvider } from '../annotation-prop';
 import { MVSAnnotationLabelTextParams, MVSAnnotationLabelTextVisual } from './visual';
 
 
@@ -45,5 +46,5 @@ export const MVSAnnotationLabelRepresentationProvider = StructureRepresentationP
     defaultValues: PD.getDefaultValues(MVSAnnotationLabelParams),
     defaultColorTheme: { name: 'uniform' }, // this ain't workin
     defaultSizeTheme: { name: 'physical' },
-    isApplicable: (structure: Structure) => structure.elementCount > 0,
+    isApplicable: (structure: Structure) => structure.elementCount > 0 && !!structure.models.some(m => m.customProperties.has(MVSAnnotationsProvider.descriptor)),
 });
--- a/src/extensions/mvs/components/custom-label/representation.ts
+++ b/src/extensions/mvs/components/custom-label/representation.ts
@@ -9,6 +9,7 @@ import { Representation, RepresentationContext, RepresentationParamsGetter } fro
 import { ComplexRepresentation, StructureRepresentation, StructureRepresentationProvider, StructureRepresentationStateBuilder } from '../../../../mol-repr/structure/representation';
 import { MarkerAction } from '../../../../mol-util/marker-action';
 import { ParamDefinition as PD } from '../../../../mol-util/param-definition';
+import { MVSAnnotationsProvider } from '../annotation-prop';
 import { CustomLabelTextParams, CustomLabelTextVisual } from './visual';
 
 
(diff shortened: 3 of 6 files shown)


arose:
@midlik made some tweaks to the MVS extensions so that the MVS related representations and color themes are only applicable (show up in the UI) when the structure has a model with a MVSAnnotationsProvider.

midlik:
Hi @arose , your change makes sense but only for `annotation-color-theme` and `annotation-label`. 
For `custom-label` and `multilayer-color-theme` it doesn't make sense, as these do not use `MVSAnnotationsProvider` and thus should be applicable always.

midlik:
Another thing is: your implementation only checks that the provider is there, but not if it has a non-empty list of annotation sources. The check should probably look more like this:

```
        // isApplicable: (ctx: ThemeDataContext) => !!ctx.structure?.models.some(m => m.customProperties.has(MVSAnnotationsProvider.descriptor)),
        isApplicable: (ctx: ThemeDataContext) => !!ctx.structure?.models.some(m => m.customProperties.has(MVSAnnotationsProvider.descriptor) && MVSAnnotationsProvider.props(m).annotations.length > 0),
```

Note: MVS load function is currently implemented in such a way that it always adds a MVSAnnotationsProvider custom property, even when the list of annotation sources is empty. Therefore your `isApplicable` always returns true if the model has been loaded via the MVS load function.

src/extensions/mvs/components/annotation-color-theme.ts
```diff
@@ -76,5 +76,5 @@ export const MVSAnnotationColorThemeProvider: ColorTheme.Provider<MVSAnnotationC
     factory: MVSAnnotationColorTheme,
     getParams: ctx => MVSAnnotationColorThemeParams,
     defaultValues: PD.getDefaultValues(MVSAnnotationColorThemeParams),
-    isApplicable: (ctx: ThemeDataContext) => true,
+    isApplicable: (ctx: ThemeDataContext) => !!ctx.structure?.models.some(m => m.customProperties.has(MVSAnnotationsProvider.descriptor)),
```
midlik, line 79: I would make it:
```
isApplicable: (ctx: ThemeDataContext) => !!ctx.structure?.models.some(m => m.customProperties.has(MVSAnnotationsProvider.descriptor) && MVSAnnotationsProvider.props(m).annotations.length > 0),
```

src/extensions/mvs/components/annotation-label/representation.ts
```diff
@@ -45,5 +46,5 @@ export const MVSAnnotationLabelRepresentationProvider = StructureRepresentationP
     defaultValues: PD.getDefaultValues(MVSAnnotationLabelParams),
     defaultColorTheme: { name: 'uniform' }, // this ain't workin
     defaultSizeTheme: { name: 'physical' },
-    isApplicable: (structure: Structure) => structure.elementCount > 0,
+    isApplicable: (structure: Structure) => structure.elementCount > 0 && !!structure.models.some(m => m.customProperties.has(MVSAnnotationsProvider.descriptor)),
```
midlik, line 49: I would make it:
```
isApplicable: (ctx: ThemeDataContext) => structure.elementCount > 0 && !!structure.models.some(m => m.customProperties.has(MVSAnnotationsProvider.descriptor) && MVSAnnotationsProvider.props(m).annotations.length > 0),
```

src/extensions/mvs/components/custom-label/representation.ts
```diff
@@ -38,12 +39,12 @@ export function CustomLabelRepresentation(ctx: RepresentationContext, getParams:
...
-    label: 'Custom Label',
+    label: 'MVS Custom Label',
     description: 'Displays labels with custom text',
     factory: CustomLabelRepresentation,
     getParams: () => CustomLabelParams,
     defaultValues: PD.getDefaultValues(CustomLabelParams),
     defaultColorTheme: { name: 'uniform' },
     defaultSizeTheme: { name: 'physical' },
-    isApplicable: (structure: Structure) => structure.elementCount > 0
+    isApplicable: (structure: Structure) => structure.elementCount > 0 && !!structure.models.some(m => m.customProperties.has(MVSAnnotationsProvider.descriptor)),
```
midlik, line 49: I would revert this change, CustomLabelRepresentationProvider should be applicable even without MVSAnnotationsProvider.

(That's also the reason I didn't use the 'MVS ' prefix in the label, but I don't have a strong opinion about it)

src/extensions/mvs/components/multilayer-color-theme.ts
```diff
@@ -137,11 +138,11 @@ export const MultilayerColorThemeName = 'mvs-multilayer';
...
     return {
         name: MultilayerColorThemeName,
-        label: 'Multi-layer',
+        label: 'MVS Multilayer',
         category: ColorTheme.Category.Misc,
         factory: (ctx, props) => makeMultilayerColorTheme(ctx, props, colorThemeRegistry),
         getParams: (ctx: ThemeDataContext) => makeMultilayerColorThemeParams(colorThemeRegistry, ctx),
         defaultValues: DefaultMultilayerColorThemeProps,
-        isApplicable: (ctx: ThemeDataContext) => true,
+        isApplicable: (ctx: ThemeDataContext) => !!ctx.structure?.models.some(m => m.customProperties.has(MVSAnnotationsProvider.descriptor)),
```
midlik, line 146: I would revert this change, makeMultilayerColorThemeProvider should be applicable even without MVSAnnotationsProvider.

(That's also the reason I didn't use the 'MVS ' prefix in the label, but I don't have a strong opinion about it)

midlik (changes requested):

arose:
@midlik I would like to separate the reprs/themes from non-MVS use. My thinking is that we either have something generic that can be in core (or maybe a generic extension). Since the 'mvs-multilayer' color theme and 'mvs-custom-label' representation are custom for the mvs extension I would like them to only be applicable in the MVS context. Instead of the MVSAnnotationsProvider I am OK to add some other custom property to a model/structure to mark them as being used in the MVS context.

arose:
@midlik what do you think of adding some kind of MVS property to mark models as being used in a MVS context? Would you propose it? Or just use MVSAnnotationsProvider for now?

midlik:
Hi, now I see you point.
The solution with a "MVS flag" custom property makes sense, I'm just not sure if it's worth it to introduce a new property just for this. I'll discuss this with @dsehnal (we have a meeting on Monday) and let you know.

dsehnal:
I think having 

```ts
const IsMVSProperty = {  // using object to avoid namespace
  set(m: Model) {
    m._staticPropertyData._isMVS_ = true;
  },
  get(m: Model) {
    return !!m._staticPropertyData._isMVS_;
  }
}
```

and setting this for all models that have been added by the extension would be a good solution.

One way to automate this would be to set `mvs-model` tag when adding the model nodes (`.apply(..., { tags: ['mvs-model'] });` and  then `plugin.state.data.selectQ(q => q.root.subtree().withTag('mvs-model')).forEach(cell => IsMVSProperty.set(cell.obj?.data))`... or just keep the list of model refs when constructing the tree.

midlik:
I will implement the "MVS flag" custom property and create a separate PR.

midlik:
I created the alternative PR: https://github.com/molstar/molstar/pull/1022

arose:
> Now I'm thinking that the "Custom Label" 3D representation could find some use cases even outside the MVS context.

there is also the Label option in the Measurement section that offers similar functionality, maybe that can be extended if need be?

arose:
closing in favor of #1022

midlik:
> > Now I'm thinking that the "Custom Label" 3D representation could find some use cases even outside the MVS context.
> 
> there is also the Label option in the Measurement section that offers similar functionality, maybe that can be extended if need be?

Oh, I was not aware of that.
Then ignore my suggestion.

PR #1036, opened 2024-01-30
Move most headless context deps to optional peer dependencies and remove `@types/jpeg-js`

### LINKED ISSUE #1031: Moving gl and friends to optional peer dependencies
Hi! It seems that on Node 20 we've run into trouble where `gl` fails to build but still takes a long time to fail and contributes to longer build times (even though we're not using the headless context at the moment).

We had some discussion in https://github.com/molstar/molstar/issues/767 about these particular dependencies that are relevant to the `HeadlessPluginContext` and now I'm thinking that it makes more sense to make them optional peer dependencies. 

Does that seem like a reasonable solution? I can put together a patch if so

cc @midlik

russellp17:
# Description

resolves #1031 

Based on discussion in issue this PR moves the optional dependencies related to running mol* on Node (i.e. in a headless context) to be optional peer dependencies instead of just optional dependencies.

One dependency that I made an exception for was `jpeg-js` - since the `@types/jpeg-js` package is now [just a stub](https://www.npmjs.com/package/@types/jpeg-js) (i.e. it does not actually provide any types), I made `jpeg-js` a regular dependency because the build will fail without the types provided by the package. I _think_ an alternative would be to make it both a devDependency and an optional peer dependency but given that it is pure JS and relatively small, I was not too concerned about making it a regular dependency.

Confirmed that the build succeeds locally after manually removing the other dependencies (`gl`, `canvas`, and `pngjs`) from `node_modules/`.

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Note that since we don't clearly distinguish between a public and private interf
   - Support `multiline` inputs
   - Support `placeholder` parameter
   - Support `disableInteractiveUpdates` to only trigger updates once the control loses focus
+- Move dependencies related to the headless context from optional deps to optional peer deps
 
 ## [v3.44.0] - 2023-01-06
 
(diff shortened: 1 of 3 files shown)


arose:
thank you @russellp17

PR #1053, opened 2024-02-28
Remove 2ICY from list

nezix:
2ICY does not have different sets of ligands, or missing ligands in different models

# Description

## Actions


arose:
There are different ligands (UPG vs U5P) in models 1 vs 2 in the second assembly.

PR #1056, opened 2024-03-01
VolumeTransform to rotate a volume, like TransformStructureConformation

### LINKED ISSUE #1055: Transforming a Volume under a rotation and translation
We are depicting structures and volumes together. We can calculate a transform matrix between structures, and use it in a transform step to superpose. We'd like to also shift the volumes over (we have a structure-volume correspondence). Right now, it doesn't seem like there is a VolumeTransform like PluginStateTransform. What's the best way to achieve this? Is it modifying a Volume.Data.Grid.Tensor.Space?

pechersky:
# Description
Provide a state transformer to rotate/translate volumes in a similar manner to structures, like in superposition.

Closes #1055.

## Actions


arose:
thanks

Issue #1040, opened 2024-02-06
Export Models fails on BCIF from ModelServer


midlik:
When I download model from https://www.ebi.ac.uk/pdbe/model-server/v1/6wvp/atoms?auth_asym_id=B&encoding=bcif,
it shows OK in Mol* Viewer, but when I try to export model (via the right hand control panel), I get this error:
```
Model export failed. See console for details.
```

And the console says:
```
molstar.js:2 Error: unreachable
    at he (molstar.js:2:1773657)
    at Ql (molstar.js:2:2087019)
    at molstar.js:2:2085120
    at hl (molstar.js:2:2085190)
    at dl (molstar.js:2:2084359)
    at molstar.js:2:2189142
    at Object.getField (molstar.js:2:2189609)
    at Ha.get [as program_version] (molstar.js:2:1993792)
    at molstar.js:2:2090663
    at Ol (molstar.js:2:2101530)
```
![download](https://github.com/molstar/molstar/assets/14818844/880188ff-61e0-4392-88ae-2f61fc92d1d8)
![export](https://github.com/molstar/molstar/assets/14818844/fb00d80b-9674-492b-a060-da62fb03490b)

When I try the same with encoding=cif, the export works fine.

midlik:
Getting the data from a different ModelServer instance doesn't help (https://models.rcsb.org/v1/6wvp/atoms?auth_asym_id=B&encoding=bcif)

dsehnal:
#1048 should address this

JonStargaryen:
Weighing in because this also affects the Java codec and isn't as easy to fix... I think this might be an issue in the ModelServer code. A while back I saw that ModelServer will chain ByteArray encodings together, which doesn't make sense and interestingly doesn't happen when the viewer or cif2bcif is used to create a BinaryCIF file.

I wanted to dig into this for quite a while now but didn't find the time yet. Maybe someone has an idea.

![double-ba](https://github.com/molstar/molstar/assets/9744615/86bf68b5-c56a-443a-bcba-63b1f0bc00fe)

dsehnal:
The cif2bcif works differently. The bug will be either in the determination heuristic for the "schema-based cif fields" or error in manually assigning the encoding for this particular field. Probably gonna be the first one.

dsehnal:
I managed to find the issues (or at least some of them to make this work). See https://github.com/molstar/molstar/pull/1057

PR #1057, opened 2024-03-03
BinaryCIF encoder fixes

dsehnal:
# Description

Fixes BinaryCIF encoder edge cases caused by re-encoding an existing BinaryCIF file. See #1040.

The issues:
- The `ArrayEncoder.fromEncoding` added extra `byte-array` encoding due to implementation details (`integer-packing` adds `byte-array` automatically).
- The automatic column encoding classifier could determine a column has a numeric type on per-entry basis (e.g. using the `cif2bcif` tool), but in the schema the column would be marked as `str`. This caused the encoder to apply numeric-only encoding to a string type, resulting a weird behavior (e.g. in `pdbx_chem_comp_identifier.program_version`)

## Actions


arose (approved):
lgtm

JonStargaryen (approved):
Works nicely now, thanks.

Issue #1058, opened 2024-03-05
Label rendering not working properly in headless mode


midlik:
Hi, there was some change in 4.0.0 that broke rendering in headless mode. @arose tried to fix it by these commits:
[Revert "require WEBGL_depth_texture extension](https://github.com/molstar/molstar/commit/0a4643ebe44de27d7c407486edc80739be15ad75)
[revert renderbuffer removal](https://github.com/molstar/molstar/commit/8dc3dae81e65cf200837fa21b83a28032c5df362)

Now it runs and renders images, but labels in the images are not rendered correctly. Compare images created by v3.45.0 and v4.0.1:
![1cbs-old](https://github.com/molstar/molstar/assets/14818844/b5689208-4637-41d5-ae02-0ed3224b8e83)
![1cbs-new](https://github.com/molstar/molstar/assets/14818844/d5af459c-2650-4c59-b344-fc413986d757)

Generated by this command: `node lib/commonjs/cli/mvs/mvs-render.js -i examples/mvs/1cbs.mvsj -o ../tmp/1cbs.png`

Could you please have a look at what could be wrong?

arose:
I think it is trying to do wboit in headless which I think never worked but was not a problem because before 4 transparency was not part of the state.

Before wboit was an init param, now it is runtime and part of the state. I think we never had wboit on in headless mode.

It is a bug but workaround would be to force blended transparency.

midlik:
Hi, thank you for a pointer.
However, I experimented a bit and I'm not sure this is the issue.
In `src/mol-canvas3d/passes/draw.ts`, `this.wboit.supported` gives `false`, so it falls back from 'wboit' to 'blended'.
So it seems like the problem is somewhere else.

(BTW, `this.dpoit.supported` gives `true` but when I try to use 'dpoit', it renders a blank image. In Molstar 3.45 there was `this.wboitEnabled` and `this.dpoitEnable`, both giving `false` when run from Node)

midlik:
I've been able to identify the commit that broke the things: https://github.com/molstar/molstar/commit/dd2593475ab52e608fd77b97eb46227e537c87b5

midlik:
OK, I found it. It was really trying to do 'wboit'. What threw me off was that there is transparency in `DrawPass` which automatically falls back to 'blended' when 'wboit' is not supported, but there is also transparency in `Scene` which can be set to whatever. Perhaps this should be addressed, but I don't want to touch your code as I don't really know what it does.
[...]

PR #1069, opened 2024-03-23
Fix transparency rendering fallback

arose:
Fix for #1058, and follow-up to #1068


midlik (approved):
Thank you @arose 
It looks fine, and headless rendering of labels still works as it should.

PR #1075, opened 2024-03-30
refactor transparency rendering

arose:
- more uniform behavior for blended, wboit, dpoit
- fix issues with text & image geometry


arose:
@giagitom @midlik please have a look and check that it does not break anything for your use cases, thank you!

giagitom (approved):
Hi @arose, everything seems to work fine on my side!  Thanks

PR #1083, opened 2024-04-14
Emissive & Bloom

arose:
Adds emissive material support and bloom post-processing

image


giagitom (commented):
Hi @arose,
Awesome new features! 

Here are my considerations and the issues I have found so far:

- Is it used a uniform emission color (like white for example), or is it used a different color for each fragment?
- Transparent background does not work when bloom is active and emission is on.
- Also I noticed blooming effect isn't working on transparent objects. Emission is working instead, but produces strange effects like this one in the image. I applied transparency on both residues in the middle and emission to only one of them. It seems blending transmits the emission effect to the overlapping opaque mesh.
image

giagitom:
An other thing I noticed is that also highlight color outline changes color when hovering over emitting residues. Is this wanted?

arose:
Thanks for testing @giagitom!

> Is it used a uniform emission color (like white for example), or is it used a different color for each fragment?

I use an emission factor that gets multiplied with the underlying color.

> Transparent background does not work when bloom is active and emission is on.

Fixed

> Also I noticed blooming effect isn't working on transparent objects.

No bloom on transparent objects (at least for now). I fixed the unintended effect you saw.

> An other thing I noticed is that also highlight color outline changes color when hovering over emitting residues. Is this wanted?

Yes, that is expected.

giagitom:
>No bloom on transparent objects (at least for now). I fixed the unintended effect you saw.

Ok good to know, is this due to a specific problem you are facing?


>Transparent background does not work when bloom is active and emission is on.

Now the background transparency is working fine! I noticed a few other issues on the canvas and while downloading a snapshot though.

There are blending issues with emission and transparent fog effect (not present in the canvas but only in the downloaded images)
image

I can also see a white halo external to outlines on transparent edges that are near to emitting objects
image

arose:
thanks again for testing

> There are blending issues with emission and transparent fog effect (not present in the canvas but only in the downloaded images)

that is tricky, I found a workaround that helps

> I can also see a white halo external to outlines on transparent edges that are near to emitting objects

did not see that, if it is still there can you share a state/steps to reproduce?

arose:
> Ok good to know, is this due to a specific problem you are facing?

overwhelming complexity... lack of time... diminishing returns...

giagitom:
Hi @arose,
Thanks for the fix, now it works.
Regarding bloom effect I see it disappears on background areas when transparency is active (both on canvas and on downloaded image). 

I'm wondering in which way this effect is different from transparent meshes/impostors on transparent background? Is there some way to ripristinate it?

Here a test state to to verify previous issue with white halo (only visible on canvas) [test-state.molj.txt](https://github.com/molstar/molstar/files/15151665/test-state.molj.txt)

arose:
> Regarding bloom effect I see it disappears on background areas when transparency is active (both on canvas and on downloaded image).

Yeah, I need to rethink how I blend everything together

giagitom:
A new issue I have found on surface representation when trying to apply a layer of transparency and a layer of emission in the same representation. This test state reproduces the error [test-state.molj.txt](https://github.com/molstar/molstar/files/15168515/test-state.molj.txt)

arose:
thank @giagitom, another round of fixes, but I think this will do it

giagitom (approved):
Hi @arose ,
Thanks for all fixes. Blooming looks very nice and has a great potential. Please, feel free to merge this as it is.

I just want to report a few differences i noticed on this comparison

- black background snapshot
- transparent background composed on gimp with both white and black background
- white background snapshot 

I'm wondering if it is possible to enhance the effect of blooming on opaque background to uniform it to the tranpsarent backgorund?
Also is it possible to apply a better bloom smoothing effect that sometimes could be useful? (maybe only when exporting snapshots if this would lead to performances issues) 

test

Also I see that blooming is applied to the axis helper. Is this wanted?

image

rtviii:
This looks slick as hell, Alex.

Issue #1085, opened 2024-04-16
Add state tree params to `addLabel`


rjdirisio:
Hello,

We have a custom viewer that labels certain parts of the protein using `plugin.managers.structure.measurement.addLabel`. We would like to have more control over making this label visible/collapsible in the state tree (via `isGhost`, etc.), as well as removing it from the state tree.

`addLabel` creates a separate `Measurement` node to the state tree (outside the `Structure`'s tree), while also associating the `Label` with the underlying `Structure` via a `dependsOn`. In the picture, one can see the label is in the separate `Measurement` node from the (redacted) structure tree that its label is associated with:
![image](https://github.com/molstar/molstar/assets/35471342/23c39bc1-f044-46f6-b841-3c528a38f249)
When one removes the structure, this leaves the empty `Measurement` node behind (with no dropdown data, because the associated `Label` was removed along with the structure)
![image](https://github.com/molstar/molstar/assets/35471342/a329f203-9d70-4483-bc5d-f816f57cbf8e)
We don't see a way to clear these upon deletion aside from manual removal from the state tree, or check, after deletion, for any nodes in the state tree that are `Measurements` with no children.

I see a few ways to address this:
- If we had the ability to hide the measurement node, or the measurement node and its children, from the state tree at the outset, this would be less visible in the front end. This would involve exposing the `state` param object to addLabel, likely for both state transforms there (`MultiStructureSelectionFromExpression `and `StructureSelectionsLabel3D`, or just `MultiStructureSelectionFromExpression`)
- Less direct: Pass back the ref associated with the Measurement / label when one calls addLabel. Currently, there's not a great way to programmatically find that label after addLabel is called, since we just get back a `Promise<void>`.
    - Similarly, one _could_ provide an API in the `StructureMeasurementManager` to delete measurements, but, I think this could be accomplished by passing refs instead because the `StructureHierarchyManager` already has a `remove` method that should work for that.

Happy to clarify any points, thanks!

dsehnal:
Hi, more customization options, including deleting the group when it's empty, could  definitely be added. If I understand it correctly, you just want the ability to hide the measurements group completely from the state tree?

As for the ability to select the labels after creation: this could be done either by you manually keeping the refs/selectors, or using the `tag` option on transforms. I think additional options would need to be added to support the latter as well.

rjdirisio:
> If I understand it correctly, you just want the ability to hide the measurements group completely from the state tree?

[...]

PR #1087, opened 2024-04-19
return selectors for each StateTransform in each measurement

rjdirisio:
# Description

Partially addresses #1085 

Return information about the state transforms performed in the `StructreMeasurementManager`'s API. This PR returns the `StateObjectSelectors` that were applied in the API call, similar to how it was done for [PsfProvider](https://github.com/molstar/molstar/blob/a9e13732037c152bbd7094acca3cc4ac6162ee62/src/mol-plugin-state/formats/topology.ts#L27). This allows callers of these functions to get attributes like the `ref`, which is useful for fully deleting the measurements, etc.

A series of state transforms are done in a loop (for each of the loci) on `addOrderLabels`. So, just the final `orderLabel` is returned. I'm also happy to leave this function as a TODO, or not modify it at all, since it's technically incomplete.

## Actions


src/mol-plugin-state/manager/structure/measurement.ts
```diff
@@ -125,6 +127,8 @@ class StructureMeasurementManager extends StatefulPluginComponent<StructureMeasu
 
         const state = this.plugin.state.data;
         await PluginCommands.State.Update(this.plugin, { state, tree: update, options: { doNotLogTiming: true } });
+        return { distance: distance.selector, distanceRep: distanceRep.selector };
```
dsehnal, line 130: Can you please either name the first argument to `distanceSelection` and the second to `distanceRepresentation`, or just use `selection` and `representation` everywhere?

rjdirisio: Sure thing, made it uniform in https://github.com/molstar/molstar/pull/1087/commits/f986688857838effeb2d0a101ca878a54f572a44

dsehnal (approved):
LGTM, thanks.

PR #1101, opened 2024-04-23
Load mvs keep camera

midlik:
# Description
I added `keepCamera` option to `loadMVS` function in MolViewSpec extension. 
I also slightly changed the way `radiusMax` is computed, to avoid aggressive clipping in cases when the camera target is not aligned with the scene's bounding sphere.

## Actions


midlik:
@arose could you please check the change I made in `[src/mol-canvas3d/canvas3d.ts](https://github.com/molstar/molstar/pull/1101/files#diff-a387ffcb1f053ad2d4ad3f869712109e398311a2846d9eb2a5db831956356adf)`?

@dsehnal was not sure if it can break anything.

src/mol-canvas3d/canvas3d.ts
```diff
@@ -602,7 +602,8 @@ namespace Canvas3D {
                 const focus = camera.getFocus(center, radius);
                 const next = typeof nextCameraResetSnapshot === 'function' ? nextCameraResetSnapshot(scene, camera) : nextCameraResetSnapshot;
                 const snapshot = next ? { ...focus, ...next } : focus;
-                camera.setState({ ...snapshot, radiusMax: getSceneRadius() }, duration);
+                const offset = snapshot.target ? Vec3.distance(snapshot.target, scene.boundingSphere.center) : 0;
```
arose, line 605: Hmm, could you set `sceneRadiusFactor` instead? This definitely breaks things.

dsehnal: Thanks, that makes more sense.

dsehnal (approved):

src/mol-canvas3d/canvas3d.ts
midlik: Do you mean like this?: https://github.com/molstar/molstar/pull/1101/commits/00ad11dd6b170ad5815c12588dbd173aca47b01d

A disadvantage of tweaking `sceneRadiusFactor` is that it then stays changed forever (`radiusMax` in the original solution would go back to normal on the next camera reset).

arose: I suppose if you don't want the aggressive clipping, you want it forever?

Also note that you have access to the camera object and could change things there from your extension.

dsehnal: The issue this is trying to solve it this:

image

Without either allowing bigger `radiusMax` or increasing the factor it seems quite hard to fix this, where the use case is loading a different state without moving the camera.

Any other ideas how to solve this @arose?

arose: > Any other ideas how to solve this @arose?

nothing at the moment

arose (approved):

PR #1103, opened 2024-04-24
DOF shader

corredD:
DOF planar effect
image
Screenshot 2024-04-24 122900
image

### DIFF
--- a/src/mol-gl/shader/shadows.frag.ts
+++ b/src/mol-gl/shader/shadows.frag.ts
@@ -80,19 +80,19 @@ float screenSpaceShadow(const in vec3 position, const in vec3 lightDirection, co
         rayCoords = uProjection * vec4(rayPos, 1.0);
         rayCoords.xyz = (rayCoords.xyz / rayCoords.w) * 0.5 + 0.5;
 
-        if (outsideBounds(rayCoords.xy))
-            return 1.0;
+        //if (outsideBounds(rayCoords.xy))
+        //    return 1.0;
 
         // Compute the difference between the ray's and the camera's depth
         float depth = getDepth(rayCoords.xy);
-        float viewZ = getViewZ(depth);
-        float zDelta = rayPos.z - viewZ;
+        float sampleViewZ = screenSpaceToViewSpace(vec3(rayCoords.xy, depth), uInvProjection).z;
+        float zDelta = rayPos.z - sampleViewZ;
 
         if (zDelta < uTolerance) {
             occlusion = 1.0;
 
             // Fade out as we approach the edges of the screen
-            occlusion *= screenFade(rayCoords.xy);
+            // occlusion *= screenFade(rayCoords.xy);
 
             break;
         }
--- a/src/mol-canvas3d/passes/draw.ts
+++ b/src/mol-canvas3d/passes/draw.ts
@@ -24,6 +24,7 @@ import { MarkingPass, MarkingProps } from './marking';
 import { CopyRenderable, createCopyRenderable } from '../../mol-gl/compute/util';
 import { isDebugMode, isTimingMode } from '../../mol-util/debug';
 import { AssetManager } from '../../mol-util/assets';
+import { DofPass } from './dof';
 
 type Props = {
     postprocessing: PostprocessingProps;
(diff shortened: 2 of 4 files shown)


dsehnal:
Cool

arose:
Hi, two quick comments

- It should use the output from the anti-alias pass as input when enabled.
- I struggle to find working values for "In Focus" and "PPM". Could you change it so it is expressed in percentage values of the (visible) scene bounding sphere radius? At least as an option.

corredD:
I use the scene bounding sphere so its easier to find the proper area to be in focus. It will be the area at distance inFocus from the scene center. PPM is the size of the area. 
image
I am trying to use the anti-alias pass, but when I use this.antialiasing.target.texture, the texture is emtpy ? 
How should the target be defined ? 
const input = antialiasingEnabled ? this.antialiasing.target.texture : (postprocessingEnabled ? this.postprocessing.target.texture : this.drawTarget.texture);
Lastly I would like to be able to pass a toggle option, where the area in focus could be either the scene center or the current selection or mark. How could I retrieve this information of current loci/mark position from the render pass ?

arose:
> I am trying to use the anti-alias pass, but when I use this.antialiasing.target.texture, the texture is emtpy ?
> How should the target be defined ?
> const input = antialiasingEnabled ? this.antialiasing.target.texture : (postprocessingEnabled ? this.postprocessing.target.texture : this.drawTarget.texture);

will have to have a look

> Lastly I would like to be able to pass a toggle option, where the area in focus could be either the scene center or the current selection or mark. How could I retrieve this information of current loci/mark position from the render pass ?

I think that is best done via a behavior that listens to selection changes and sets the dof params accordingly

arose:
One thing I notices (specifically with 1TQN or another small example) is that there is blur even in the middle of the focus region.

arose:
Very nice addition with he camera target. Should it be the default?

Should PPM be exponential to make it more use-friendly in UI? Or maybe we should add ParameterControl opion for that, @dsehnal?

Parts that are in focus get some blurry halo, see below. Can the halo be avoided?

image

src/mol-canvas3d/passes/dof.ts
```diff
@@ -0,0 +1,193 @@
+/**
+ * Copyright (c) 2020 mol* contributors, licensed under MIT, See LICENSE file for more info.
```
arose, line 2: please change the date and use your name

src/mol-gl/shader/dof.frag.ts
```diff
@@ -0,0 +1,97 @@
+/**
+ * Copyright (c) 2019-2022 mol* contributors, licensed under MIT, See LICENSE file for more info.
```
arose, line 2: correct date and your name please

src/mol-gl/shader/shadows.frag.ts
```diff
@@ -80,19 +80,19 @@ float screenSpaceShadow(const in vec3 position, const in vec3 lightDirection, co
         rayCoords = uProjection * vec4(rayPos, 1.0);
         rayCoords.xyz = (rayCoords.xyz / rayCoords.w) * 0.5 + 0.5;
 
-        if (outsideBounds(rayCoords.xy))
-            return 1.0;
+        //if (outsideBounds(rayCoords.xy))
```
arose, line 83: are changes to this file needed for DoF?

corredD: not need for dof, should if be a different PR ?

dsehnal:
> Should PPM be exponential to make it more use-friendly in UI? Or maybe we should add ParameterControl opion for that, @dsehnal?

Can you be a little more specific please?

arose:
> > Should PPM be exponential to make it more use-friendly in UI? Or maybe we should add ParameterControl opion for that, @dsehnal?
> 
> Can you be a little more specific please?

exponential step-size, so you have finer control at smaller values

arose (approved):
did a few tweaks (see my commits), good to go from my side

src/mol-gl/shader/shadows.frag.ts
arose: removed it

Issue #1082, opened 2024-04-12
Occlusion Blurr kernel issue 


giagitom:
I think I have found something while playing with occlusion on mesoscale explorer. 
There is a strange halo on darker points of the canvas that disappears disabling occlusion blurring

**Blur kernel size: 3**
![image](https://github.com/molstar/molstar/assets/20627977/db0fc07e-1593-48c0-bc74-50da2c1bfb6d)

**Blur kernel size: 1**
![image](https://github.com/molstar/molstar/assets/20627977/f306cc5e-6414-4214-9ab5-2ae0247889eb)

giagitom:
It seems also related to Multiscale occlusion

arose:
Hmm, not sure where this is coming from. It is not from multi-scale. If you switch it of and set radius to 12.5 you get it too. It might be because we are reconstructing normals from depth and screen position.

corredD:
looks like its not only a normal issue ( I tried the more accurate calculation from https://gist.github.com/bgolus/a07ed65602c009d5e2f753826e8078a0) . Its seems to be rather the ssao-blur, and especially the depth threshold and its test. Have a look at the patch :  
[ssaopatch.patch](https://github.com/molstar/molstar/files/15081417/ssaopatch.patch)

![Screenshot 2024-04-23 104719](https://github.com/molstar/molstar/assets/539461/06db0628-eaad-4f4f-ac3f-536f18b4483a)

PR #1104, opened 2024-04-25
Fix SSAO artifacts (@corredD, #1082)

arose:
See #1082, thanks @corredD for the patch and thanks @giagitom for identifying the issue.


giagitom (approved):
Nice, It seems artifacts are gone and everything is looking good.

PR #1110, opened 2024-04-28
fix bumpiness artifact

### LINKED ISSUE #1107: Issue using Global Bumpiness !== 0 and  Layer level Bumpiness === 0
I noticed this issue on cartoons and also on impostor cylinders (but not on impostor shperes)
image

arose:
Fixes #1107


giagitom (approved):
Thanks @arose,
It works as expected now

Issue #641, opened 2022-12-01
Ring in NAI is detected as aromatic


JonStargaryen:
C4N in [NAI](https://www.rcsb.org/ligand/NAI) is sp3 hybridized but Mol* marks the corresponding ring as aromatic (https://molstar.org/viewer/?load-pdb=1t2c). This is because all other criteria for aromaticity hold and the ring happens to be planar enough to pass the final planarity check.

![ring1](https://user-images.githubusercontent.com/9744615/204894108-5d848072-3a0a-4f07-ba27-0525234646dd.png)

![ring2](https://user-images.githubusercontent.com/9744615/204894131-fd3cde88-278b-4610-b12e-313b5125cfbe.png)

We could check whether a ring contains >0 double bonds and skip the planarity check if there's a carbon with only single bonds (however, that will fail for the cyclopentadienyl anion). Don't think there's a clean solution here without counting pi-electrons. What do you think?

arose:
If `chem_comp_bonds` are included (which is not generally the case) in the mmcif file you can remove `computed` from the `includeTypes` param. An improvement could be to check in the repr getParams function if the source data of all models of a structure are of mmcif fromat and have the `chem_comp_bonds` category, then remove `computed` from the `includeTypes` param.

JonStargaryen:
Ah, I see, knowledge of double bonds would also require presence of the `chem_comp_bond` category... But good to know that the inclusion of computed bonds can be toggled in the UI. Thanks!

PR #1113, opened 2024-05-03
Stop showing `computed` bonds if `chem_comp_bond` is defined

JonStargaryen:
# Description
Follow-up to #641, implementing Alex' proposal.

Works for standard ball-and-stick but doesn't play nicely with focus, which includes `computed` interactions by default (and adding it to `excludeTypes` removes other bonds, but removing it from `includeTypes` gives the visuals I'd like, at least for [1t2c](https://molstar.org/viewer/?load-pdb=1t2c)).

Do you have an idea for that? All archive mmCIF will have `chem_comp_bond` by the end of the year, currently available for ~160k.

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Note that since we don't clearly distinguish between a public and private interf
     - Wrong operator checks in `findPairBonds`
 - Fix SSAO artifacts (@corredD, #1082)
 - Fix bumpiness artifact (#1107)
+- Stop showing `computed` bonds if `chem_comp_bond` is defined
 
 ## [v4.1.0] - 2023-03-31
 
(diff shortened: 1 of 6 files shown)


arose (changes requested):
Did you try adding logic to StructureFocusRepresentationBehavior.focus?

The lines representation should have the same check as you did for the ball-and-stick representation.

JonStargaryen:
Thanks for the pointers. How about these changes?
The focus surroundings now toggle `includeTypes` depending on the structure.

dsehnal (commented):
I am not quite sure this will have the desired effect, since chem comp bond only contains bonds within a single residue and you seem to apply this in a much wider context for all bonds. Or am I missing something?

JonStargaryen:
Maybe there's also something I don't get about `includeTypes` and `excludeTypes`. I assumed that removing `computed` from `includeTypes` would have the same effect as adding it to `excludeTypes`, but the results differ.

Adding it to `excludeTypes` will corrupt bonds between components (using the default 1tqn as example):
exclude


Removing it to `includeTypes` gives the desired visuals:
include


Original:
orig



It's also fine to keep things as is. A team member asked about these visuals but lets ignore if this does more harm than good.

arose:
> It's also fine to keep things as is. A team member asked about these visuals but lets ignore if this does more harm than good.

Yes, let's can it for now.

I think to make it work for mmcif files we at least need a way to derive polymer bonds from the sequence and mark them as not computed.

PR #1118, opened 2024-05-06
Add quantized lighting option

giagitom:
# Description
This PR adds quantized lighting at representation level. Here are some examples
1TQN (1)
1TQN (2)
1TQN (3)

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,6 @@ Note that since we don't clearly distinguish between a public and private interf
 
 ## [Unreleased]
 
-- Add representation level quantized lighting
 - Create a transformer to deflate compressed data
 - Adjust Quick Styles panel button labels
 - Improve camera interpolation code (interpolate camera rotation instead of just position)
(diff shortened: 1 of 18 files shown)


giagitom:
image

arose:
How about adding options for steps and tint? Something like in the patch below.

[quant-tweaks.diff.txt](https://github.com/molstar/molstar/files/15286790/quant-tweaks.diff.txt)

arose (changes requested):
How about adding options for steps and tint? Something like in the patch below.

[quant-tweaks.diff.txt](https://github.com/molstar/molstar/files/15286795/quant-tweaks.diff.txt)

giagitom:
Nice, I added the tweaks

giagitom:
So i think this should be seen as an intermediate between shading and flat coloring. Like a toon shading maybe? Something similar without extra options is implemented in PyMol https://wiki.pymol.org/index.php/Ray. 
image

giagitom:
Maybe just keeping quantization steps without the extra color option would be better?

arose:
image

Got a proof of concept for cel-shading. We need to do it together with the light calculation so we get light intensities for the shading levels. If you do it afterwards only from color, dark colors get overly dark.

giagitom:
It is very nice! It much better preserves original colors

arose:
Hey, made this into a PR for Cel-shading

- Add Cel-shading support
    - `celShaded` geometry parameter
    - `celSteps` renderer parameter

You can get a "sketched" looking imaged by enabling `celShaded`, enabling `outline` and setting the `threshold` to 0.1, plus setting `bumpiness` to 0.2. The below image also has a bit of `emissive` for he residues in the center.

image

arose (approved):

giagitom:
Thanks @arose for the changes. It looks very nice!

PR #1120, opened 2024-05-07
Add round caps on tubular alpha helices

giagitom:
# Description
### This PR adds round caps option on tubular helices

image

### Known issues
 - Normals at the extremities should be smoothed in some way. 
image

- Also not sure why, if caps are near gaps something goes wrong on the cap shape
image
 

## Actions


arose (changes requested):
Normals definitely look wrong. I like this options but the two issues you note should be fixed before merging.

giagitom:
@arose in last commit i tried to smooth normals. Now vertices are smoothed but colors are wrong. 
Could you check on the approach I am using? I'm not sure this is the correct way of doing this. 
 
image

giagitom:
> Also not sure why, if caps are near gaps something goes wrong on the cap shape

Should be fixed now with [d2192d6](https://github.com/molstar/molstar/pull/1120/commits/d2192d609aa36e4c1c27660787b6937d63d2bdd8) (tested with PDB ID 2XUT)
image

giagitom:
@arose everything should be now fixed.

arose (approved):
thanks, lgtm

Issue #1122, opened 2024-05-09
Occlusion issue on edges 


giagitom:
Here a test state to reproduce [test.molj.txt](https://github.com/molstar/molstar/files/15260514/test.molj.txt)

![image](https://github.com/molstar/molstar/assets/20627977/ecb98878-33e8-45c8-b379-016eebcf3725)

arose:
Don't know why this happens, but the strength needs to be very high to to be apparent, so maybe a clamp is needed somewhere.

giagitom:
I'm seeing this also with default occlusion strength, it is less perceivable of course, but it seems always there when zooming in

![image](https://github.com/molstar/molstar/assets/20627977/934ac33b-4cca-4336-bc78-c6ff38e1e4b1)

corredD:
Looks like it's come from getMappedDepth and the different resoltution, plus the multisample. If I do the following change: `float sampleDepth = getDepth(offset.xy); // getMappedDepth(offset.xy, selfCoords); ` and disable the multisample effect the artifact disappear.

giagitom:
Hi @corredD, 
Thanks, In my system I am able to resolve the issue just by disabling multisample. ```getMappedDepth``` seems to work fine to me.
I have found something else though, that remains even keeping multisample disabled. Here a [test-state.molj.txt](https://github.com/molstar/molstar/files/15308909/test-state.molj.txt) to reproduce

![image](https://github.com/molstar/molstar/assets/20627977/4c6dea6e-f926-4898-8986-b85e4395b171)

giagitom:
I seems to be related to the blur kernel. When disabled it disappears

corredD:
It seems related to the near and far clipping plane values, which are used to linearize the depth value. In your example try changing the clipping value.

giagitom:
I see, it is visible in the front and back clipping planes. 

![image](https://github.com/molstar/molstar/assets/20627977/d4b0bb20-bb43-4e7a-a325-51b1a530fc6d)

And disabling far clipping plane it disappears on the back only
![image](https://github.com/molstar/molstar/assets/20627977/54b46d42-0a87-4420-9a2a-1cd39851aac2)

arose:
Thanks for investigating this. I am addressing this in #1134.

PR #1134, opened 2024-05-17
fixes for SSAO edge artifacts

arose:
Reported in #1122.

Add `reuseOcclusion` parameter to multi-sample pass. This helps with the first artifact reported. It happens because we reuse the occlusion data in the multi-sample pass for performance. We try to correct for the jitter but apparently it dos not (fully) work. As an alternative/workaround we can disable by default in screenshots.

Add `blurBias` parameter to occlusion pass. Not sure if there is an automatic way to calculate a "correct" value but 0.5 seems to work well. There was a previous issue related to that value in the synaptic gap model, @giagitom, can you check?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Note that since we don't clearly distinguish between a public and private interf
 - Fixes SSAO edge artifacts (#1122)
     - Add `reuseOcclusion` parameter to multi-sample pass
     - Add `blurBias` parameter to occlusion pass
+    - Handle near clip in ssao-blur
 
 ## [v4.2.0] - 2023-04-05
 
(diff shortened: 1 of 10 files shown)


giagitom (approved):
Thanks @arose, I have checked synaptic gap in mesoscale explorer and the classic viewer. 

The first issue is solved by disabling ```reuseOcclusion```.  I think it is a good idea to disable it by default when taking snapshots.

Regarding second one,  I am still able to detect it when clipping is high. In worst case scenario where ```clipping``` is at 99 (and ```far``` is enabled) I need ```blurBias``` to be 0.05 to completely get rid of the artifact. 

An other issue I notice on (mesoscale explorer only) are some spare non-occluded pixels that are fixed on the screen when moving the camera. (pixels position depends on screen size and ```resolutionScale```)
image

arose:
> Regarding second one, I am still able to detect it when clipping is high. In worst case scenario where clipping is at 99 (and far is enabled) I need blurBias to be 0.05 to completely get rid of the artifact.

Not sure how to handle that automatically. Since it is such an edge case and it can be manually handled, let's leave it.

> An other issue I notice on (mesoscale explorer only) are some spare non-occluded pixels that are fixed on the screen when moving the camera. (pixels position depends on screen size and resolutionScale)

That is odd, not sure it is related. Best create a new issue with a state to reproduce. Thanks.

PR #1138, opened 2024-05-20
fix RMSD minification bug

dsehnal:
# Description

Fix a bug where the result was `NaN` if the first array way longer then the 2nd.


rtviii:
You guys seem to be releasing monthly is that more or less correct? Thank you.

dsehnal:
This one has been released already, just forgot to change the tag at GH. We don't have a fixed release schedule.

PR #1140, opened 2024-05-21
await the the snapshot setting to avoid rendering while applying snap…

corredD:
# Description
Rendering a mp4 of the current snapshot animation was not waiting for the snapshot to be applied.

# Help 
I need help with making sure the  trackball controller camera animation is also render.  Should this be done in the setPartialSnapshot function like the camera transition, or there is a missing option for the canvas3d.tick call  ? 

playing in the browser
https://github.com/molstar/molstar/assets/539461/09bc3b4d-9441-4700-8541-988033c7e5b2

after exporting the animation
https://github.com/molstar/molstar/assets/539461/4c68c47a-6800-4e4a-910c-bba9f44f6a08

### DIFF
--- a/src/mol-plugin-state/animation/built-in/state-snapshots.ts
+++ b/src/mol-plugin-state/animation/built-in/state-snapshots.ts
@@ -5,7 +5,6 @@
  */
 
 import { PluginContext } from '../../../mol-plugin/context';
-// import { now } from '../../../mol-util/now';
 import { PluginStateSnapshotManager } from '../../manager/snapshots';
 import { PluginStateAnimation } from '../model';
 
--- a/src/mol-plugin-state/animation/built-in/state-snapshots.ts
+++ b/src/mol-plugin-state/animation/built-in/state-snapshots.ts
@@ -77,7 +77,7 @@ export const AnimateStateSnapshots = PluginStateAnimation.create({
             return { kind: 'skip' };
         }
 
-        setPartialSnapshot(ctx.plugin, animState.snapshots[i]);
+        await setPartialSnapshot(ctx.plugin, animState.snapshots[i]);
 
         return { kind: 'next', state: { ...animState, currentIndex: i } };
     }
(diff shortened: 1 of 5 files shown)


dsehnal:
Thanks, nice find. I actually ran into this yesterday and wasn't sure what the reason was.

To make the trackball work during the animation export, it will likely work to add `{ updateControls?: boolean }` option to `Canvas3D.tick` and `PluginAnimationLoop.tick` and call `controls.update()` in the `Canvas3D.tick` if the option is set.

corredD:
Thanks @dsehnal, it almost worked! The rock or spin are only applyed when set in the first snapshot, and then never change. Is it possible the setPartialSnapshot function doesnt apply this settings ? or the controller doesnt get update when applying the state ?

dsehnal:
If you turn off "canvas state" when storing the snapshot (this is in the "options" section in the standard mol* ui, I think it could be set globally for the mesoscale explorer instance), it should work as long as the rocking animation is on when the user clicks the "export movie"

corredD:
Thats ok if I want to rock for all snapshot, but using different mode in different snapshot will not be possible, and thats what I am trying to do. Its seems await plugin.runTask(plugin.state.data.setSnapshot(entry.snapshot.data)); doesnt change the  plugin.canvas3d?.props.trackball.animate mode. Why is this setSnapshot different from the one use in the viewport ?

corredD:
Got it, seems to behave as expected now, let me know if the changes are ok.

src/mol-plugin-state/animation/built-in/state-snapshots.ts
```diff
@@ -5,13 +5,20 @@
  */
 
 import { PluginContext } from '../../../mol-plugin/context';
+// import { now } from '../../../mol-util/now';
```
dsehnal, line 8: Remove please

dsehnal (approved):
LGTM. Can you please update the changelog too before we merge this?

PR #1152, opened 2024-06-07
enable illustrative color scheme in the tree

corredD:
# Description
Enable to switch to illustrative color scheme in the tree. Use previously applied color at entity level so we can change the color. 

Is there an easy way to access more coloring options ?

### DIFF
--- a/src/apps/mesoscale-explorer/ui/entities.tsx
+++ b/src/apps/mesoscale-explorer/ui/entities.tsx
@@ -18,7 +18,7 @@ import { CombinedColorControl } from '../../../mol-plugin-ui/controls/color';
 import { MarkerAction } from '../../../mol-util/marker-action';
 import { EveryLoci, Loci } from '../../../mol-model/loci';
 import { deepEqual } from '../../../mol-util';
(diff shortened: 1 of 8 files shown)


arose:
I like the idea of an illustrative option but I think it should be a boolean and would require a custom uniform theme. Allowing all themes has massive performance/memory implications and breaks the color boxes in the UI, your way of enabling the illustrative theme already breaks it.

corredD:
I will look at the broken color boxes, a boolean will not allow for the illustrative per chain theme.

corredD:
What about adding a generative per-chain in the list and a toggle for illustrative ? would that make more sense ?

arose:
That would be great!

corredD:
I made a boolean switch for the illustrative scheme. Doesn't seem to broke anything.

src/apps/mesoscale-explorer/ui/entities.tsx
```diff
@@ -507,8 +507,11 @@ export class GroupNode extends Node<{ filter: string }, { isCollapsed: boolean,
             const c = type === 'generate' ? groupColors[i] : value;
             update.to(entities[i]).update(old => {
                 if (old.type) {
-                    old.colorTheme.params.value = c;
-                    old.colorTheme.params.lightness = lightness;
+                    if (illustrative) {
+                        old.colorTheme = { name: 'illustrative', params: { style: { name: 'uniform', params: { value: c, lightness: lightness } } } };
```
arose, line 511: `lightness: lightness` => `lightness`

arose:
> I made a boolean switch for the illustrative scheme. Doesn't seem to broke anything.

Nice!

When coloring type is custom in a group node, the illustrative option should be hidden as it does nothing

arose:
nice, thanks

PR #1156, opened 2024-06-12
mesoscale explorer centerOnly behavior

corredD:
…'. How can we expose all the MesoFocusLociProps in the UI ? So that user can easily toggle the centerOnly and change the transition duration...

# Description
In mesoscale explorer the camera left button click was translating the camera to the selection. Instead we want to have the option between this behavior and a zoom and center behavior. I exposed the toggle as a key binding. But would prefer to have all the MesoFocusLociParams exposed in a collapsible <ParamControler> that I could add somewhere in the interface. Another option would be to use a doubleClick to trigger the centerFocus, but I havnt found a binding for it.

## Actions


arose:
lgtm, thanks

PR #1157, opened 2024-06-12
Mesoscale explorer entities selection description

corredD:
# Description
ME - entities can have description text retrieve in the file if any in the entity field or in the manifest. The description appear for the current selection in the left panel and in the tips when mouse hover in the 3d view. The markdown will use the same handling as the Snapshot description defined in the me_ui_label PR. 

## Actions


arose (changes requested):
I updated the mmcif.ts file using the script as it is generated

image
This could use a bit nicer styling. Also, when loading mmcif files from the pdb archive, it always says 'undefined'.

corredD:
I changed the styling, and add a default description for mmCIF. I also realized this PR also include a different click behavior on the tree.

src/apps/mesoscale-explorer/ui/entities.tsx
```diff
@@ -777,7 +808,34 @@ export class EntityNode extends Node<{}, { action?: 'color' | 'clip', isDisabled
         if (e.ctrlKey) {
             this.toggleSelect(e);
         } else {
-            this.center(e);
+            const d = getEntityDescription(this.plugin, this.cell);
+            if (this.cell?.obj?.data.sourceData.state.models.length !== 0) {
+                const repr = this.cell?.obj?.data.repr;
+                if (repr) {
```
corredD, line 814: When you click on a given leaf node in the right panel tree, it will center the camera in a sequential manner iterating other the instances. I found it usefull.

arose (changes requested):
Hi, I made some tweaks to the style, but I am not sure the "selection description" is in the right place and has the right name. From the behavior it seems more like what you last clicked on, not what you selected (which could be more things). Maybe a header "Focus Info" or so? Not sure.

corredD:
make sense, I like the idea of the Header Focus Info. We can change the name of the variable, maybe "current_description" or "focus_info" as the header ?

src/apps/mesoscale-explorer/data/mmcif/model.ts
```diff
@@ -186,9 +186,10 @@ const MmcifStructure = PluginStateTransform.BuiltIn({
             } else {
                 structure = Structure.create(units);
             }
-
+            // could also use _struct_ref.pdbx_db_accession to point to uniprot with _struct_ref.db_name == UNP
             const label = entities.data.pdbx_description.value(idx).join(', ') || 'model';
-            return new PSO.Molecule.Structure(structure, { label, description: `${a.description}` });
+            const description = '*Entity id* ' + entities.data.id.value(idx) + ' *src_method* ' + entities.data.src_method.value(idx) + ' *type* ' + entities.data.type.value(idx);
```
dsehnal, line 191:
```suggestion
            const description = `*Entity id* ${entities.data.id.value(idx)} *src_method* ${entities.data.src_method.value(idx)} *type* ${entities.data.type.value(idx)}`;
```

Also perhaps consider adding fallbacks with `?? 'fallback value'` as `${undefined}` renders as `undefined` which is not quite desirable.

src/apps/mesoscale-explorer/ui/entities.tsx
```diff
@@ -229,6 +242,46 @@ export class SelectionInfo extends PluginUIComponent<{}, { isDisabled: boolean }
...
+        this.subscribe(this.plugin.state.data.behaviors.isUpdating, v => {
+            this.setState({ isDisabled: v });
+        });
+
+        this.subscribe(this.plugin.managers.structure.selection.events.changed, e => {
+            if (!this.state.isDisabled) {
+                this.forceUpdate();
+            }
+        });
+    }
```
dsehnal, line 261: The `isDisabled` state is not used anywhere.

```suggestion
    componentDidMount() {
        this.subscribe(combineLatest([
          this.plugin.state.data.behaviors.isUpdating,
          this.plugin.managers.structure.selection.events.changed
        ]), ([isUpdating]) => {
          if (!isUpdating) this.forceUpdate();
        });
    }
```

src/apps/mesoscale-explorer/ui/entities.tsx
```diff
@@ -229,6 +242,46 @@ export class SelectionInfo extends PluginUIComponent<{}, { isDisabled: boolean }
...
+
+    get info() {
+        // const infos: { label: string, key: string, description?: string }[] = [];
```
dsehnal, line 264:
```suggestion
```

src/apps/mesoscale-explorer/ui/entities.tsx
```diff
@@ -229,6 +242,46 @@ export class SelectionInfo extends PluginUIComponent<{}, { isDisabled: boolean }
...
+        const info: { focusInfo: string } = { focusInfo: '' };
```
dsehnal, line 265: Return just the `focusInfo` string since that is the only property?

corredD: Thanks @dsehnal I simplified the function in the last commit. The component is at the bottom right panel, if you think there is a better place for it let me know. I like it here as it help isolated the entity in the tree.

arose (approved):

Issue #1168, opened 2024-06-23
celShading issue using material 1 and roughness 0


giagitom:
I have found an issue on celShading in this configuaration: 

```
metalness: 1
roughness: 0
```

![image](https://github.com/molstar/molstar/assets/20627977/f03766f8-43cb-40f9-acbb-9f4540c105f1)

PR #1170, opened 2024-06-23
cel-shading artifacts

arose:
- clamp metalness & roughness to look better

fix for #1168


giagitom (approved):
Thanks! It is working as expected now

PR #1180, opened 2024-06-27
doNotDisposeCanvas3DContext option in PluginContext.dispose

dsehnal:
# Description

Allow sharing single canvas3d context with multiple plugins that need to be disposed.

## Actions


alisterburt:
Wonderful, thanks for the quick response!

Issue #1188, opened 2024-07-03
adjustCylinderLength affects aromatic links


bergwerf:
PubChem CID 2244:

| Default ball and stick representation settings | adjustCylinderLength = true |
|----|----|
| ![image](https://github.com/molstar/molstar/assets/4458965/be57b9f7-a52b-4b93-aad8-0cc1026fd597) | ![image](https://github.com/molstar/molstar/assets/4458965/4463c638-b08f-458f-a35b-b4db8cf364fb)

I think the dashed aromatic bonds should still stretch out fully. It does makes sense to adjust the bond cylinders to center the separation between the colors of both atoms.

arose:
I agree, care to do a PR to make `adjustCylinderLength` not affect aromatic offset bonds?

bergwerf:
I will give it a try. I have a bit of a hard time following the code. The BallAndStickRepresentationProvider introduces IntraUnitBondCylinderVisual which flags some bonds as LinkStyle.Aromatic. There is createIntraUnitBondCylinderMesh, which creates builder props that contain the position which takes adjustCylinderLength into account. Ultimately createLinkCylinderMesh is where the fix needs to happen.
- What is the difference between referencePosition and position?
- Should we add a boolean to position which indicates if it should be adjusted?

bergwerf:
Something needs to happen at https://github.com/molstar/molstar/blob/354438052e5161b18104d12070bbfb6d441a7357/src/mol-repr/structure/visual/util/link.ts#L432. I think vm is supposed to be the midpoint, but it is used as endpoint for addFixedCountDashes.

PR #1191, opened 2024-07-07
Add unadjusted position for aromatic link dashes

bergwerf:
# Description

Address #1188.

@arose This is not the prettiest solution because it adds an ad-hoc parameter to LinkBuilderProps, but it does work. If you are happy with it I will check off the points below.

## Actions

### DIFF
--- a/src/mol-repr/structure/visual/util/link.ts
+++ b/src/mol-repr/structure/visual/util/link.ts
@@ -80,7 +80,7 @@ export function calculateShiftDir(out: Vec3, v1: Vec3, v2: Vec3, v3: Vec3 | null
 
 export interface LinkBuilderProps {
     linkCount: number
-    position: (posA: Vec3, posB: Vec3, edgeIndex: number) => void
+    position: (posA: Vec3, posB: Vec3, edgeIndex: number, unadjusted?: boolean) => void
     radius: (edgeIndex: number) => number,
     referencePosition?: (edgeIndex: number) => Vec3 | null
     style?: (edgeIndex: number) => LinkStyle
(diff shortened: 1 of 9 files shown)


src/mol-repr/structure/visual/util/link.ts
```diff
@@ -80,7 +80,7 @@ export function calculateShiftDir(out: Vec3, v1: Vec3, v2: Vec3, v3: Vec3 | null
 
 export interface LinkBuilderProps {
     linkCount: number
-    position: (posA: Vec3, posB: Vec3, edgeIndex: number) => void
+    position: (posA: Vec3, posB: Vec3, edgeIndex: number, unadjusted?: boolean) => void
```
dsehnal, line 83: I think having this as optional param could cause compiler de-optimization for this function since the signature could change between calls. Could you make this a mandatory unadjusted: boolean and be explicit with the true/false in the function calls?

bergwerf: I thought this would impact a lot of code, but it doesn't because most existing builder props can just ignore this extra argument. I'll push an update changing this into a positive verb `adjust`.

bergwerf:
@dsehnal I made the adjust argument required.

src/mol-repr/structure/visual/util/link.ts
```diff
@@ -80,7 +81,7 @@ export function calculateShiftDir(out: Vec3, v1: Vec3, v2: Vec3, v3: Vec3 | null
 
 export interface LinkBuilderProps {
     linkCount: number
-    position: (posA: Vec3, posB: Vec3, edgeIndex: number) => void
+    position: (posA: Vec3, posB: Vec3, edgeIndex: number, adjust: boolean) => void
```
dsehnal, line 84: This function is used in quite a few files. Does your current solution compile?

image

bergwerf: For me `npm run build` succeeds and the viewer works. In JavaScript all those instances are allowed to ignore trailing arguments, making this a fairly lightweight change.

bergwerf:
It looks like my test file doesn't conform to the style format here, I'll fix that.

arose (approved):
thanks @bergwerf, looking good!

committed some tweaks in the interest of time

PR #1193, opened 2024-07-08
Make Trackball spin in radians per second

bergwerf:
# Description

The trackball spin animation is carried out via the same mechanism as pointer drag rotation, which takes the canvas aspect ratio into account via `getRotateFactor`. It might make more sense to set a spin animation to a constant speed in radians per second. The least invasive way to do this is to compensate for `getRotateFactor` when the spin animation is applied.

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Note that since we don't clearly distinguish between a public and private interf
 - Fix DoF with pixel ratios =! 1
 - Fix DoF missing transparent depth
 - Fix trackball pinch zoom and add pan
+- Change trackball animate spin speed unit to radians per second
 
 ## [v4.4.1] - 2023-06-30
 
--- a/src/mol-canvas3d/controls/trackball.ts
+++ b/src/mol-canvas3d/controls/trackball.ts
@@ -67,7 +67,7 @@ export const TrackballControlsParams = {
     animate: PD.MappedStatic('off', {
         off: PD.EmptyGroup(),
         spin: PD.Group({
-            speed: PD.Numeric(1, { min: -20, max: 20, step: 1 }),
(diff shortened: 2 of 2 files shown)


dsehnal:
> The trackball spin animation is carried out via the same mechanism as pointer drag rotation, which takes the canvas aspect ratio into account via getRotateFactor. It might make more sense to set a spin animation to a constant speed in radians per second. The least invasive way to do this is to compensate for getRotateFactor when the spin animation is applied.

I am ok with changing this to `rad/s` instead of scaling this to "60 fps" as it is now, but I don't see reasoning behind the `getRotateFactor`, since that is there because it is tied to user input, but the "spin" isn't. 

@arose what do you think?

bergwerf:
In the current implementation the spin animation gets multiplied by a perhaps unintended `getRotateFactor`, so if the viewport is very elongated the spin animation is several times slower. This change counteracts that by dividing that factor out. When the spin animation is applied, it adds a delta-x to `_rotCurr`. When the camera is updated, the difference between `_rotCurr` and `_rotPrev` are translated into a rotation. This mechanism was presumably first developed for the user dragging a pointer. Instead of making the code more complex with a separate mechanism for absolute rotation (e.g. in terms of an axis and an angle, instead of two pointer positions), this change just divides out the `getRotateFactor`.

To me changing into radians per second is not essential, but it would be nice if the rotation speed is constant no matter the aspect ratio of the viewport.

bergwerf:
This is the relevant line: https://github.com/molstar/molstar/blob/88e7c4c640059d29aa88217922bfc57c6b7dcd79/src/mol-canvas3d/controls/trackball.ts#L212

dsehnal:
Right, I missed that. Makes sense.

dsehnal:
@bergwerf Can you also add the changelog note, etc?

bergwerf:
@dsehnal Yes done.

dsehnal (approved):

PR #1207, opened 2024-07-17
Separed postprocessing passes

giagitom:
# Description
Split postprocessing passes in different files to simplify postprocessing file structure.

## Actions


arose (changes requested):
Thanks for doing this!

Played around with it to see if it all still works and found changing the occlusion resolution parameter around a few times (lower then higher) does work correctly and will eventually throw an error.

giagitom:
>Played around with it to see if it all still works and found changing the occlusion resolution parameter around a few times (lower then higher) does work correctly and will eventually throw an error.

@arose do you have a test state to reproduce? I tried extensively modifying every occlusion parameter but haven't found errors so far.

arose:
Cant be shown with a state. I get it by moving the resolutionScale param around

image

giagitom:
So I wasn't seeing anything because I was using viewport settings. I'm not sure why but it seems that using viewport settings it is also triggered the resize of passes. Instead using the other settings resize is not triggered. Maybe this is causing the error? 

Any Idea on what is going on?

image

giagitom:
Fixed the issue causing errors with last commit

giagitom:
@arose I resolved conflicts due to #1211

It should be ready now

giagitom:
Is it correct to use ```nearest``` here?

https://github.com/molstar/molstar/blob/18b1492d54ba3e2beee47bda6ac262ef8b8881c7/src/mol-canvas3d/passes/postprocessing.ts#L142-L144

arose:
should be 'linear' if available, but actually that schema info it is not used.

Issue #896, opened 2023-08-22
Coloring by Model-index not working?


giagitom:
Maybe I have found an issue with model index coloring. 
I tried using it with a random multi state structure (PDB Id 1G03) and it seems to color each state in the same way. is this an expected behavior? 

Here is a [test state](https://github.com/molstar/molstar/files/12410809/test.molj.txt)

arose:
model-index colors by models in the state tree. You probable want trajectory-index.

giagitom:
Hi @arose, I tried your suggestion but it seems not to work. Every model is shown as black using trajectory-index

![image](https://github.com/molstar/molstar/assets/20627977/35a65e0c-973f-4d01-a373-2c1522e22723)

giagitom:
Oh.. I have found that when I load all models together in my implementation, trajectory index is working as expected. 

I have also discovered that utilizing the entity index yields identical outcomes as the trajectory index. However, shouldn't all entities be colored uniformly instead?

![image](https://github.com/molstar/molstar/assets/20627977/4b8e7169-01d0-4cf4-91d2-a2e688e883df)

arose:
something is of but I don't time to look into it at the moment

PR #1209, opened 2024-07-20
ensure color-theme gets correctly updated

arose:
- fixes trajectory-index color-theme (#896)

https://github.com/user-attachments/assets/b10b3522-a9bf-4da7-bc12-eb1f50e5f6e8


src/mol-repr/structure/complex-visual.ts
```diff
@@ -157,9 +161,13 @@ export function ComplexVisual<G extends Geometry, P extends StructureParams & Ge
                 throw new Error('expected renderObject to be available');
             }
 
+            if (updateState.updateColor || updateState.updateSize || updateState.updateTransform) {
+                // console.log('update locationIterator');
```
dsehnal, line 165:
```suggestion
```

And the others too?

dsehnal (approved):

src/mol-repr/structure/complex-visual.ts
arose: I often use those log statements so I like to keep them around (commented out since we don't have good way for verbose logging).

giagitom (commented):
Thanks @arose, 

I'm still having an issue when updating sphere visual only. This isn't due to the changes in the current PR since I can also see the issue in the web app. Let me know if you want this as a separate issue.

The issue is the following:
I have made a custom element-symbol theme having the ability to color carbon alpha with trajectory-index color scheme (this is currently not available on molstar app)

In this particular case I noticed that when changing model index, spheres visual color is never updated.

This is a test state I have made that presents the issue. You just need to change the model index to notice it.

[test.molx.txt](https://github.com/user-attachments/files/16334665/test.molx.txt)

image

arose:
The state works fine with this branch for me.

image


Note that in your own color theme you will need to update the contextHash when the theme depends on a custom property

giagitom:
Sorry that was the wrong example. Plain trajectory index coloring is working fine on the new PR. 
The problem arises for example when I try to integrate trajectory index for carbon color in element-symbol color scheme.

To obtain something like this
image

I tested PDB 1G03 with the new PR but changing model index in this case I'm not seeing a correct update. On my implementation spheres are not updated, on molstar app instead both spheres and cylinders are not updated (I'm not sure why there is a difference). To trigger the update, I need to change other color parameters like the palette for example.
 




To integrate trajectory index carbon color on element-symbol I just did this: 


image

arose:
should all work now, added trajectory-index in a few more places and handle missing contextHash

giagitom (approved):
Thanks @arose 
Everything is working fine now!

giagitom:
One last thing I observed was this https://github.com/molstar/molstar/issues/896#issuecomment-1689551377
When using ```entity-id``` coloring in structures like 1G03 and showing all states together, all entities (that should be equal) are assigned with a different color.

arose:
> One last thing I observed was this [#896 (comment)](https://github.com/molstar/molstar/issues/896#issuecomment-1689551377) When using `entity-id` coloring in structures like 1G03 and showing all states together, all entities (that should be equal) are assigned with a different color.

Looks OK to me. How are you loading them? If is still an issue lets do another PR.

image

giagitom:
Yes, I'm still experiencing this.

I am using this to load all models together

```js
Structure.ofTrajectory(trajectory, ctx)
```

PR #1213, opened 2024-07-24
Pass current colorTheme props to theme provider

hui2000ji:
# Description

I am writing a colorTheme which allows the user to customize the underlying parameters of the CustomStructurePropertyProvider. I found that current implementations do not allow `props` to be passed from colorTheme to CustomStructurePropertyProvider, so I did the following modifications:

```typescript
    export async function ensureDependencies(ctx: CustomProperty.Context, theme: ThemeRegistryContext, data: ThemeDataContext, props: Props) {
        // old: await theme.colorThemeRegistry.get(props.colorTheme.name).ensureCustomProperties?.attach(ctx, data);
        // old: await theme.sizeThemeRegistry.get(props.sizeTheme.name).ensureCustomProperties?.attach(ctx, data);
        await theme.colorThemeRegistry.get(props.colorTheme.name).ensureCustomProperties?.attach(ctx, data, props.colorTheme);
        await theme.sizeThemeRegistry.get(props.sizeTheme.name).ensureCustomProperties?.attach(ctx, data, props.sizeTheme);
    }
```

This allows for more detailed control over property computation, for example:

```typescript
import { AccessibleSurfaceAreaProvider, AccessibleSurfaceAreaParams } from '../accessible-surface-area';
export const AccessibleSurfaceAreaColorThemeParams = {
    list: PD.ColorList('rainbow', { presetKind: 'scale' }),
    ...AccessibleSurfaceAreaParams // now user can modify AccessibleSurfaceAreaParams in color theme setting!
};

export const AccessibleSurfaceAreaColorThemeProvider: ColorTheme.Provider<AccessibleSurfaceAreaColorThemeParams, 'accessible-surface-area'> = {
    ...
    ensureCustomProperties: {
        attach: (ctx: CustomProperty.Context, data: ThemeDataContext, props?: Partial<PD.Values<AccessibleSurfaceAreaColorThemeParams>>) => {
            return data.structure ? AccessibleSurfaceAreaProvider.attach(ctx, data.structure, props, true) : Promise.resolve()
        },
        detach: (data) => data.structure && AccessibleSurfaceAreaProvider.ref(data.structure, false)
    }
};
```

## Actions

### DIFF
--- a/src/mol-theme/theme.ts
+++ b/src/mol-theme/theme.ts
@@ -54,8 +54,8 @@ namespace Theme {
     }
 
     export async function ensureDependencies(ctx: CustomProperty.Context, theme: ThemeRegistryContext, data: ThemeDataContext, props: Props) {
-        await theme.colorThemeRegistry.get(props.colorTheme.name).ensureCustomProperties?.attach(ctx, data);
-        await theme.sizeThemeRegistry.get(props.sizeTheme.name).ensureCustomProperties?.attach(ctx, data);
+        await theme.colorThemeRegistry.get(props.colorTheme.name).ensureCustomProperties?.attach(ctx, data, props.colorTheme);
+        await theme.sizeThemeRegistry.get(props.sizeTheme.name).ensureCustomProperties?.attach(ctx, data, props.sizeTheme);
     }
 
     export function releaseDependencies(theme: ThemeRegistryContext, data: ThemeDataContext, props: Props) {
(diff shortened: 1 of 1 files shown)


arose:
The custom properties should be updated separately from the theme. The theme must define a contexHash that uses the custom property version. If that is done updating the custom property will trigger coloring to be updated. This is how it is done for AccessibleSurfaceArea and others.

hui2000ji:
Sorry, I don't see the downside for this change yet. This modification simply allows users to change the parameters of the CustomStructureProperty calculation, e.g. `ShrakeRupleyComputationParams`, in the color theme params, which IMHO would be very convenient for users. Could you elaborate further on why it's not recommended do this?

I am writing a color theme to visualize the antibody CDR regions, and the visualization depends on the CDR definition (e.g. Kabat, Chothia, or IMGT) which can be customized by the user. In this case, what is the recommended implementation?

Thanks a lot in advance!

PR #1220, opened 2024-08-03
Avoid calculating bonds for water units when `ignoreHydrogens` is on

arose:
Avoid calculating bonds for water units when `ignoreHydrogens` is on

- Add `Water` trait to `Unit`
- Remove unsued code from `Structure.ofModel`
- Try reuse boundary in element-cross visual

### DIFF
--- a/src/mol-repr/structure/visual/bond-inter-unit-cylinder.ts
+++ b/src/mol-repr/structure/visual/bond-inter-unit-cylinder.ts
@@ -16,12 +16,13 @@ import { createLinkCylinderImpostors, createLinkCylinderMesh, LinkBuilderProps, 
 import { ComplexMeshParams, ComplexVisual, ComplexMeshVisual, ComplexCylindersParams, ComplexCylindersVisual } from '../complex-visual';
 import { VisualUpdateState } from '../../util';
 import { BondType } from '../../../mol-model/structure/model/types';
-import { BondCylinderParams, BondIterator, getInterBondLoci, eachInterBond, makeInterBondIgnoreTest } from './util/bond';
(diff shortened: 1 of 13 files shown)


src/mol-repr/structure/visual/bond-inter-unit-cylinder.ts
```diff
@@ -246,7 +253,7 @@ export function InterUnitBondCylinderImpostorVisual(materialId: number): Complex
                 state.updateColor = true;
             }
 
-            if (newStructure.interUnitBonds !== currentStructure.interUnitBonds) {
+            if (hasStructureVisibleBonds(newStructure, newProps) && newStructure.interUnitBonds !== currentStructure.interUnitBonds) {
```
dsehnal, line 256: Should the 1st condition be `hasStructureVisibleBonds(newStructure, newProps) !== hasStructureVisibleBonds(currentStructure, currentProps)`?

src/mol-repr/structure/visual/element-cross.ts
```diff
@@ -68,10 +76,18 @@ export function createElementCross(ctx: VisualContext, unit: Unit, structure: St
         builder.add(s[0], s[1], s[2], e[0], e[1], e[2], i);
     }
 
+    const oldBoundingSphere = lines ? Sphere3D.clone(lines.boundingSphere) : undefined;
```
dsehnal, line 79: Move below the early return?

arose: done

src/mol-repr/structure/visual/bond-inter-unit-cylinder.ts
arose: no, not needed

Issue #1221, opened 2024-08-03
When using ```entity-id``` coloring in multi-model structures all entities are assigned with a different color.


arose:
When using ```entity-id``` coloring in structures like 1G03 and showing all states together, all entities (that should be equal) are assigned with a different color.

I am using this to load all models together

```Structure.ofTrajectory(trajectory, ctx)```

_Originally posted by @giagitom in https://github.com/molstar/molstar/issues/1209#issuecomment-2254550394_

PR #1222, opened 2024-08-03
Improve entity-id coloring for structures with multiple models from the same source

arose:
Improve entity-id coloring for structures with multiple models from the same source, #1221

### DIFF
--- a/src/mol-theme/color/entity-id.ts
+++ b/src/mol-theme/color/entity-id.ts
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2021-2022 mol* contributors, licensed under MIT, See LICENSE file for more info.
+ * Copyright (c) 2021-2024 mol* contributors, licensed under MIT, See LICENSE file for more info.
  *
  * @author Alexander Rose <alexander.rose@weirdbyte.de>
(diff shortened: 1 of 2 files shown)


giagitom (approved):
Thanks @arose! It is working fine now

src/mol-theme/color/entity-id.ts
```diff
@@ -84,15 +97,15 @@ export function EntityIdColorTheme(ctx: ThemeDataContext, props: PD.Values<Entit
             let serial: number | undefined = undefined;
             if (StructureElement.Location.is(location)) {
                 const entityId = getEntityId(location);
-                const modelIndex = location.structure.models.indexOf(location.unit.model);
-                const k = key(entityId, modelIndex);
+                const sourceSerial = sourceSerialMap.get(location.unit.model.sourceData) ?? -1;
```
dsehnal, line 100: Should this fallback to `location.structure.models.indexOf(location.unit.model)` instead of the `-1`?

arose: no, if it is not found there is some error and its better to show the DefaultColor.

PR #1227, opened 2024-08-16
allow nested components

hui2000ji:
## Before change
Currently, we can create component within a component (i.e. a nested component) by running "Apply Action > Selection" on a component, but the child component does not appear in the Components tab on the right, as demonstrated by the screenshot below ("chain C" is not in the Components tab).

molstar org_viewer_

I found that this is due to the StructureComponentRef for all nested components are never created, due to the following check when building the structure hierarchy:

```typescript
// src/mol-plugin-state/manager/structure/hierarchy-state.ts
[
    // Component check function (type: TestCell)
    (cell, state) => {
        if (state.currentComponent || !state.currentStructure || cell.transform.transformer.definition.isDecorator)
            return false; // NOTE: if `currentComponent` is not null (as in the case of nested components), will not call the apply function
        return SO.Molecule.Structure.is(cell.obj);
    },
    // Component apply function (type: ApplyRef)
    (state, cell) => {
        if (state.currentStructure) {
            state.currentComponent = createOrUpdateRefList(state, cell, state.currentStructure.components, StructureComponentRef, cell, state.currentStructure);
        }
    },
    // Component cleanup function (type: LeaveRef)
    state => state.currentComponent = void 0
]
```

When a nested component is being processed, because `state.currentComponent` is already set to its parent component (which has been processed earlier), its StructureComponentRef will not be created, and its representations (which will be processed soon) will be linked to its parent's StructureComponentRef rather than its.

## Changes

- Added an attribute `parentComponents` to the BuildState, which serves as a stack when building the structure hierarchy.
- Whenever a component is being processed (i.e. in the apply function/ApplyRef), if `currentComponent` is already set (to the parent of the REAL current component), push it to the stack.
- Whenever a component finished being processed (i.e. in the cleanup function/LeaveRef), restore the `currentComponent` to the parent of the current component by popping from the stack.

Now we could see that the nested components show up in the components tab, and can be correctly focused and selected!

molstar_build_viewer_index html

## Example use case for nested component

- Polymer
    - surface
        - epitope
    - core
- Ligand
- Water

This is much clearer than:

- Polymer
- Ligand
- Water
- surface
- epitope
- core

## Actions


dsehnal:
Thanks for the contribution.

You implementation is quite straightforward, but the maximum one level of nesting was a design decision I made when originally writing this code. Back then I had a reason for it... let me think a bit about if this doesn't break some assumption downstream.

I think it should be fine and since it's optional, it should not break anything in the standard way the state tree is used.

dsehnal (commented):
Please also edit the header of the `hierarchy-state.ts` file with your contribution and add a note to `CHANGELOG.md`

hui2000ji:
Done. Please review the changes, thank you.

dsehnal:
@hui2000ji sorry this fell out of the radar for me. Getting this in now.

PR #1228, opened 2024-08-16
Support for AlphaFold DB BinaryCIF files and 4-character PDB-Dev IDs

JonStargaryen:
# Description
- add option to fetch BinaryCIF from AlphaFold DB and make that the default
  - I saw @midlik starting to use these files, so I'm assuming that AlphaFold DB finally fully supports BinaryCIF
- PDB-Dev now uses PDB-style identifiers alongside the old style (e.g., PDBDEV_00000012 = 8ZZC)
  - future depositions will only have PDB-style identifiers but not the old PDBDEV_00000012 style
  - this PR ensures support for both styles
  - also note that PDB-Dev currently fails to serve data via the old identifiers but this will be fixed by next Wednesday (in the meantime, [pdb-dev-beta.wwpdb.org](https://pdb-dev-beta.wwpdb.org/bcif/PDBDEV_00000001.bcif) works for testing)

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ Note that since we don't clearly distinguish between a public and private interf
 - Avoid calculating bonds for water units when `ignoreHydrogens` is on
 - Add `Water` trait to `Unit`
 - Improve entity-id coloring for structures with multiple models from the same source (#1221)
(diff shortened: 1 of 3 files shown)


arose (approved):
lgtm, thanks

PR #1233, opened 2024-08-21
ModelServer & VolumeServer: add health-check

JonStargaryen:
# Description
- adds a new endpoint to ModelServer & VolumeServer to report service health
- useful to monitor API status, a non-200 status code makes this easier when used e.g. in Kubernetes
- adds new optional `healthCheckPath` config prop, which can point to 1...n files to ensure that these are readable (i.e. asserting that e.g. both X-ray and EM data are available, depending on the setup)

## Actions

### DIFF
--- a/src/servers/volume/server/web-api.ts
+++ b/src/servers/volume/server/web-api.ts
@@ -1,13 +1,15 @@
 /**
- * Copyright (c) 2018-2019 mol* contributors, licensed under MIT, See LICENSE file for more info.
+ * Copyright (c) 2018-2024 mol* contributors, licensed under MIT, See LICENSE file for more info.
  *
  * Taken/adapted from DensityServer (https://github.com/dsehnal/DensityServer)
  *
  * @author David Sehnal <david.sehnal@gmail.com>
  * @author Alexander Rose <alexander.rose@weirdbyte.de>
+ * @author Sebastian Bittrich <sebastian.bittrich@rcsb.org>
  */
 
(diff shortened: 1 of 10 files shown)


JonStargaryen:
Will add the same for ModelServer if you deem this useful.

src/servers/volume/server/web-api.ts
```diff
@@ -206,4 +211,33 @@ function queryDone() {
...
+            healthCheckResponse(res, false, 'Failed to access data from file system.');
+            return;
+        }
+    }
+    healthCheckResponse(res, true);
+}
+
+function healthCheckResponse(res: express.Response, success: boolean, msg?: string) {
+    res.writeHead(success ? 200 : 500, {
+        'Content-Type': 'text/plain',
```
dsehnal, line 237: Should this be `application/json` instead?

dsehnal (approved):

src/servers/volume/server/web-api.ts
JonStargaryen: `text/plain` tends to be easier to work with, at least for the monitoring tools that RCSB PDB uses (that match with an expected string indicating a healthy status).

Could wrap everything in an object and deliver as JSON. A simple string and `text/plain` should be more appropriate for our use-case.

JonStargaryen:
Moved the health-check impl to the common/util.ts file and enabled the same behavior for ModelServer.

PR #1238, opened 2024-08-24
Error messages in UI for MVSJ file referencing unavailable annotations

aliaksei-chareshneu:
# Description
- Adds UI toasts and records in the log at the bottom with explicit error messages appearing when the user tries to open a `.mvsj` file referencing unavailable annotations
- Makes the error more explicit to the user

## Actions

I would kindly ask @midlik to review this.

### DIFF
--- a/src/mol-model-props/common/custom-property.ts
+++ b/src/mol-model-props/common/custom-property.ts
@@ -10,6 +10,7 @@ import { ParamDefinition as PD } from '../../mol-util/param-definition';
 import { ValueBox } from '../../mol-util';
 import { OrderedMap } from 'immutable';
 import { AssetManager, Asset } from '../../mol-util/assets';
+import { PluginContext } from '../../mol-plugin/context';
 
 export { CustomProperty };
 
@@ -34,7 +35,7 @@ namespace CustomProperty {
         readonly getParams: (data: Data) => Params
         readonly defaultParams: Params
         readonly isApplicable: (data: Data) => boolean
-        readonly attach: (ctx: Context, data: Data, props?: Partial<PD.Values<Params>>, addRef?: boolean) => Promise<void>
+        readonly attach: (ctx: Context, data: Data, props?: Partial<PD.Values<Params>>, addRef?: boolean, plugin?: PluginContext) => Promise<void>
         readonly ref: (data: Data, add: boolean) => void
         readonly get: (data: Data) => ValueBox<Value | undefined>
         readonly set: (data: Data, props: PD.Values<Params>, value?: Value) => void
--- a/src/mol-model-props/common/custom-property.ts
+++ b/src/mol-model-props/common/custom-property.ts
@@ -2,7 +2,6 @@
  * Copyright (c) 2019-2020 mol* contributors, licensed under MIT, See LICENSE file for more info.
  *
  * @author Alexander Rose <alexander.rose@weirdbyte.de>
- * @author Aliaksei Chareshneu <chareshneu.tech@gmail.com>
  */
 
 import { RuntimeContext } from '../../mol-task';
@@ -11,7 +10,6 @@ import { ParamDefinition as PD } from '../../mol-util/param-definition';
 import { ValueBox } from '../../mol-util';
 import { OrderedMap } from 'immutable';
 import { AssetManager, Asset } from '../../mol-util/assets';
-import { PluginContext } from '../../mol-plugin/context';
 
 export { CustomProperty };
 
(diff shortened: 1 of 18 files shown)


src/mol-model-props/common/custom-property.ts
```diff
@@ -34,7 +36,7 @@ namespace CustomProperty {
         readonly getParams: (data: Data) => Params
         readonly defaultParams: Params
         readonly isApplicable: (data: Data) => boolean
-        readonly attach: (ctx: Context, data: Data, props?: Partial<PD.Values<Params>>, addRef?: boolean) => Promise<void>
+        readonly attach: (ctx: Context, data: Data, props?: Partial<PD.Values<Params>>, addRef?: boolean, plugin?: PluginContext) => Promise<void>
```
midlik, line 39: Instead of adding another parameter, I would include PluginContext in Context. 
So we would have:
```
    export interface Context {
        runtime: RuntimeContext
        assetManager: AssetManager
        plugin?: PluginContext
    }
    ...
    readonly attach: (ctx: Context, data: Data, props?: Partial<PD.Values<Params>>, addRef?: boolean) => Promise<void>
```
This applies to both `attach` and `obtain` functions. I think in this way we will keep the code more tidy.

@arose / @dsehnal  please confirm if what I'm saying makes sense.

midlik (commented):
Thank you, Aliaksei. This will be very helpful to the users.
I left just one comment regarding custom property functions signatures.

src/mol-model-props/common/custom-property.ts
dsehnal: Let's remove parameter altogether, suggesting a different solution.

dsehnal:
I don't like passing around the `PluginContext` as it introduces coupling between the `mol-` sub-components that should not be there.

Instead, I suggest this solution similar to how React handles hooks:

- Add `class _MVSErrorContext` with `add`, `reset`, and `report(ctx: PluginContext)` methods; and `export const MVSErrorContext` singleton instance to `mvs/helpers/errors.ts`
- Instead of passing the `PluginContext` around, use the singleton `MVSErrorContext.add` to remember the errors.
- Modify the `loadMVS` method to:

```ts
MVSErrorContext.reset();
try {
   // existing code
} catch {
   // existing code
} finally {
  if (!options.doNotReportErrors) { // new parameter in options
    MVSErrorContext.report(plugin);
  }
  MVSErrorContext.reset();
}
```

----------

What do you think @midlik and @aliaksei-chareshneu  ?

aliaksei-chareshneu:
> I don't like passing around the `PluginContext` as it introduces coupling between the `mol-` sub-components that should not be there.
> 
> Instead, I suggest this solution similar to how React handles hooks:
> 
> * Add `class _MVSErrorContext` with `add`, `reset`, and `report(ctx: PluginContext)` methods; and `export const MVSErrorContext` singleton instance to `mvs/helpers/errors.ts`
> * Instead of passing the `PluginContext` around, use the singleton `MVSErrorContext.add` to remember the errors.
> * Modify the `loadMVS` method to:
> 
> ```ts
> MVSErrorContext.reset();
> try {
>    // existing code
> } catch {
>    // existing code
> } finally {
>   if (!options.doNotReportErrors) { // new parameter in options
>     MVSErrorContext.report(plugin);
>   }
>   MVSErrorContext.reset();
> }
> ```
> 
> What do you think @midlik and @aliaksei-chareshneu ?

I do agree

aliaksei-chareshneu:
@dsehnal, @midlik, I have implmented the David's feedback. Could you review it?

midlik:
Hi @dsehnal , I'm a bit worried about how this singleton pattern will work with async functions.
If two functions run simultaneously, their calls to the singleton `MVSErrorContext` can interfere.

(Sorry I'm writing this now, after @aliaksei-chareshneu  started implementing it, but I just had this thought now.)

midlik:
Would it make sense to add `ErrorContext` to the `CustomProperty.Context`?

```
    export interface Context {
        runtime: RuntimeContext,
        assetManager: AssetManager,
        errors?: ErrorContext,
    }
    ...
    readonly attach: (ctx: Context, data: Data, props?: Partial<PD.Values<Params>>, addRef?: boolean) => Promise<void>{
        ...
        ctx.errors?.add('Something went wrong');
        ...
    }
```

dsehnal (approved):

dsehnal:
> Hi @dsehnal , I'm a bit worried about how this singleton pattern will work with async functions.
> If two functions run simultaneously, their calls to the singleton `MVSErrorContext` can interfere.

Well, this would break if the user tried to load multiple MVS states at the same time in different plugins.

> Would it make sense to add ErrorContext to the CustomProperty.Context? 

This is a good idea actually. I will try a separate PR with this approach.

@aliaksei-chareshneu I like the current solution as well, but let's try more generic approach @midlik is suggesting first. I will make a separate PR to see how it would work out.

dsehnal:
@midlik @aliaksei-chareshneu See https://github.com/molstar/molstar/pull/1254

dsehnal:
Closing this in favor for #1254. Thanks @midlik and @aliaksei-chareshneu for the discussion that led to that solution.

Issue #1232, opened 2024-08-19
Mol2 file status_bit display error.


QDMarkMan:
Hi, here is my  issue description:

### 1. Bug Description
If my MOL2 file contains status_bit in atoms or bonds, it will render an incorrect structure, as shown below:

**Expected Behavior:**
DIsplay the right version from mol2 file. 

**Actual Behavior:**
![image](https://github.com/user-attachments/assets/3050a83a-3eb5-4062-8c40-3b0843dd9e49)
![image](https://github.com/user-attachments/assets/adda9f96-d5a8-4096-b5fe-ff5446e9b5b6)

### 2. Environment
- molstar: [The latest version]

### 3. Possible Solution

After checked the code, I found teh code below maybe cause a mol2 parse error:
![image](https://github.com/user-attachments/assets/14c09b11-b89c-4ef3-8da5-1b6ca947f85f)

arose:
Thanks for reporting this. Could you try to do a PR with a fix?

QDMarkMan:
> Thanks for reporting this. Could you try to do a PR with a fix?

Thanks for your replay! I’ll try to submit a PR later.

PR #1251, opened 2024-08-30
Fix mol2 status_bit read error

QDMarkMan:
# Description
- ~~Add a `readLineElements ` for get column count.~~
- fix Mol2 file status_bit display error. [#1232](https://github.com/molstar/molstar/issues/1232)

## Actions


dsehnal (commented):
Hi, thanks for the contribution.

By doing the `dataLines: data.split(/\r?\n/),` in the constructor, your implementation beats the purpose of the "streaming tokenizer". This change would make tasks like parsing large CIF files significantly slower due to the extra string allocations and running the regex on the entire file.

I would suggest an alternate implementation with getting rid of the `columnCount` field and `for (let j = 0; j < _lineColumnCount; j++) {` loops and instead do something like this:

```ts
//  Outside the  await chunkedSubtask, similarly for the bonds.
const tokenIndexToColumn = [
  atom_idTokens,
  atom_nameTokens,
  // ...
];

// Inside the for (let i = 0; i < linesToRead; i++) loop
let tokenIndex = 0;
skipWhitespace(tokenizer);
while (true) {
  skipStrictWhitespace(tokenizer); // new function added to Tokenizer that doesn't consider newlines whitespace
  tokenizer.tokenStart = tokenizer.position;
  eatValue(tokenizer);
  if (tokenizer.tokenStart === tokenizer.tokenEnd) break;

  const col = tokenIndexToColumn [tokenIndex++];
  if (!col) continue;
  TokenBuilder.addUnchecked(col, tokenizer.tokenStart, tokenizer.tokenEnd);
}
```

What do you think?

QDMarkMan:
Oh, thank you for your suggestion! I had overlooked performance testing due to a lack of understanding of the `streaming tokenizer` design. I’ll review your suggestion and redesign my code accordingly.

src/mol-io/reader/_spec/mol2.spec.ts
```diff
@@ -476,6 +726,30 @@ describe('mol2 reader', () => {
...
+    it('bond status_bit', async () => {
+        const parsed = await parseMol2(Mol2BondWithStatusBit, '').run();
+        if (parsed.isError) {
+            throw new Error(parsed.message);
+        }
+        const mol2File = parsed.result;
+        const data = mol2File.structures[0];
+        const statusBits = data.bonds.status_bits.toArray();
+        expect(statusBits.length).toEqual(1);
+        expect(statusBits[0]).toEqual('BACKBONE|DICT|INTERRES');
```
dsehnal, line 750: There is a bug where the status bits are assigned to the wrong atom. This is caused by the field not being present in the 1st row, but rather 2nd in case of the atoms, and 18th in case of the bonds.

So the test should say 

```
expect(statusBits.length).toEqual(data.atoms.count);
expect(statusBits[1]).toEqual('BACKBONE|DICT|DIRECT');
for (let i = 0; i < data.atoms.count; i++) {
  if (i !== 1) expect(statusBits[i]).toEqual('');
}
```

(and `statusBits[17]` for the bonds).

To fix this, you need to include

```ts
for (let cI = tokenIndex; cI < tokenIndexToColumn.length; cI++) {
  TokenBuilder.addUnchecked(col, 0, 0); // I hope this works as expected
}
```

to `_readColumnsAndGetMaxCount` at the end of the for loop there. 

The original parsed was written with the assumption that each row has the same number of columns so it was not present.

QDMarkMan: Oh thanks for your help, It seems I missunderstood the data structure of status_bits.

dsehnal: Good. Still need to remove the `dataLines` you've added.

QDMarkMan: Got it. Thanks for the heads-up!

dsehnal:
Almost there, looking good. Before merging this, can you also please complete the checklist in the PR template:

image

QDMarkMan:
Got it! it’s all set!

dsehnal (approved):

PR #1253, opened 2024-09-03
Illumination mode

arose:
- Path-traced SSGI
- Automatic thickness (estimate)
    - Base thickness as max(backface depth) - min(frontface depth)
    - Per object density factor to adjust thickness
- Progressively trace samples to keep viewport interactive
- Toggle on/off by pressing "G"

Try https://molstar.org/dev/illumination/

6AU3

1TQN - 2024-09-02T222715 234


dsehnal:
Very cool! 

I went ahead and added a fallback to standard rendering when the user is interacting with the viewport.

src/mol-canvas3d/passes/image.ts
```diff
@@ -124,8 +112,9 @@ export class ImagePass {
                 if (isTimingMode) this.webgl.timer.markEnd('ImagePass.render');
                 if (runtime.shouldUpdate) {
                     await runtime.update({ current: this.illuminationPass.iteration });
+                    await Scheduler.requestAnimationFramePromise(); // allow UI to render
                 }
-                await waitForRequestAnimationFrame(); // allow GPU to render
+                await Scheduler.immediatePromise(); // allow GPU to render
```
dsehnal, line 117: Tweaked this part. On my machine it rendered correctly with this approach.

corredD:
Looks incredible! Great work @arose !

giagitom:
Thanks for this amazing addition @arose! 
Results are really impressive. 

I have made some tests, and so far I just noticed a few things:

- A few times I had issues changing shading parameters at representation level. But I haven't found a way to reproduce so far.  
- I tried exporting images in 4K in the mesoscale explorer and the first time I experienced a webgl context lost during the tracing process, the second time the downloaded image was empty and the third time I succeeded in exporting it. Also it requires a lot of time for the tracing process (something like a few minutes), Is this due to the increased number of samples? Is this normal? 
- In this [test.molj](https://github.com/user-attachments/files/16849255/test.molj.txt) I used a single light with 0 ambient lighting. Shouldn't the shaded parts be completely or close to black?

image

giagitom:
I tried for fun the new illumination mode on my cell phone and found these artifacts. Is this just a matter of performances or there could be something else going on?

image

giagitom:
With this test state [test.molj](https://github.com/user-attachments/files/16887731/test-2.molj.txt), if I try to increase "refine steps" parameter more than 4, I always get a context lost while downloading UHD image

src/mol-gl/shader/illumination/compose.frag.ts
```diff
@@ -0,0 +1,197 @@
...
+//  float threshold   - edge sharpening threshold
+
+float NormalWeightStrength = 6.0;
+
+vec4 smartDeNoise(sampler2D tex, vec2 uv)
+{
+    float sigma = 3.0;
+    float kSigma = 2.0;
+    float threshold = 0.1;
+    float radius = sigma * kSigma;
```
giagitom, line 80: radius and radQ seems unused

arose:
> With this test state [test.molj](https://github.com/user-attachments/files/16887731/test-2.molj.txt), if I try to increase "refine steps" parameter more than 4, I always get a context lost while downloading UHD image

Please try again, with [490bc82](https://github.com/molstar/molstar/pull/1253/commits/490bc82ee605bc8c706127b739b077c0f833f3a9) and [b6885a0](https://github.com/molstar/molstar/pull/1253/commits/b6885a0d760ce57b864223c05993048d23a089fa) this should work much better.

src/mol-gl/shader/illumination/compose.frag.ts
arose: removed

arose:
> I tried for fun the new illumination mode on my cell phone and found these artifacts. Is this just a matter of performances or there could be something else going on?
> 
> image

I see this too on a Pixel 5. Not idea what causes it. Appreciate any suggestions & fixes.

arose:
> In this [test.molj](https://github.com/user-attachments/files/16849255/test.molj.txt) I used a single light with 0 ambient lighting. Shouldn't the shaded parts be completely black?

I added an option in [4ee4788](https://github.com/molstar/molstar/pull/1253/commits/4ee47883782da281d6e077e0ce60caa7ac3de88c) to handle this.

giagitom:
> > In this [test.molj](https://github.com/user-attachments/files/16849255/test.molj.txt) I used a single light with 0 ambient lighting. Shouldn't the shaded parts be completely black?
> 
> I added an option in [4ee4788](https://github.com/molstar/molstar/pull/1253/commits/4ee47883782da281d6e077e0ce60caa7ac3de88c) to handle this.

Thanks for the glow option, it is working nicely now.
An other thing I noted is that with a single light everything is working well. But when adding two lights it seems that the effect of the first light is no more taken into account (or changes in some way). Here is a test [test.molj.txt](https://github.com/user-attachments/files/16930618/test.molj.txt)  

### Single light on 
image

### Two lights on
image

giagitom:
When using non-black outlines here is what i am seeing. Is this wanted? 
If not, maybe it could be an option to uniform outline color with pixel luminosity? 

image

arose:
> I tried for fun the new illumination mode on my cell phone and found these artifacts. Is this just a matter of performances or there could be something else going on?

fixed it, forgot to specify high int precision (always high on desktop), so random number were not random enough...

image

arose:
> When using non-black outlines here is what i am seeing. Is this wanted?

expect, works the same way as without ssgi

arose:
> An other thing I noted is that with a single light everything is working well. But when adding two lights it seems that the effect of the first light is no more taken into account (or changes in some way).

this was due to shadows enabled. multiple lights were not properly supported with shadows (neither when ssgi is off, fix in #1262)

should be better now but don't expect too much from shadows

arose:
I am happy with it as is.

giagitom (approved):
Thanks for all fixes @arose

PR #1254, opened 2024-09-04
ErrorContext util & usage in MVS

dsehnal:
# Description

Introduce new `ErrorContext` in `PluginContext` and use it in `CustomProperty.Context` as suggested by @midlik in #1238.


arose (approved):
lgtm

PR #1262, opened 2024-09-14
fix shadows with multiple lights

arose:
image


corredD (approved):
Looks good to me, thanks !

giagitom (approved):
Nice! Thanks for the fix

PR #1264, opened 2024-09-16
Fix binarySearchPredIndexRange

midlik:
# Description

This fixes a bug in `binarySearchPredIndexRange`, `findPredecessorIndex`, `findPredecessorIndexInInterval`.

Current implementation does not always return the first occurrence of the query value if it occurs more than once, e.g. `findPredecessorIndex([0, 1, 2, 3, 4, 6, 6, 6, 8, 9, 10, 11, 12], 6))` returns 6, but it should return 5 according to the docstring.

## Actions


dsehnal:
What is the use case for this? The design goal for this was to make this fast of arrays with unique numbers, so this was not a consideration, so perhaps we might consider changing the docstring and adding an additional method instead?

Alternatively, would just changing `return mid;` to something like 

```ts
if (mid === 0 || xs[mid - 1] !== v) return min;
min = mid - 1;
while (xs[min] === v) min--;
return min + 1;
```

work?

midlik:
Some MolViewSpec code calls this and relies on the behavior described in the docstring (i.e. that it always returns index of the FIRST element which is >= query). Because of the bug, some MolViewSpec selections might be missing atoms (it would be only a few atoms, so hard to spot visually, but still incorrect).

> Alternatively, would just changing return mid; to something like ...
Yeah, first I tried to fix the existing code but, to be honest, I got a bit lost in how that code works. So I just re-implemented it and the result is simpler than the original, IMHO (it has fewer loops and ifs, at least).

dsehnal:
@midlik This works:

```ts
function binarySearchPredIndexRange(xs: Nums, value: number, start: number, end: number) {
    let min = start, max = end - 1;
    while (min < max) {
        // do a linear search if there are only 10 or less items remaining
        if (min + 11 > max) {
            for (let i = min; i <= max; i++) {
                if (value <= xs[i]) {
                    return i === min ? normalizePred(xs, start, end, i) : i;
                }
            }
            return max + 1;
        }
        const mid = (min + max) >> 1;
        const v = xs[mid];
        if (value < v) max = mid - 1;
        else if (value > v) min = mid + 1;
        else return normalizePred(xs, start, end, mid);
    }
    if (min > max) return max + 1;
    return xs[min] >= value ? min : min + 1;
}

function normalizePred(xs: Nums, start: number, end: number, idx: number) {
    if (idx <= start) return idx;

    const value = xs[idx];
    let i: number;
    for (i = idx - 1; i >= start; i--) {
        if (xs[i] < value) break;
    }
    return i + 1;
}
```

Can you also please add these test cases:

```ts
    const aDuplSmall = SortedArray.ofSortedArray([1, ...new Array(2).fill(3), 3]);
    test('predIndexDuplSmall', SortedArray.findPredecessorIndex(aDuplSmall, 2), 1);
    const aDuplBig = SortedArray.ofSortedArray([1, ...new Array(333).fill(2), ...new Array(666).fill(3), 4]);
    test('predIndexDuplBig', SortedArray.findPredecessorIndex(aDuplBig, 3), 334);
```

--------------------------------------

If you feel like doing performance tests, for example by extending `perf-tests/binary-search.ts` and your solution is faster, then all I am for using your version.

dsehnal (approved):
LGTM, thanks.

PR #1273, opened 2024-09-23
Move away from deprecated SASS explicit color by using `invert` and `change`

pechersky:
# Description

## Actions


arose:
the replacement for the comp function produces different colors

image
with your change


image
without your change

pechersky:
Thanks for flagging this. I was mistaken in how I inlined `increase-constrast` -- I now see that it is reversed in the light vs dark/blue skins. Pushed commits to fix. This is how it looks like now:

image

arose:
thanks @pechersky

PR #1275, opened 2024-09-25
fix preset coloring in ME

corredD:
# Description
I removed the argument 'ent:' which was breaking the quick style preset 

## Actions


dsehnal (approved):

arose:
How does it break things @corredD? For me, not having the 'ent:' breaks things.

corredD:
When using the 'ent:' filter the color scheme is not apply on some element. See below for comparison. It really show with the illustrative and the siny style. Left is without the flag, right is with. I loaded the HIV petworld. But same behavior with cellpack models as well. @arose  what model are you testing, and what is not working without the flag ? 
Screenshot 2024-09-26 114248
Screenshot 2024-09-26 114304
Screenshot 2024-09-26 114214

arose:
I see, there are many more issues with applying the illustrative color theme. I need to look at the changes from back when the illustrative theme was added.

arose:
The issue, I was trying to fix with the 'ent:' is that the generated color list is all wrong because getAllLeafGroups will double count entities when called without a tag.

PR #1277, opened 2024-09-28
ME: color and style fixes

arose:
Fixes for coloring/styling in ME

see #1277


arose:
Hi @corredD, I changed to quick-styles so that the coloring is not re-applied but only style/light options are changed. This allows to change the style while keeping any custom colors.

I still need to fix some things when changing options in the entity tree.

corredD (commented):
Thanks @arose, looks great. The shiny-dof style should not use the cell shading I think.

arose:
> The shiny-dof style should not use the cell shading I think.

fixed

PR #1278, opened 2024-09-30
basic raw volume data support

aliaksei-chareshneu:
# Description

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ Note that since we don't clearly distinguish between a public and private interf
     - `native`, no changes
 - Add `CustomProperty.Context.errorContext` to support reporting errors during loading of custom properties (#1254)
   - Use in MolViewSpec extension
-- Add basic support for volumetric data
+- Add basic support for volumetric data to MolViewSpec extension
 
 ## [v4.6.0] - 2024-08-28
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ Note that since we don't clearly distinguish between a public and private interf
     - `native`, no changes
 - Add `CustomProperty.Context.errorContext` to support reporting errors during loading of custom properties (#1254)
   - Use in MolViewSpec extension
+- Add basic support for volumetric data
 
 ## [v4.6.0] - 2024-08-28
 
--- a/src/extensions/mvs/load.ts
+++ b/src/extensions/mvs/load.ts
@@ -92,7 +92,6 @@ export interface MolstarLoadingContext {
     nearestReprMap?: Map<MolstarNode, MolstarNode<'representation'>>,
     focus?: { kind: 'camera', params: ParamsOfKind<MolstarTree, 'camera'> } | { kind: 'focus', focusTarget: StateObjectSelector, params: ParamsOfKind<MolstarTree, 'focus'> },
     canvas?: ParamsOfKind<MolstarTree, 'canvas'>,
-    pluginContext?: PluginContext
 }
 
 
@@ -155,6 +154,7 @@ const MolstarLoadingActions: LoadingActions<MolstarTree, MolstarLoadingContext> 
         return model;
     },
     raw_volume(updateParent: UpdateTarget, node: SubTreeOfKind<MolstarTree, 'raw_volume'>, context: MolstarLoadingContext): UpdateTarget {
+        // TODO: write map metadata to the context
         const props = rawVolumeProps(node);
         const rawVolume = UpdateTarget.apply(updateParent, VolumeFromCcp4, props);
         // TODO: tooltips, labels, etc.
--- a/src/extensions/mvs/load.ts
+++ b/src/extensions/mvs/load.ts
@@ -242,9 +242,7 @@ const MolstarLoadingActions: LoadingActions<MolstarTree, MolstarLoadingContext> 
             ...visualParams,
             ...volumeRepresentationProps(node.params),
         };
-        const update = UpdateTarget.apply(updateParent, VolumeRepresentation3D, {
-            ...params
-        });
+        const update = UpdateTarget.apply(updateParent, VolumeRepresentation3D, params);
         return update;
     },
     color: undefined, // No action needed, already loaded in `representation`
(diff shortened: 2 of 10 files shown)


src/extensions/mvs/tree/mvs/mvs-builder.ts
```diff
@@ -209,6 +219,20 @@ export class Component extends _Base<'component' | 'component_from_uri' | 'compo
...
+    color(params: ParamsOfKind<MVSTree, 'color'>): VolumeRepresentation {
+        this.addChild('color', params);
+        return this;
+    }
+    /** Add a 'color_from_uri' node and return builder pointing back to the volume_representation node. 'color_from_uri' node instructs to apply colors to a visual representation. The colors are defined by an external annotation resource. */
+    colorFromUri(params: ParamsOfKind<MVSTree, 'color_from_uri'>): VolumeRepresentation {
+        this.addChild('color_from_uri', params);
+        return this;
+    }
+}
```
midlik, line 234: We don't have `color_from_uri` on `volume_representation` in backend, and it doesn't even make much sense at this stage, as there are no selectors which could be applied to volumes.

src/extensions/mvs/tree/molstar/molstar-tree.ts
```diff
@@ -29,6 +29,13 @@ export const MolstarTreeSchema = TreeSchema({
                 format: RequiredField(MolstarParseFormatT),
             },
         },
+        raw_volume: {
+            ...FullMVSTreeSchema.nodes.raw_volume,
+            params: {
+                source: RequiredField(RawVolumeSourceT),
+                options: OptionalField(RawVolumeOptionsT)
+            }
+        },
```
midlik, line 38: If this node is exactly the same as in `MVSTreeSchema`, do not include it here explicitely. It is already included via `...FullMVSTreeSchema.nodes` on line 18.

The nodes explicitely listed here are only those that have different params in `MolstarTreeSchema` than in `MVSTreeSchema`.

src/extensions/mvs/tree/molstar/conversion.ts
```diff
@@ -46,6 +47,13 @@ const mvsToMolstarConversionRules: ConversionRules<FullMVSTree, MolstarTree> = {
             { kind: 'structure', params: omitObjectKeys(node.params, ['block_header', 'block_index', 'model_index']) },
         ] satisfies MolstarNode[];
     },
+    'raw_volume': (node, parent) => {
+        if (parent?.kind !== 'parse') throw new Error('Parent of "structure" must be "parse".');
+        // const { format } = ParseFormatMvsToMolstar[parent.params.format];
+        return [
+            { kind: 'raw_volume', params: { source: node.params.source, options: node.params.options } }
+        ] satisfies MolstarNode[];
+    }
```
midlik, line 56: If a node is converted into exactly the same node, it shouldn't be included here. This is only for non-trivial conversion rules.

src/extensions/mvs/load.ts
```diff
@@ -216,6 +224,29 @@ const MolstarLoadingActions: LoadingActions<MolstarTree, MolstarLoadingContext>
...
+                    'value': 1
+                }
+            }
+        };
+        const params = {
+            ...visualParams,
+            ...volumeRepresentationProps(node.params),
+        };
+        const update = UpdateTarget.apply(updateParent, VolumeRepresentation3D, {
+            ...params
```
midlik, line 246: No need to use `{...params}` instead of just `params` here. Or am I missing something.

src/extensions/mvs/load.ts
```diff
@@ -152,6 +153,13 @@ const MolstarLoadingActions: LoadingActions<MolstarTree, MolstarLoadingContext>
         });
         return model;
     },
+    raw_volume(updateParent: UpdateTarget, node: SubTreeOfKind<MolstarTree, 'raw_volume'>, context: MolstarLoadingContext): UpdateTarget {
+        // TODO: write map metadata to the context
+        const props = rawVolumeProps(node);
+        const rawVolume = UpdateTarget.apply(updateParent, VolumeFromCcp4, props);
+        // TODO: tooltips, labels, etc.
+        return rawVolume;
+    },
```
midlik, line 162: I'm not sure if different values of `node.options.source` will affect implementation here.
If so, pls include a check here or in `rawVolumeProps`, so that once we add more allowed values for `source`, it will throw error instead of trying to apply wrong implementation.

midlik (commented):
Great, thank you Aliaksei.
I didn't test this, just went through the code and added a few comments.

src/extensions/mvs/tree/mvs/mvs-builder.ts
aliaksei-chareshneu: Yes, that's absolutely correct, thanks for the catch.

src/extensions/mvs/tree/molstar/molstar-tree.ts
aliaksei-chareshneu: Right, thanks for the explanation, did not know that. Will remove it from here.

src/extensions/mvs/tree/molstar/conversion.ts
aliaksei-chareshneu: Ok, got it, thanks. Removed it already.

src/extensions/mvs/load.ts
aliaksei-chareshneu: Yes, you are right, I have overdid it a bit 😄. Removed that.

src/extensions/mvs/load.ts
aliaksei-chareshneu: Yes, you are right, currently the different values of `node.options.source` will indeed not affect the implementation. I have added the `node.options.source` attribute since in we potentially may add support for other sources of volumetric data in the future, e.g., for OMEZarr. In that case, we will need other transforms or mechanisms for parsing them and transforming to the Mol* `Volumes`.

Sure, will add a check here.

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ Note that since we don't clearly distinguish between a public and private interf
     - `native`, no changes
 - Add `CustomProperty.Context.errorContext` to support reporting errors during loading of custom properties (#1254)
   - Use in MolViewSpec extension
+- Add basic support for volumetric data
```
arose, line 28: this is just for MVS, right?

aliaksei-chareshneu: > this is just for MVS, right?

Yes, you are right, I will edit the description to highlight that this is just for MVS. I apologise for missing this point initially.

aliaksei-chareshneu: > this is just for MVS, right?

Fixed in ba42726f85b2056013ea5a3e4912d28e604985e8. Thanks again for the catch.

PR #1293, opened 2024-10-08
Lammps data

corredD:
# Description
Naive implementation of Lammps Data (https://docs.lammps.org/2001/data_format.html) and Traj parser ( string with https://docs.lammps.org/dump.html ). We can load the model directly from the first frame of the simulation, but I am not sure I did that properly. 
Also a lammps trajectory can have different number of atoms between frame, not sure its feasible, and I don't have example of files for it. 

examples 
taken from https://zenodo.org/records/8217689
Screenshot 2024-10-07 161448
and 
from my own data
https://github.com/user-attachments/assets/6ed438c0-d892-4f7a-bef3-afeab731890f


arose:
Nice, will look at it later.

> Also a lammps trajectory can have different number of atoms between frame, not sure its feasible, and I don't have example of files for it.

that is in principle supported but probably needs some extra code when creating the Trajectory object from model and coordinates.

corredD:
I was looking at the data here https://zenodo.org/records/8217689 and realized they scale the coordinate by 1/400.0. SO I try to add a scaling option in the OpenFiles state, probably not in the way it should have been done, but I didn't know how to do differently.  Results looks great though
Screenshot 2024-10-10 120630

arose:
There seems to be a limited number of units available, see https://docs.lammps.org/units.html. Maybe that can be used instead of a numeric scale field? I found the units for a simulation in the "systems.init" file.

The units param should only be shown when a specific format is picked, not generally.

corredD:
I am happy with the current state of the parser. Let me know how can I clean up and improve it for a merge.
LAMMPSTRJ_animate-trajectory(6)(1)

src/mol-io/reader/lammps_data/parser.ts
```diff
@@ -0,0 +1,231 @@
...
+ * @author Ludovic Autin <ludovic.autin@gmail.com>
+ */
+
+import { Task, RuntimeContext, chunkedSubtask } from '../../../mol-task';
+import { Tokenizer, TokenBuilder } from '../common/text/tokenizer';
+import { ReaderResult as Result } from '../result';
+import { TokenColumnProvider as TokenColumn } from '../common/text/column/token';
+import { Column } from '../../../mol-data/db';
+
+export interface LammpDataFile {
```
arose, line 14: Lammps, right?

arose (commented):
Please use a dash instead of an underscore in file/directory names

src/mol-plugin-state/transforms/model.ts
```diff
@@ -375,6 +400,49 @@ const TrajectoryFromXYZ = PluginStateTransform.BuiltIn({
     }
 });
 
+type TrajectoryFromLammpsData = typeof TrajectoryFromLammpsData
+const TrajectoryFromLammpsData = PluginStateTransform.BuiltIn({
+    name: 'trajectory-from-lammps-data',
+    display: { name: 'Parse Lammps Data', description: 'Parse Lammps Data string and create trajectory.' },
+    from: [SO.Data.String],
+    to: SO.Molecule.Trajectory,
+    params: {
```
arose, line 409: This should be PD.Select with the options from https://docs.lammps.org/units.html. Maybe call it unitsStyle in line with the lammps docs?

arose (changes requested):

arose:
> I was looking at the data here https://zenodo.org/records/8217689 and realized they scale the coordinate by 1/400.0. SO I try to add a scaling option in the OpenFiles state, probably not in the way it should have been done, but I didn't know how to do differently. Results looks great though

Is the scaling of the coordinates really needed here? It seems they are angstrom and adjusting the size theme of the representation is fine, or?

arose:
Where did you get the trajectory in the animation from? https://zenodo.org/records/8217689 seems to only have a single frame, or am I doing something wrong?

corredD:
Thanks @arose for looking at it. I put some example lammpstrj here : 
https://mesoscope.scripps.edu/data/sim_data/26_293_nvt.lammpstrj
https://mesoscope.scripps.edu/data/sim_data/47CAG_200um_293.lammpstrj
https://mesoscope.scripps.edu/data/sim_data/Nup1.lammpstrj 
the lats one correspond to the capsid through the nuclear pore simulation. I had to re-run it on my computer, so I only have 20frame so far. 
About the scaling, it was my misunderstanding, in the trajectory file sometimes they use the sx,sy,sz scaling approach which normalize the coordinate to the bounding box. So the scaling option is not really needed for now. But I will try to add the unit option.

src/mol-io/reader/lammps/data/parser.ts
```diff
@@ -0,0 +1,212 @@
...
+                switch (atom_style) {
+                    case 'full': {
+                        switch (j) {
+                            case 0: TokenBuilder.addUnchecked(atomId, tokenizer.tokenStart, tokenizer.tokenEnd); break;
+                            case 1: TokenBuilder.addUnchecked(moleculeType, tokenizer.tokenStart, tokenizer.tokenEnd); break;
+                            case 2: TokenBuilder.addUnchecked(atomType, tokenizer.tokenStart, tokenizer.tokenEnd); break;
+                            case 3: TokenBuilder.addUnchecked(charge, tokenizer.tokenStart, tokenizer.tokenEnd); break;
+                            case 4: TokenBuilder.addUnchecked(x, tokenizer.tokenStart, tokenizer.tokenEnd); break;
+                            case 5: TokenBuilder.addUnchecked(y, tokenizer.tokenStart, tokenizer.tokenEnd); break;
+                            case 6: TokenBuilder.addUnchecked(z, tokenizer.tokenStart, tokenizer.tokenEnd); break;
```
dsehnal, line 69: Instead of this, I would recommend doing:

```ts
const columns = {
  full: [atomId, moleculeType, ...],
  atomic: [atomId, atomType, ...],
  // ...
}

// ...

const column = columns[atom_style]?.[j];
if (column)  TokenBuilder.addUnchecked(column, tokenizer.tokenStart, tokenizer.tokenEnd);
```

src/mol-io/reader/lammps/traj/parser.ts
```diff
@@ -0,0 +1,168 @@
...
+        if (line.includes('ITEM: TIMESTEP')) {
+            timestep = parseInt(readLine(state.tokenizer).trim());
+            f.times.push(timestep);
+        } else if (line.includes('ITEM: NUMBER OF ATOMS')) {
+            numAtoms = parseInt(readLine(state.tokenizer).trim());
+        } else if (line.includes('ITEM: ATOMS')) {
+            // this line provide also the style of the output and will give the order of the columns
+            const parts = line.split(' ').slice(2);
+            const frame: LammpsFrame = await handleAtoms(state, numAtoms, parts);
+            frames?.push(frame);
```
dsehnal, line 131:
```suggestion
            frames.push(frame);
```

src/mol-plugin-state/transforms/model.ts
```diff
@@ -375,6 +400,49 @@ const TrajectoryFromXYZ = PluginStateTransform.BuiltIn({
...
+        unitsStyle: PD.Select('real', PD.arrayToOptions(['real', 'metal', 'si', 'cgs', 'electron', 'micro', 'nano', 'lj'])),
```
dsehnal, line 410: Should the unit kind/factor be passed here?

dsehnal (commented):
Nice. Before merging this, please consider the suggested improvements to the parser.

src/mol-io/reader/lammps/data/parser.ts
```diff
@@ -0,0 +1,186 @@
+/**
+ * Copyright (c) 2019 mol* contributors, licensed under MIT, See LICENSE file for more info.
```
arose, line 2: please update the headers at some point

src/mol-script/util/id-list.ts
```diff
@@ -91,10 +114,18 @@ function parseAtomListSelection(input: string): [number, number][] {
 }
 
 // parses a list of residue ranges, e.g. A 10-100, B 30, C 12:i
-export function compileIdListSelection(input: string, idType: 'auth' | 'label' | 'atom-id') {
+export function compileIdListSelection(input: string, idType: 'auth' | 'label' | 'atom-id' | 'atom-type') {
```
arose, line 117: please change atom-type to element-symbol

src/mol-io/reader/lammps/schema.ts
```diff
@@ -0,0 +1,206 @@
...
+    viscosity: string;
+    charge: string;
+    dipole?: string;
+    electricField?: string;
+    density: string;
+    scale: number;
+}
+
+export const unitStyles: { [key: string]: LammpsUnitStyle } = {
+    lj: {
```
arose, line 27: can we actually support that? How would that data look?

src/mol-io/reader/lammps/traj/parser.ts
```diff
@@ -0,0 +1,168 @@
...
+            f.times.push(timestep);
+        } else if (line.includes('ITEM: NUMBER OF ATOMS')) {
+            numAtoms = parseInt(readLine(state.tokenizer).trim());
+        } else if (line.includes('ITEM: ATOMS')) {
+            // this line provide also the style of the output and will give the order of the columns
+            const parts = line.split(' ').slice(2);
+            const frame: LammpsFrame = await handleAtoms(state, numAtoms, parts);
+            frames.push(frame);
+        } else if (line.includes('ITEM: BOX BOUNDS')) {
+            /* we could parse the bounding box if needed */
```
arose, line 133: you parse it, right?

src/mol-model-formats/structure/lammps_data.ts
```diff
@@ -0,0 +1,145 @@
+/**
```
arose, line 1: please rename to lammps-data.ts

src/mol-model-formats/structure/lammps_trajectory.ts
```diff
@@ -0,0 +1,172 @@
+/**
```
arose, line 1: please rename to lammps-trajectory.ts

src/mol-model-formats/structure/lammps_trajectory.ts
```diff
@@ -0,0 +1,172 @@
...
+import { Column, Table } from '../../mol-data/db';
+import { Trajectory } from '../../mol-model/structure';
+import { BasicSchema, createBasic } from './basic/schema';
+import { ComponentBuilder } from './common/component';
+import { EntityBuilder } from './common/entity';
+import { createModels } from './basic/parser';
+import { MoleculeType } from '../../mol-model/structure/model/types';
+import { ModelFormat } from '../format';
+
+export function coordinatesFromLammpsTrajectory(file: LammpsTrajectoryFile, unitsStyle: string = 'real'): Task<Coordinates> {
```
arose, line 21: please type unitsStyle

src/mol-model-formats/structure/lammps_data.ts
```diff
@@ -0,0 +1,145 @@
...
+import { RuntimeContext, Task } from '../../mol-task';
+import { ModelFormat } from '../format';
+import { createModels } from './basic/parser';
+import { BasicSchema, createBasic } from './basic/schema';
+import { ComponentBuilder } from './common/component';
+import { EntityBuilder } from './common/entity';
+import { IndexPairBonds } from './property/bonds/index-pair';
+import { AtomPartialCharge } from './property/partial-charge';
+
+async function getModels(mol: LammpsDataFile, ctx: RuntimeContext, unitsStyle: string = 'real') {
```
arose, line 22: please type unitsStyle

src/mol-plugin-state/formats/trajectory.ts
```diff
@@ -78,6 +78,10 @@ export const CifCoreProvider: TrajectoryFormatProvider = {
 function directTrajectory<P extends {}>(transformer: StateTransformer<PluginStateObject.Data.String | PluginStateObject.Data.Binary, PluginStateObject.Molecule.Trajectory, P>, transformerParams?: P): TrajectoryFormatProvider['parse'] {
     return async (plugin, data, params) => {
         const state = plugin.state.data;
+        // I dont think this is correct, but I'm not sure what the correct way to do this is
```
arose, line 81: why do it in the first place? what does it solve?

src/mol-plugin-state/formats/trajectory.ts
```diff
@@ -112,6 +116,24 @@ export const XyzProvider: TrajectoryFormatProvider = {
...
+    parse: directTrajectory(StateTransforms.Model.TrajectoryFromLammpsData),
+    visuals: defaultVisuals
+};
+
+export const LammpsTrajDataProvider: TrajectoryFormatProvider = {
+    label: 'Lammps Traj Data',
+    description: 'Lammps Traj Data',
+    category: TrajectoryFormatCategory,
+    stringExtensions: ['lammpstrj'],
+    parse: directTrajectory(StateTransforms.Model.TrajectoryFromLammpsTrajData),
```
arose, line 133: Maybe use *LammpsTrajectory throughout. Would match  LammpsTrajectoryFile

src/mol-plugin-state/formats/registry.ts
```diff
@@ -53,11 +53,11 @@ export class DataFormatRegistry {
     }
 
     constructor() {
+        for (const [id, p] of BuiltInTrajectoryFormats) this.add(id, p);
```
arose, line 56: any particular reason to change the order? Is it important? If so, make a note.

arose (changes requested):

arose:
This is nice @corredD, thanks! I made a few tweaks. The biggest is using the moleculeId as the asym_id. This has the effect that the capsid example has ~90k units which breaks the default rendering, but changing from 'element-spheres' to 'structure-element-spheres' handles that. It is nice because it allows to easily select molecules/chains.

src/mol-io/reader/lammps/schema.ts
corredD: I feel like its more like metadata that we can use to annotate/labels thing. I don't mind removing it.

src/mol-plugin-state/formats/registry.ts
corredD: The .lammpstrj format should be considered as a Trajectory when drag/drop or File Open, but as a Coordinates when using the Load Trajectory button that expect a topology (here .dat) and a trajectory (.lammpstrj). So because .lammpstrj  is in both category, I reorder them so that it will first check for the BuiltInTrajectoryFormats rather than the BuiltInCoordinatesFormats category. For instance, I should be able to load NPC-NUP98_Capsid.data + Nup1.lammpstrj with Load Trajectory, but also simply dropping Nup1.lammpstrj directly. Does that make sense ? Do you think there is another way to approach this? Also the Load Trajectory doesn't work currently, so I need to fix it/

corredD:
Thanks @arose this is great! I went through your comments and hopefully answered them. See also my other answer above.

src/mol-plugin-state/transforms/model.ts
corredD: If we want this params to appears in the gui ( not necessary ), how does that work ?

src/mol-plugin-state/formats/trajectory.ts
corredD: I try to rename and I ended up with : 
LammpsDataProvider: TrajectoryFormatProvider
LammpsTrajectoryDataProvider: TrajectoryFormatProvider
and 
LammpsTrajectoryProvider = DataFormatProvider
the .lammps should be both loadable directly and as a coordinate file for the .dat. What about maybe in trajectory.ts 
LammpsProvider: TrajectoryFormatProvider
LammpsTrajectoryProvider: TrajectoryFormatProvider
and in coordinate.ts
LammpsTrajectoryDataProvider: DataFormatProvider

src/mol-plugin-state/formats/registry.ts
arose: I'll need some time to think this through. As it is, it would definitely need a comment so the next person is not just changing it...

src/mol-io/reader/lammps/schema.ts
arose: It's fine to keep it, just curious. Also, you can get the type and name for the param from this. Something like `keyof type unitStyles` and `Object.key(unitStyles)`. We usually uppercase the first letter for such constant data.

src/mol-plugin-state/formats/registry.ts
corredD: I revert to the default order so I can use the Load Trajectory UI, and it follow the behavior of the other format. Maybe it could be possible to popup a small widget after the drop to pick trajectory or coordinate provider ?

arose: Good call to move this issue out of the PR.

arose (approved):

arose:
Thanks @corredD, nice contribution!

giagitom:
I am not sure how to load the trajectory. 
I am using https://mesoscope.scripps.edu/data/sim_data/Nup1.lammpstrj as trajectory and ```NPC-NUP98_Capsid.data``` as model file.

Am I doing something wrong?

image

corredD:
at https://molstar.org/viewer/ I had no issue loading it with the same input :
image

Maybe you can try to load the traj directly : 
image

giagitom:
I have just found that it seems a browser related issue strangely.. It doesn't work in my chrome browser but it work in Firefox instead. Not sure why.. I am on Ubuntu with latest Chrome version Version 131.0.6778.85 (Official Build) (64-bit)

corredD:
you are right, found the same issue with Edge. I will test on safari tonight.

giagitom:
The issue seems not related to the parser itself. ```parseLammpsTrajectory``` is receiving an empty string as input. ```a.data``` is an empty string for some reason.

https://github.com/molstar/molstar/blob/ba38fe2474734460332392b842c822cc6fbe027f/src/mol-plugin-state/transforms/model.ts#L162-L177

This is SO.Data.String
image

giagitom:
@arose, with https://github.com/molstar/molstar/commit/8bda5103788d85ebeb019c89207313865589fa76 I am now able to visualize https://mesoscope.scripps.edu/data/sim_data/Nup1.lammpstrj using ball and stick representation without losing context. 

Something I noticed is that the system is generally very slow in updating ball and stick representation. Even just updating the structure or parameters like sizeFactor make the browser freeze for a couple of minutes. I'm not sure if this is due to the complexity of the system or there is something else going on.

giagitom:
This instead is an ssao related issue that is visible at a certain zoom level. Not sure why this happens.  

image

corredD:
The biggest file I could load was ~520Mb ( ~22frames from https://mesoscope.scripps.edu/data/sim_data/Nup1.lammpstrj )

giagitom:
Hi @corredD, no problem at all! 

It seems the new chrome update is able to handle bigger files. I am now able to load the 3 frames trjectory while still unable to load ~655Mb trajectory with 29 frames. 

> I am not sure why this happened, but chrome FileReader will return an empty string when loading files too big.

What seems strange to me is that I am able to load > 1.6 Gb trajectories in XTC format without any problems. (Maybe because XTC is in binary format ? )

arose:
There is a browser ans system dependent maximum string length, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/length#description.

Binary data is not affected by it.

The workaround would be to parse the string data in smaller chunks from binary data - but that would require quite a few changes. PR welcome :-)

corredD:
Thanks @arose, so since we can't build large string object, the workaround would need the lammps provider to actually process/parse the chunk? So does that imply changing async function processFile(file: Asset.File, plugin: PluginContext, format: string, visuals: boolean) to handle the case of provider parsing by chunk ?

corredD:
@giagitom, @arose see [PR](https://github.com/molstar/molstar/pull/1413)

PR #1297, opened 2024-10-12
fix backfaces visible using blended transparency on impostors

### LINKED ISSUE #1285: Issue with backfaces always visible using blend transparency on impostor spheres and cylinders
Wboit and Dpoit are working fine, the issue is present only using blend transparency. Here a test [state.molx](https://github.com/user-attachments/files/17243396/test-6.molx.txt) to reproduce.

**Blend**
image

**Wboit**
image

arose:
fix backfaces visible using blended transparency on impostors - fixes #1285


giagitom (approved):
Nice, thanks @arose

PR #1302, opened 2024-10-20
Basic support for Predicted Aligned Error parsing and plotting

dsehnal:
# Description

Adds support for pairwise quality assessment scores

- [x] Proof of concept
  - [x] Load data from `ma_qa_metric_local_pairwise` present in the input file
  - [x] Draw simple PNG
- [x] Plot interactivity
- [x] Support loading data from separate mmCIF 
- [x] Structure interactivity 
  - [x] Coloring based on plot selection
  - [x] Mouse over highlight
- [x] `PluginConfig` option for disabling the UI component
- [x] Standalone mode (make it easy to render the plot outside the main Mol* UI)
  - [x] Example loading data from AlphaFoldDB

### Example

Data from https://www.modelarchive.org/doi/10.5452/ma-bak-cepc-0944

image

Standalone example:

image

## Actions


arose (approved):
Only two observations
- the orange axis label is hard to read with the default viewer theme
- using overpaint for the interactivity means that any user-applied overpaint is lost which may or may not be ok

dsehnal:
@arose 

> * the orange axis label is hard to read with the default viewer theme

Made the color darker.


> * using overpaint for the interactivity means that any user-applied overpaint is lost which may or may not be ok

Added a function that re-applies the previous states to restore it.

---------------

If people will start using this, I am sure they there will be plenty feedback about making this better. My goal with this was to have some initial support we can iterate on later if needed.

PR #1303, opened 2024-10-20
SSAO improvements

arose:
Improve SSAO speed and quality. Favor speed in defaults.

- Add `blurStepSize` to reduce blur cost. Works best with multi-sample enabled to mitigate artefacts. Defaults to 2 (since multi-sample is enabled by default). Set to 1 for screenshots.
- Add `blurNormalBias` to improve creases between overlapping spheres. Quite expensive, use with care. Disabled by default (set to zero).
- Remove `devicePixelRatio` adjustment of resolution. This is a bit more expensive but avoids ugly edge artefacts.
- Reduce default `samples` from 32 to 24. Still at 128 for screenshots.
- Other
    - fix missing viewport/scissor calls
    - only render half/quarter when multi-scale is enabled
    - calculate normal from position derivative


corredD:
Nice subtle optimization! My eyes could not see a difference in ME, and was able to see some in molstar. Only issue I notice was that the blurNormalBias at 1.0 introduce noise  : 
image

corredD (approved):
nice thing to make the expansive normal calculation optional !

arose:
> My eyes could not see a difference in ME, and was able to see some in molstar.

Note, that it is disabled by default in ME, because with the lower kernel size and small spheres it is noticeable.

> Only issue I notice was that the blurNormalBias at 1.0 introduce noise

Right, need to prevent 1.0.

giagitom (commented):
Thanks for the improvements @arose. 
I noticed some artifacts related to resolution scale. Here is a test [test.molx](https://github.com/user-attachments/files/17458749/test.molx.txt)

image

arose:
> I noticed some artifacts related to resolution scale.

Right that is due to getting the normal from derivatives. It is OK for 0.5 and 0.25. Maybe we should just remove the resolution parameter for now? Or hide it from the UI?

giagitom:
> Right that is due to getting the normal from derivatives. It is OK for 0.5 and 0.25. Maybe we should just remove the resolution parameter for now? Or hide it from the UI?

Maybe hide?

arose:
> What about halving automatically resolution scale when outlines are enabled?

Nice idea, but I would not want to do that generally, it looses too much detail. It can be done with a custom UI.

arose:
@dsehnal @papillot I think you have access to laptops with high devicePixelRatio. Would you mind trying this branch and report on performance? Thanks!

papillot:
@arose I've tried it on a M1 MB pro and could not notice issues with performance.

dsehnal:
M3 runs well too in both Safari and FF.

PR #1308, opened 2024-10-26
Improve performance of `IndexPairBonds` assignment

arose:
Improve performance of `IndexPairBonds` assignment when operator keys are available.

When the operator keys are available we know the unit pairing (if units have a key set in their conformation) and can avoid looping over all bonds over and over again. Useful when there are many units.


src/mol-model/structure/structure/unit/bonds/inter-compute.ts
```diff
@@ -206,6 +207,68 @@ function findPairBonds(unitA: Unit.Atomic, unitB: Unit.Atomic, props: BondComput
...
+                    if (pairs.has(h)) pairs.get(h)!.add(i);
+                    else pairs.set(h, new Set([i]));
+                }
+            }
+        }
+
+        const { invertedIndex } = Model.getInvertedAtomSourceIndex(m);
+        const unitIds: [number, number] = [-1, -1];
+        pairs.forEach((indices, h) => {
+            const [unitIdA, unitIdB] = invertCantorPairing(unitIds, h);
```
dsehnal, line 252: Optionally, to not rely on the pairing inversion, it should work to use `new Map<number, [unitA: number, unitB: number, indices: Set<number>]>()` instead for the `pairs`.

dsehnal (approved):

Issue #1306, opened 2024-10-24
Labels look odd in Linux with nvidia GPU


josemduarte:
Not sure if this is reproducible in other Linux (without nvidia) or not. But I can see it happening consistently in my box (Chrome 130.0.6723.58, Ubuntu 22.04, Nvidia RTX2060, driver version 535.183.01).

Examples

- The ligand summary page at rcsb.org (e.g. https://www.rcsb.org/ligand/HEM) with labels enabled:
![Screenshot from 2024-10-24 16-01-50](https://github.com/user-attachments/assets/9024ff1c-64ee-4c03-ae22-5a632ec56340)

- Creating a label from the molstar.org app:
![Screenshot from 2024-10-24 16-03-19](https://github.com/user-attachments/assets/bd34177a-f810-49dc-a47b-def931ac44f5)

- Mesoscale explorer, here the effect is a bit different with an all-black effect:
![Screenshot from 2024-10-24 16-04-48](https://github.com/user-attachments/assets/8bae56b4-4b9c-4f93-a776-d6d690b9479b)

giagitom:
same issue https://github.com/molstar/molstar/issues/1208

dsehnal:
Can you try changing transparency to "Blended" in the bottom left options?

<img width="337" alt="image" src="https://github.com/user-attachments/assets/276b6c45-fdc1-450f-86fa-9558ecfa8224">

giagitom:
>Can you try changing transparency to "Blended" in the bottom left options in https://molstar.org/viewer/ ?

Hi @dsehnal,

I tried this too and didn't see any change.

josemduarte:
> Can you try changing transparency to "Blended" in the bottom left options in https://molstar.org/viewer/ ?

Same result here: no change when switching to "Blended"

arose:
Does changing

`float w = clamp(smoothness * (abs(dFdx(vTexCoord.x)) + abs(dFdy(vTexCoord.y))), 0.0, 0.5);`

to

`float w = 0.0;`

in src\mol-gl\shader\text.frag.ts fix it?

giagitom:
> float w = 0.0;
>
>in src\mol-gl\shader\text.frag.ts fix it?

Not on my side. Same as before

giagitom:
I have found that if I change this part

```if (a < 0.5) discard;```

to this

```if (sdf < 0.5) discard;```

it works fine on my side

[...]

PR #1309, opened 2024-10-27
remove extra anti-aliasing from text shader

### LINKED ISSUE #1208: Label issues
I have found two issues on labels:

Labels are ovelapping
image

and seems broken on mesoscale explorer
image

arose:
Remove extra anti-aliasing from text shader. We do enough with post-processing and multi-sampling.

Fixes #1208 & #1306 (hopefully, can't test myself).


giagitom (approved):
Nice thanks,
This solves the issue for me.

@arose may the issue be caused by a different behavior of dFdx dFdy in certain conditions?

PR #1325, opened 2024-11-03
Structure visuals

arose:
- Add `Structure.intraUnitBondMapping`
- Add more structure-based visuals to avoid too many (small) render-objects
    - `structure-intra-bond`, `structure-ellipsoid-mesh`, `structure-element-point`, `structure-element-cross`

### DIFF
--- a/src/mol-model/structure/structure/mapping.ts
+++ b/src/mol-model/structure/structure/mapping.ts
@@ -1,0 +1,92 @@
+/**
+ * Copyright (c) 2024 mol* contributors, licensed under MIT, See LICENSE file for more info.
+ *
+ * @author Alexander Rose <alexander.rose@weirdbyte.de>
+ */
+
+import { OrderedSet } from '../../../mol-data/int/ordered-set';
+import { ElementIndex } from '../model/indexing';
+import { Structure } from './structure';
+import { Unit } from './unit';
+
+/** Serial index of an element in the structure across all units */
+export type SerialIndex = { readonly '@type': 'serial-index' } & number
+
+export interface SerialMapping {
+    /** Cumulative count of preceding elements for each unit */
+    cumulativeUnitElementCount: ArrayLike<number>
+    /** Unit index for each serial element in the structure */
+    unitIndices: ArrayLike<number>
+    /** Element index for each serial element in the structure */
+    elementIndices: ArrayLike<ElementIndex>
+    /** Get serial index of element in the structure */
+    getSerialIndex: (unit: Unit, element: ElementIndex) => SerialIndex
+}
+
+export function getSerialMapping(structure: Structure): SerialMapping {
+    const { units, elementCount, unitIndexMap } = structure;
+    const cumulativeUnitElementCount = new Uint32Array(units.length);
+    const unitIndices = new Uint32Array(elementCount);
+    const elementIndices = new Uint32Array(elementCount) as unknown as ElementIndex[];
+    for (let i = 0, m = 0, il = units.length; i < il; ++i) {
+        cumulativeUnitElementCount[i] = m;
+        const { elements } = units[i];
+        for (let j = 0, jl = elements.length; j < jl; ++j) {
+            const mj = m + j;
+            unitIndices[mj] = i;
+            elementIndices[mj] = elements[j];
+        }
+        m += elements.length;
+    }
+    return {
+        cumulativeUnitElementCount,
+        unitIndices,
+        elementIndices,
+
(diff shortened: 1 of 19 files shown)


src/mol-model/structure/structure/mapping.ts
```diff
@@ -0,0 +1,92 @@
...
+/** Serial index of an element in the structure across all units */
+export type SerialIndex = { readonly '@type': 'serial-index' } & number
+
+export interface SerialMapping {
+    /** Cumulative count of preceding elements for each unit */
+    cumulativeUnitElementCount: ArrayLike<number>
+    /** Unit index for each serial element in the structure */
+    unitIndices: ArrayLike<number>
+    /** Element index for each serial element in the structure */
+    elementIndices: ArrayLike<ElementIndex>
```
dsehnal, line 21: Do we need this? The both unit and element indices can be obtained from a linear index and the `cumulativeUnitElementCount` by doing a `findPredecessorIndex` and subtracting to get `UnitIndex`.

Even if we want to skip the `findPredecessorIndex` (since the log access can have some downsides which I am not aware of now), the `elementIndices` are still reduntant for the same reason (subtraction gets you the `UnitIndex` from which you can get the `ElementIndex`). That's a lot of memory saved for large structures.

src/mol-model/structure/structure/mapping.ts
```diff
@@ -0,0 +1,92 @@
...
+    /** Get serial index of element in the structure */
+    getSerialIndex: (unit: Unit, element: ElementIndex) => SerialIndex
+}
+
+export function getSerialMapping(structure: Structure): SerialMapping {
+    const { units, elementCount, unitIndexMap } = structure;
+    const cumulativeUnitElementCount = new Uint32Array(units.length);
+    const unitIndices = new Uint32Array(elementCount);
```
dsehnal, line 29:
```suggestion
    const unitIndices = new Uint32Array(elementCount);
```

If unit count is <65536 it can be `Uint16`

src/mol-model/structure/structure/mapping.ts
```diff
@@ -0,0 +1,92 @@
...
+    };
+}
+
+//
+
+export interface IntraUnitBondMapping {
+    bondCount: number
+    unitIndex: ArrayLike<number>
+    unitEdgeIndex: ArrayLike<number>
+    unitGroupIndex: ArrayLike<number>
```
dsehnal, line 56: If the `edgeIndex` is linear within the entire group (let's call it `unitGroupEdgeIndex `), the group index could be computed by `Math.floor(unitGroupEdgeIndex / unit.bonds.edgeCount)`

dsehnal (commented):
If the goal of this is to make large structure rendering a lot faster, I think we can have substantial memory savings too.

src/mol-model/structure/structure/structure.ts
```diff
@@ -340,6 +342,10 @@ class Structure {
         return this.state.serialMapping || (this.state.serialMapping = getSerialMapping(this));
     }
 
+    get intraUnitBondMapping() {
+        return this.state.intraUnitBondMapping || (this.state.intraUnitBondMapping = getIntraUnitBondMapping(this));
```
dsehnal, line 346:
```suggestion
        return this.state.intraUnitBondMapping ??= getIntraUnitBondMapping(this);
```

src/mol-model/structure/structure/mapping.ts
```diff
@@ -0,0 +1,92 @@
...
+}
```
dsehnal, line 57: Instead of this, would it make sense to refactor the `LinkBuilderProps` (or add `StructureLinkBuilderProps`) with this pattern:

```ts
interface StrutureLinkBuilderProps {
  callback(unitIndex: number, unitGroupIndex: number, unitEdgeIndex: number): T {
  }
}
```

src/mol-model/structure/structure/structure.ts
```diff
@@ -340,6 +342,10 @@ class Structure {
         return this.state.serialMapping || (this.state.serialMapping = getSerialMapping(this));
```
dsehnal, line 342:
```suggestion
        return this.state.serialMapping ??= getSerialMapping(this);
```

arose:
> If the goal of this is to make large structure rendering a lot faster, I think we can have substantial memory savings too.

The goal is mainly to improve performance for structures with maybe up to 100k atoms in 10k units. It is a bit of a special case. The `serialMapping` was here for a long time, I just moved it. Let me merge this for now, use/test it and improve later.

PR #1328, opened 2024-11-05
Membrane orientation: improve `isApplicable` check and error handling

sbittrich:
# Description
- turns out adjusting the `isApplicable()` method isn't enough as selection symbols don't check that
- checking based on entity data seemed pretty cheap
- tested for 5xes, 5bna ("Membrane Orientation" hidden, selection symbol returns console warning) and 1brr, 6w6l

## Actions

### DIFF
--- a/src/extensions/anvil/prop.ts
+++ b/src/extensions/anvil/prop.ts
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2020 mol* contributors, licensed under MIT, See LICENSE file for more info.
+ * Copyright (c) 2020-2024 mol* contributors, licensed under MIT, See LICENSE file for more info.
  *
  * @author Sebastian Bittrich <sebastian.bittrich@rcsb.org>
  * @author Alexander Rose <alexander.rose@weirdbyte.de>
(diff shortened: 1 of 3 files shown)


giagitom:
thanks for the PR @sbittrich 

I just tested it, and it works for my use case

src/extensions/anvil/prop.ts
```diff
@@ -67,13 +67,31 @@ export const MembraneOrientationProvider: CustomStructureProperty.Provider<Membr
...
+            console.warn('Failed to predict membrane orientation. This happens for short peptides and entries without amino acids.');
+            return { value: undefined };
+        }
     }
 });
 
+function isApplicable(structure: Structure) {
+    if (!structure.isAtomic) return false;
+
+    const { byEntityKey } = structure.model.sequence;
```
arose, line 86: Could you iterate over all models? Also, accessing .model when there is more than one model will throw (@dsehnal how about we change it so that it return undefined?)

arose (changes requested):

src/extensions/anvil/prop.ts
sbittrich: Thanks, updated with https://github.com/molstar/molstar/pull/1328/commits/2ed165f9a5ff78815e6246cdf7308127136f140e.

arose (approved):
lgtm, thanks!

PR #1329, opened 2024-11-05
Update to express v5

JonStargaryen:
# Description
Updates the included servers to express v5. Migration guide is here: https://expressjs.com/en/guide/migrating-5.html.
Seemed pretty straightforward and I kept changes minimal.

Changes:
- body-parser (v2) is part of express v5, so probably good to drop that dep and use the one managed by express directly
- ModelServer: `v1/query-many` just called `req.setTimeout` -- don't think that does anything and maybe should be a config prop
- VolumeServer: simplified defined paths (changed semantics for some special characters in express v5)

Only had a closer look at ModelServer and VolumeServer but everything else looked OK. Some test instances here if you want to have a look: http://modelserver-express5.rcsb.org/ and http://volumeserver-express5.rcsb.org/ (note: no SSL certs are set up).

## Actions


arose (approved):
thank you @sbittrich, lgtm

Issue #1345, opened 2024-11-15
SWC (Next.js): "invalid destructuring target" error for array swap


simeonborko:
I discovered a bug in SWC (the default compiler for Next.js) which leads to an error in *production* build of Next.js project using molstar.

It's reported in SWC's repository: https://github.com/swc-project/swc/issues/9739

The problematic code line in molstar: https://github.com/molstar/molstar/blob/85b72ae3b0bef58927b15db2e321236609156c05/src/mol-math/misc.ts#L57

Next.js chunks all the necessary code and it seems that it runs it through SWC. Transpiled code is: `[0, -1] = [1, 0]` which is obviously syntactically incorrect.

**Workaround:** If you use Next.js and you encounter this issue, create `.babelrc` file to fall back to Babel instead of SWC, until the bug is resolved. The simplest `.babelrc`:
```
{
  "presets": ["next/babel"],
  "plugins": []
}
```

dsehnal:
Ok, I think we can replace this with standard variable assignment. Can you please make a PR with this?

simeonborko:
Sure, there it is: https://github.com/molstar/molstar/pull/1346

PR #1346, opened 2024-11-15
Refactor value swapping in molstar-math to fix SWC (Next.js) build (#1345)

simeonborko:
# Description

The current version of SWC (1.9.2) cannot handle swap using array destructuring, if the left-hand-side values are not variables but array elements. To make code in molstar as simple as possible, I got rid of any array and use variables for both storing deltas and swapping.

## Actions


dsehnal (approved):
Thanks.

Issue #1257, opened 2024-09-07
Interactions: Charge and H donor perceived on neutral Nitrogen when residue is protonated


papillot:
I've noticed the following interactions with the PDB structure 6KNR that has all hydrogens defined on its ligand DL9:
notice the highlighted Nitrogen which has 3 covalent single bonds (hence should be neutral) but participates in a Hbond and a cation-π interaction.
<img width="789" alt="image" src="https://github.com/user-attachments/assets/9e8ed5b8-ef48-4d0c-9021-2e6795a5e939">
https://www.rcsb.org/3d-view/6KNR?preset=ligandInteraction&label_asym_id=C

The mmCIF does not specify formal charge for any of the atoms (symbol `?`)
The ValenceModel seems to infer that the Nitrogen highlighted in the screen capture has a positive charge and is a putative hydrogen donor.
Can we improve this based on the protonation that is fully specified for this residue?
(I'd be happy to contribute a PR)

arose:
The valence model has options for when to assign charge/H. The complication is that only part of the structure has H/charge given. Maybe add an to check if H are given for  each unit?

arose:
> Maybe add an to check if H are given for each unit?

@papillot would you be interested inn trying this?

papillot:
Fixed

PR #1349, opened 2024-11-18
Do not infer implicit Hs when explicit Hs are set #1257

papillot:
Screenshot 2024-11-18 at 17 25 58

`6KNR` from the linked issue: the cation-Pi interaction that was previously perceived does not appear anymore by default (`assignH` set to `auto`)

# Description
This solves issue #1257  with the Valence Model where implicit hydrogens and consequently, charges, can be assigned even if the protonation is known at the level of a single unit.

## Discussion
This code does a traversal of the unit considered for computing the valence model and breaks as soon as an hydrogen has been found. I've considered a possible optimization were the upper bound could be of a few residues (not just one in case there is a capping) before we could assume that the unit has no explicit hydrogen. I don't know if that's a performance improvement worth implementing (it may introduce edge cases on some kind of polymers). 
I've decided against an arbitrary number of atoms as, some file formats like SDF/MOL favor setting hydrogens at the end of the atom block.
Another option could be to loop through the atoms and break as soon as we can find one where the expected valence does not match the explicit connectivity + charge or as soon as we find an hydrogen whichever comes first.

There are also edge cases of units that are not organic (like iron sulfur clusters `FES` in `5D8V`) where the explicit protonation of the model cannot be detected via this model. This is a larger problem than this fix though.

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ Note that since we don't clearly distinguish between a public and private interf
 - Fix `StructConn.isExhaustive` for partial models (e.g., returned by the model server)
 - Refactor value swapping in molstar-math to fix SWC (Next.js) build (#1345)
 - Fix transform data not updated when structure child changes
+- Fix do not compute implicit hydrogens when unit is explicitly protonated (#1257)
 
 ## [v4.8.0] - 2024-10-27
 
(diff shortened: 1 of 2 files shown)


dsehnal (approved):
LGTM.

@arose any objections before merging?

arose (approved):
lgtm, I think an exhaustive check is good, want it to be correct and the valence model data is cached.

PR #1355, opened 2024-11-21
Servers object storage

midlik:
# Description

Support input from Google Cloud Storage (gs://) for ModelServer and VolumeServer.

## Actions


package.json
```diff
@@ -154,6 +154,7 @@
     "webpack-cli": "^5.1.4"
   },
   "dependencies": {
+    "@google-cloud/storage": "^7.14.0",
```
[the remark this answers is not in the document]
midlik: OK, i changed it.

src/servers/model/utils/fetch-retry.ts
```diff
@@ -31,3 +34,15 @@ export async function fetchRetry(url: string, timeout: number, retryCount: numbe
...
+export async function extendedFetch(url: string, init?: fetch.RequestInit): Promise<Response> {
+    if (url.startsWith('gs://')) return fetch_GS(url, init);
+    else return fetch(url, init);
+}
+
+async function fetch_GS(url: string, init?: fetch.RequestInit): Promise<Response> {
+    if (init?.signal?.aborted) throw new AbortError('The user aborted a request.');
+    const data = await downloadGs(url);
+    return new Response(data, init);
+}
```
dsehnal, line 48: If there is no await, no need for async.

```suggestion
export function wrapFetch(url: string, init?: fetch.RequestInit): Promise<Response> {
    if (url.startsWith('gs://')) return fetchGS(url, init);
    else return fetch(url, init);
}

async function fetchGS(url: string, init?: fetch.RequestInit): Promise<Response> {
    if (init?.signal?.aborted) throw new AbortError('The user aborted a request.');
    const data = await downloadGs(url);
    return new Response(data, init);
}
```

dsehnal (approved):
LGTM, just a small tweak

src/servers/model/utils/fetch-retry.ts
midlik: I always get a bit confused by functions that are not async but return a promise :D
But you're right.
I changed it.

PR #1356, opened 2024-11-21
Improve ssao resolution scale quality

giagitom:
# Description

Render a scaled ssao texture but keep ssao blurring at full resolution improving the overall quality when scaleFactor < 1

**0.5 resolutionScale**
image
image

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Note that since we don't clearly distinguish between a public and private interf
 
 ## [Unreleased]
 
+- Improve ssao resolution scale quality
 - Enable double rounded capping on tubular helices
 - Fix single residue tubular helices not showing up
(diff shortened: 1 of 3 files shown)


arose:
Hmm, what does this actually do? You still get the artifacts at depth discontinuities but you get more AO details. However you can also get more AO detail by reducing the blur kernel size by the same scale factor. May be that is what the change should be?

giagitom:
> Hmm, what does this actually do? You still get the artifacts at depth discontinuities but you get more AO details. 

I was trying to implement something similar to what we discussed privately during the transparent ssao implementation, but effectively there is no change in the artifacts that seems the main issue to me. Probably this will be not useful.

arose:
> Probably this will be not useful.
Agree, closing

PR #1363, opened 2024-11-25
fix get bounding spheres

giagitom:
# Description

This fixes an issue on my side when I need an updated canvas3d.boundingSphereVisible  

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Note that since we don't clearly distinguish between a public and private interf
 
 ## [Unreleased]
 
+- fix get bounding spheres
 - Fix artifacts when using xray shading with high xrayEdgeFalloff values
 - Enable double rounded capping on tubular helices
 - Fix single residue tubular helices not showing up
--- a/src/mol-canvas3d/canvas3d.ts
+++ b/src/mol-canvas3d/canvas3d.ts
@@ -1013,8 +1013,8 @@ namespace Canvas3D {
                 cameraResetRequested = true;
             },
             camera,
-            boundingSphere: scene.boundingSphere,
-            boundingSphereVisible: scene.boundingSphereVisible,
+            get boundingSphere() { return scene.boundingSphere; },
+            get boundingSphereVisible() { return scene.boundingSphereVisible; },
             get notifyDidDraw() { return notifyDidDraw; },
             set notifyDidDraw(v: boolean) { notifyDidDraw = v; },
             didDraw,


arose:
under what circumstance is this an issue?

giagitom:
When I remove representations using ```canvas3d.remove(repr)``` and than try to reset camera position in this way:

```js
        const sphere = canvas3d.boundingSphereVisible;
        const origin = sphere.center;
        const radius = Math.max(sphere.radius + extraRadius, minRadius);
        const snapshot = canvas3d.camera.getFocus(origin, radius);
        canvas3d.requestCameraReset({ durationMs, snapshot });
```

I noticed that the reset doesn't work the first time after removing a representation; It remains in the previous position instead. I need to trigger the reset twice to make it working properly.

This PR seems to solve the problem for me.

arose:
Hmm, do you subscribe to `canvas3d.commited`? `canvas3d.remove(repr)` is not immediate.

giagitom:
I tried subscribing to commited event, and trigger the camera reset inside of it, but the issue remains

giagitom:
Hmm I never used CodeSandbox before. I tried but was unable to make a reproduction so far.

But we can just use the master branch with exposed canvas3d object in order to check if ```canvas3d.boundingSphereVisible.center``` is updated after removing something. (even updating the structure of a representation allows to see the issue)

I just tried in this way:

- load two structures: 2E4A and 1COK.
- Check the boundingSphereVisible center ( it should be ```[25.058691196258494, 36.19581985512225, 7.846723160022061]``` )
- Than remove something from the selection to change the bounding sphere center.
- Now check again the boundingSphereVisible center and ti should be the same as before (here is the issue)

If you try to reset the zoom from the viewer window it will work (probably because internally it is using ```scene.boundingSphereVisible```). 
The zoom reset will also trigger the update of ```canvas3d.boundingSphereVisible```, and probably this is why I need to trigger the camera reset twice in order to get it in the right place.

dsehnal:
This is expected functionality, the bounding sphere does not reset on removal. Check `shouldResetCamera` logic.

dsehnal:
> > If you want this, perhaps we should add an `alwaysResetCamera` flag to the canvas props or something alone these lines?
> 
> I would prefer having a way to "force" the update of `boundingSphereVisible` only when needed, as I use `manualReset = true` in my implementation and want to be able to control when the camera resets.
> 

Well, in this case the camera should always be reset and maybe you are calling it too soon as Alex pointed out above.

Have you tried putting 

```ts
await Scheduler.immediatePromise();
```

before your reset call? (or just calling it in a `requestAnimationFrame`.

> Regarding the `calculateBoundingSphere` function, shouldn’t it also exclude fully transparent renderables when `onlyVisible` is enabled?

Seems like it should. Guess an oversight on our part.

giagitom:
> Well, in this case the camera should always be reset and maybe you are calling it too soon as Alex pointed out above.

I tried using Scheduler but the issue remains. The thing is that even if I try to remove the representation, wait some time (lets say 10 seconds) and then try to reset the camera, the issue is still present. When triggering the camera reset the second time it works as expected instead.

giagitom:
So I have just found that if I execute ```canvas3d.requestCameraReset();``` without any parameters it resets the camera as expected.

PR #1378, opened 2024-12-07
Add ModelWithCoordinates transform

dsehnal:
# Description

Adds new `ModelWithCoordinates` transform that allows modifying existing model coordinates.

## Actions


arose (approved):
Looks good!

Only thing that would be nice, is a frame index and count in the transform params - and use that to with TrajectoryInfo.set. With that the trajectory-index color theme would work.

hainm:
@arose is it possible to make a release for this so I can start working on nglview. thanks.

PR #1392, opened 2024-12-19
Sequence panel focus marker

midlik:
# Description

This adds a blue marker to residues in sequence panels which are focused (i.e. click and show balls and sticks).

Also fixes a bug in selecting in SelectionMode (select 1-50 then select 40-60 -> final selection 1-39 + 41-60).

## Actions

### DIFF
--- a/src/mol-util/marker-action.ts
+++ b/src/mol-util/marker-action.ts
@@ -4,7 +4,7 @@
  * @author Alexander Rose <alexander.rose@weirdbyte.de>
  */
 
-import { OrderedSet, Interval } from '../mol-data/int';
+import { Interval, OrderedSet } from '../mol-data/int';
 import { BitFlags } from './bit-flags';
 import { assertUnreachable } from './type-helpers';
 
@@ -15,7 +15,9 @@ export enum MarkerAction {
     Select = 0x4,
     Deselect = 0x8,
     Toggle = 0x10,
-    Clear = 0x20
+    Clear = 0x20,
+    Focus = 0x40,
+    RemoveFocus = 0x80,
 }
 
 export type MarkerActions = BitFlags<MarkerAction>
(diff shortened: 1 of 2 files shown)


src/mol-util/marker-action.ts
```diff
@@ -35,19 +37,36 @@ export namespace MarkerActions {
...
     }
 }
 
-export function setMarkerValue(array: Uint8Array, status: 0 | 1 | 2 | 3, count: number) {
+/** Bitflags for 3 types of marker (1 = highlighted, 2 = selected, 4 = focused) */
+export type MarkerValue = 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7;
+export const MarkerValue = {
+    Highlighted: 1,
+    Selected: 2,
+    Focused: 4,
```
dsehnal, line 63: @arose do you think we could find a use case for the focused stated in 3D visualization as well?
Otherwise I've suggested to Adam to implement this login only for the Sequence component.

midlik: i found a way to treat focus markers separate, but in an efficient way

arose (commented):
looks generally good, how about using a css class for the focus and use `font-weight: bold;` for the default style?

midlik:
Thank you @arose ,
is this what you have in mind?:
 
seq-focus-bold
seq-focus-bold-underline

(sending a version with bold and bold+underline)

midlik:
> looks generally good, how about using a css class for the focus and use `font-weight: bold;` for the default style?

Thank you @arose ,
is this what you have in mind?:

seq-focus-bold
seq-focus-bold-underline

(sending a version with bold and bold+underline)

midlik:
Also, I'd like to suggest a change to how modifier keys work.

Current setting:
Normal Mode: default = select only, Shift = add to selection
Selection Mode: default = add to selection, Shift = extend last selected range

My suggestion:
Normal Mode: default = select only, Ctrl/Cmd = add to selection, Shift = extend last selected range
Selection Mode: default = add to selection, Shift = extend last selected range

This means a more consistent behavior of the Shift key. It is also more similar to how other apps work (Google Docs, MS Word; VSCode uses Alt instead of Ctrl/Cmd).

Please let me know what you think, @dsehnal @arose

dsehnal:
That's a good suggestion. If this change is accompanied by a help text somewhere (top right next to the questionmark icon?) it might work well.

midlik:
LOL, the new modifier key behavior was already described in Help, even though it was not implemented before :D :D :D
Screenshot 2025-01-09 at 10 39 28

arose (approved):
lgtm, thanks @midlik

PR #1397, opened 2024-12-29
add external-structure theme

arose:
# Description

Add external-structure theme that colors any geometry by structure properties.

EMD-45213 and PDB entry 9C5B take 200ms with default parameters on my machine:

9C5B

9C5B (1)

## Actions

### DIFF
--- a/src/mol-theme/color/external-structure.ts
+++ b/src/mol-theme/color/external-structure.ts
@@ -46,8 +46,8 @@ export const ExternalStructureColorThemeParams = {
         'structure-index': PD.Group(StructureIndexColorThemeParams),
     }),
     defaultColor: PD.Color(Color(0xcccccc)),
-    maxDistance: PD.Numeric(8, { min: 0.1, max: 24, step: 0.1 }),
-    takeFirstDistance: PD.Numeric(4, { min: 0.1, max: 12, step: 0.1 }),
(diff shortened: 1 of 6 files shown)


src/mol-theme/color/external-structure.ts
```diff
@@ -0,0 +1,157 @@
...
+
+            const rf = s.lookup3d.first(x, y, z, takeFirstDistance, lookupFirstCtx);
+            if (rf.count > 0) {
+                l.unit = rf.units[0];
+                l.element = l.unit.elements[rf.indices[0]];
+                return styleTheme.color(l, isSecondary);
+            }
+
+            const rn = s.lookup3d.nearest(x, y, z, 1, lookupNearestCtx);
+            if (rn.count === 0) return defaultColor;
```
dsehnal, line 124: I can imagine this could get rather slow. Would it make sense to add a grid mode for this? This would trade memory for linear lookup complexity by preparing a coloring grid ahead of time.

arose: It is quite OK. For my example, PDB entry 9C5B, it takes 200ms with default parameters on my machine. Filling a 3D grid at comparable resolution will take longer. A grid mode could be beneficial if the isolevel needs to be changed often and smoothly.

arose:
EMD-22345 and 7JIL

image

dsehnal (approved):

PR #1405, opened 2025-01-07
Consider fully transparent renderables as not visible inside Scene

giagitom:
# Description

Consider fully transparent renderables as not visible inside Scene (except for MarkerAverage calculation) 

## Actions


arose:
Hmm, it is by design to not include opacity/transparency. Can you set the visibility yourself instead?

giagitom:
Ok, no problem.

PR #1407, opened 2025-01-08
Rebrand PDB-Dev as PDB-IHM

sbittrich:
# Description
wwPDB is rebranding PDB-Dev as PDB-IHM (https://pdb-ihm.org).
This PR renames UI options and constants/functions to the new PDB-IHM name.
pdb-dev.wwpdb.org now redirects to pdb-ihm.org. These changes also ensure that data is fetched from the correct location.

## Actions

### DIFF
--- a/src/apps/mesoscale-explorer/ui/states.tsx
+++ b/src/apps/mesoscale-explorer/ui/states.tsx
@@ -211,15 +211,15 @@ export async function loadPdb(ctx: PluginContext, id: string) {
     await createHierarchy(ctx, data.ref);
 }
 
-export async function loadPdbDev(ctx: PluginContext, id: string) {
+export async function loadPdbIhm(ctx: PluginContext, id: string) {
     await reset(ctx);
     let url: string;
     // 4 character PDB id, TODO: support extended PDB ID
     if (id.match(/^[1-9][A-Z0-9]{3}$/i) !== null) {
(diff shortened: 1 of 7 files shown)


sbittrich:
Let me know if you'd like more or less breaking changes. E.g., I see little reason to keep the deprecated `pdb-dev` URL param around.

The NPC remote state example also broke with the URL move from pdb-dev.wwpdb.org to pdb-ihm.org.

src/apps/mesoscale-explorer/ui/states.tsx
```diff
@@ -211,15 +211,15 @@ export async function loadPdb(ctx: PluginContext, id: string) {
     await createHierarchy(ctx, data.ref);
 }
 
-export async function loadPdbDev(ctx: PluginContext, id: string) {
+export async function loadPdbIhm(ctx: PluginContext, id: string) {
```
dsehnal, line 214: How about leaving both functions for backwards compat for now and remove in 5.0?

arose (changes requested):
looks good but I agree with @dsehnal that we should keep the existing function in the 2 apps and tag them as @deprecated (and remove in v5).

src/apps/mesoscale-explorer/ui/states.tsx
sbittrich: Sounds good, updated the code to keep both functions around while marking them as deprecated as suggested.

PR #1411, opened 2025-01-13
Quick styles

midlik:
# Description

- Reorganized the Quick Styles section in UI so that representation type is independent of stylized look.

    before
    Screenshot 2025-01-13 at 13 57 21
    after
    Screenshot 2025-01-13 at 13 23 35

- Added `overrideWater` parameter to entity-id color theme so we can have consistent water color across structures (hope it makes sense).

- Tweaked style of toggle buttons so on/off state is distinguishable also when button is hovered.

    before (on/off)
    before-off
    after (on/off)
    after-off

- Increased scrollbar visibility.

    before
    Screenshot 2025-01-13 at 14 42 36
    after
    Screenshot 2025-01-13 at 14 42 46

## Actions


src/mol-plugin-ui/structure/quick-styles.tsx
```diff
@@ -31,110 +33,131 @@ export class StructureQuickStylesControls extends CollapsableControls {
...
+}
+
+
+export class QuickStyles extends PurePluginUIComponent<{}, QuickStylesState> {
+    state: QuickStylesState = { props: { preset: '', stylized: false }, busy: false };
+
+    componentDidMount() {
+        const invalidatePreset = () => this.setState(old => ({ props: { ...old.props, preset: '' } }));
+        this.subscribe(this.plugin.state.data.events.cell.created, invalidatePreset);
+        this.subscribe(this.plugin.state.data.events.cell.stateUpdated, invalidatePreset);
```
dsehnal, line 56: Can you filter this by `isAnimating` and I think `changed` event should be enough? Something like:

```ts
       this.subscribe(plugin.state.data.events.changed, e => {
            if (e.inTransaction || plugin.behaviors.state.isAnimating.value) return;
            this.sync(true);
        });

        this.subscribe(plugin.behaviors.state.isAnimating, isAnimating => {
            if (!isAnimating && !plugin.behaviors.state.isUpdating.value) this.sync(true);
        });
```

The constant UI updates could reduce the FPS quite a bit during animation.

src/mol-plugin-ui/structure/quick-styles.tsx
```diff
@@ -31,110 +33,131 @@ export class StructureQuickStylesControls extends CollapsableControls {
...
+                    toggle={() => this.changePreset('default')} isSelected={this.state.props.preset === 'default'} disabled={this.state.busy} />
+                <ToggleButton label='Cartoon' title='Applies cartoon polymer + ball-and-stick ligand representation preset'
+                    toggle={() => this.changePreset('cartoon')} isSelected={this.state.props.preset === 'cartoon'} disabled={this.state.busy} />
+                <ToggleButton label='Spacefill' title='Applies spacefill representation preset'
+                    toggle={() => this.changePreset('spacefill')} isSelected={this.state.props.preset === 'spacefill'} disabled={this.state.busy} />
+                <ToggleButton label='Surface' title='Applies molecular surface representation preset'
+                    toggle={() => this.changePreset('surface')} isSelected={this.state.props.preset === 'surface'} disabled={this.state.busy} />
+            </div>
+            <ParameterControls params={QuickStyleParams} values={this.state.props} onChangeValues={props => this.changeValues(props)} isDisabled={this.state.busy} />
```
dsehnal, line 160: Here I wonder if this should be a toggle. Perhaps instead add two buttons 

image

This way you can actually remove the `componentDidMount` logic. And even thinking about that: when you click, it triggers the change, meaning the `isSelected={this.state.props.preset === '...'}` will never be applied.

So perhaps the UI can further be updated to

image

src/mol-plugin-ui/structure/quick-styles.tsx
midlik: OK, i changed it.

src/mol-plugin-ui/structure/quick-styles.tsx
midlik: After discussion with @dsehnal , I changed the Quick Styles UI to this:

Screenshot 2025-01-17 at 11 21 09

(It doesn't show what the current representation or style is, as that's tricky to keep track of.)

dsehnal (approved):
LGTM

@arose  any comments before merging this?

arose (approved):

arose:
> LGTM
> 
> @arose any comments before merging this?

good to go, thanks

Issue #1116, opened 2024-05-05
Split-up large draw calls


arose:
Split-up large draw calls to prevent Firefox from crashing, see https://bugzilla.mozilla.org/show_bug.cgi?id=1849433

arose:
Handled with 12c71dc5ba206c9d17690a25ff96d782bc14fd7d

PR #1413, opened 2025-01-14
Lammpstraj dev

corredD:
# Description
A proposal to overcome the. google chrome string length limitation of 512Mb. I changed the type of the format to binary to force using readAsArrayBuffer, and pass the data as Uint8Array to the parser, which will parse line by line the byteArray. Not the best way but I am assuming that one frame will not be bigger that the string length limit. 
Also I put back the fake asym_id, but maybe we can make it depends on the size of the system. This make the playback much much smoother. 

## Actions


corredD:
I should have mention this is only for the lammps trajectory data open files options in the trajectory category. And if you load only this file there is no information of bonds. The bonds information comes from the lammps data file. 
image
I will test with the data file and the bonds to see how it looks.

giagitom:
Oh.. OK thanks, now I understand. 
I always use the trajectory importer using topology and coordinates files.


> I will test with the data file and the bonds to see how it looks.

In my system I find ball and sticks not working well. First time creating the representation it is "fast". Next times updating the representation parameters or structure leads to freezing the system for a few minutes. 

A question: do we already have a way to toggle on/off unrealistic bonds derived from periodic boundary conditions?

corredD:
I don't see any bonds in firefox, but I see bonds when I do OpenFile and load only the .data file, I need to dig to understand whats going on.
Currently the parsing is really naive, so there is unrealistic bonds.

corredD:
Maybe the Lammps data should be handle as Topology and not as Trajectory. I can see that when using Topology there is code to handle the bonds. Also I am not sure why I see them in chrome and not firefox when using Open files, and nothing when using Load Trajectory. @arose any hint to help ?

arose:
I can have a look over the weekend, but the general approach seems right.

giagitom:
> Also I am not sure why I see them in chrome and not firefox when using Open files

This may be the issue with Firefox I suspect. 
image

corredD:
@giagitom also this issue is only in Firefox, not in chrome or safari. Since I added a fix for the periodic bonds, there is two remaining issue, when using Load Trajectory, there is no bond displayed at all, when using Open Files, there is bonds in chrome and safari but not firefox possibly due to the max indexCount. @arose a big thanks if you can look at it over the week end !

src/mol-io/reader/lammps/traj/parser.ts
```diff
@@ -105,63 +105,127 @@ async function handleAtoms(state: State, count: number, parts: string[]): Promis
...
+        }
+    }
+
+    // If no newline is found, decode the remaining part
+    if (nextIndex === currentIndex) {
+        line = decoder.decode(data.subarray(currentIndex + 1), { stream: false });
+        nextIndex = data.length; // Set the index to the end of the data
+    }
+
+    return [line.trim(), nextIndex];
```
dsehnal, line 225: Given how big the files can be, this allocation and the subsequent deconstructions are very expensive. Instead of this I would suggest passing a "result array (or object)" as an argument and adding to it here.

Also, you are calling the trim twice.

src/mol-io/reader/lammps/traj/parser.ts
```diff
@@ -105,63 +105,127 @@ async function handleAtoms(state: State, count: number, parts: string[]): Promis
...
+function readNextLine(
+    data: Uint8Array,
+    currentIndex: number,
+    decoder: TextDecoder
+): [line: string, nextIndex: number] {
+    let line = '';
+    let nextIndex = currentIndex;
+
+    for (let i = currentIndex; i < data.length; i++) {
+        if (data[i] === 10) { // ASCII code for '\n'
```
dsehnal, line 212: Can there be /r in the lines too?

src/mol-model-formats/structure/lammps-data.ts
```diff
@@ -97,10 +101,58 @@ async function getModels(mol: LammpsDataFile, ctx: RuntimeContext, unitsStyle: U
...
             );
+            IndexPairBonds.Provider.set(first, pairBonds);*/
+            // Assuming 'atoms' contains the coordinates after wrapping
+            const newBonds = [];
+            for (let i = 0; i < bonds.count; i++) {
+                const [atom1Index, atom2Index] = [indexA.value(i), indexB.value(i)];
+
+                // Here, you would ideally have access to the wrapped positions of atom1 and atom2
+                // If not, you might need to apply wrapping logic here:
+                const pos1 = [atoms.x.value(atom1Index), atoms.y.value(atom1Index), atoms.z.value(atom1Index)];
```
dsehnal, line 117: The array allocation here is not necessary, can use a single array and reassign it.

dsehnal (commented):
Few performance tips.

corredD:
Thanks @dsehnal ! I tried to follow your tips! About the performance, I get playback performance when loading only the lammpstraj directly in OpenFiles, compare to loading the data file with the lammpstraj...
About the end of line, cant we assume there will always be a \n? I think lammps can write file with \r\n if run on windows system, but I think its not that usual.

dsehnal:
It depends on whether trim removes the \r. Maybe can still check if data[i-1]===\r and decrement the end index in that case.

giagitom:
> @giagitom also this issue is only in Firefox, not in chrome or safari. Since I added a fix for the periodic bonds, there is two remaining issue, when using Load Trajectory, there is no bond displayed at all, when using Open Files, there is bonds in chrome and safari but not firefox possibly due to the max indexCount. @arose a big thanks if you can look at it over the week end !

Thanks for fixing periodic bonds!

arose:
> > Also I am not sure why I see them in chrome and not firefox when using Open files
> 
> This may be the issue with Firefox I suspect. image

This is due to #1116. I will have a look if we can do it in the graphics layer automatically as needed.

src/mol-model-formats/structure/lammps-data.ts
```diff
@@ -32,23 +32,27 @@ async function getModels(mol: LammpsDataFile, ctx: RuntimeContext, unitsStyle: U
...
         offset++;
     }
 
     const MOL = Column.ofConst('MOL', count, Column.Schema.str);
-    const asym_id = Column.ofLambda({
-        value: (row: number) => atoms.moleculeId.value(row).toString(),
-        rowCount: count,
-        schema: Column.Schema.str,
-    });
+    const asym_id = (count > 200000) ? Column.ofConst('A', count, Column.Schema.str)
```
arose, line 49: can you please make this a parameter like unitsStyle?

src/mol-model-formats/structure/lammps-data.ts
```diff
@@ -97,10 +101,64 @@ async function getModels(mol: LammpsDataFile, ctx: RuntimeContext, unitsStyle: U
             const key = bonds.bondId;
             const order = Column.ofConst(1, bonds.count, Column.Schema.int);
             const flag = Column.ofConst(BondType.Flag.Covalent, bonds.count, Column.Schema.int);
+            /*
```
arose, line 104: please remove code that is not needed anymore here and elsewhere

arose:
> > > Also I am not sure why I see them in chrome and not firefox when using Open files
> > 
> > 
> > This may be the issue with Firefox I suspect. image
> 
> This is due to #1116. I will have a look if we can do it in the graphics layer automatically as needed.

will have a fix for it

arose:
> > > > Also I am not sure why I see them in chrome and not firefox when using Open files
> > > 
> > > 
> > > This may be the issue with Firefox I suspect. image
> > 
> > 
> > This is due to #1116. I will have a look if we can do it in the graphics layer automatically as needed.
> 
> will have a fix for it

handled with 12c71dc5ba206c9d17690a25ff96d782bc14fd7d

giagitom:
Hi @arose ,

I tested this on master. The warning message no longer appears in Firefox, but the sticks still don’t show up. Only the sticks connected to the first element seem to be displayed for some reason.

src/mol-model-formats/structure/lammps-data.ts
corredD: you mean like a boolean flag, called for instance "asymId": true | false ?

corredD:
I got it working using Open Files, @arose the fix in the render-items didnt handle the case of using mdbDataList. I added it, but would you prefer this to be in a different PR? 

For the bonds, it still doesnt work with the Load Trajectory. There is something different in the processing of the model file, I dont know if its ModelFromTrajectory or TrajectoryFromModelAndCoordinates in structure.ts. The bonds are parsed, but are lost at some point.

arose:
> @arose the fix in the render-items didnt handle the case of using mdbDataList. I added it, but would you prefer this to be in a different PR?

I see, there was an issue in the logic when to cull, should be fixed with f5b5109d0fa365697bf60623028fa520e8637f51.

I would like to avoid having to handle mdbDataList wit respect to the FF limit because it is not as straightforward for multiDrawElementsInstancedBaseVertexBaseInstance and multiDrawArraysInstancedBaseInstance.

src/mol-model-formats/structure/lammps-data.ts
arose: yes, make it a boolean parameter here. In CoordinatesFromLammpstraj & TrajectoryFromLammpsTrajData (src\mol-plugin-state\transforms\model.ts) you can do a param with 3 options: on, off, auto. For auto you can use a hardcoded value (for now, but we can make it easily configurable because it is a the plugin level)

src/mol-model/structure/model/model.ts
```diff
@@ -122,7 +122,10 @@ export namespace Model {
                 const symmetry = ModelSymmetry.fromCell(f.cell.size, f.cell.anglesInRadians);
                 ModelSymmetry.Provider.set(m, symmetry);
             }
-
+            if (oldIndex) {
+                // can I update here the indexPair Bond to take in account the periodicity?
```
corredD, line 126: Is this were the periodicity of the bonds be handled ? or elsewhere ?

corredD:
I added the options for the asymId information in the model, as well as adding the bonds when loading both the .data. and the lammpstraj if any. But the perdiocity error on the bounds appeared in the remaining frame ( see the other comment ).

arose:
What is a good example to try and how to load it?

src/mol-model/structure/model/model.ts
arose: How about setting a maxDistance when creating the IndexPairBonds and then use the dynamicBonds param for the structure

corredD:
I am using [NPC-NUP98_Capsid.data](https://drive.google.com/file/d/1LreAjHmieI-sWGvq589lxp82ze5Zu5iV/view?usp=sharing) that has bond information, with [small5.lammpstrj](https://drive.google.com/file/d/1N1oAI9EyEZhnhBTJMx9Ti-XzrgzNUK_4/view?usp=sharing) to get a small trajectory system with it. I didnt know about the dynamicBonds param, the maxDistance can definietely be set using the bounding box size. Can you show how to use the dynamicBonds param ?

PR #1417, opened 2025-01-17
MVS: Initial support for customizable representation parameters

dsehnal:
# Description

Adds the option to customize representation parameters using MolViewSpec. 

@midlik Good job on the union param spec, worked out of the box.

## Actions


src/extensions/mvs/tree/mvs/mvs-representations.ts
```diff
@@ -0,0 +1,39 @@
+/**
```
midlik, line 1: Please rename this file to "mvs-tree-representations.ts", to keep consistency with "mvs-tree-primitives.ts"

src/extensions/mvs/load-helpers.ts
```diff
@@ -285,19 +285,21 @@ export function componentFromXProps(node: MolstarNode<'component_from_uri' | 'co
...
 export function representationProps(node: MolstarSubtree<'representation'>): Partial<StateTransformer.Params<StructureRepresentation3D>> {
     const alpha = alphaForNode(node);
-    switch (node.params.type) {
+    const params = node.params;
+    switch (params.type) {
         case 'cartoon':
             return {
-                type: { name: 'cartoon', params: { alpha } },
+                type: { name: 'cartoon', params: { alpha, tubularHelices: params.tubular_helices } },
+                sizeTheme: { name: 'uniform', params: { value: params.size_factor ?? 1 } },
```
midlik, line 293: There is a mechanism responsible for adding default values for optional fields. So `?? 1` is superfluous here. (the same for other repr types below).

src/extensions/mvs/load-helpers.ts
```diff
@@ -285,19 +285,21 @@ export function componentFromXProps(node: MolstarNode<'component_from_uri' | 'co
...
             };
         case 'ball_and_stick':
             return {
-                type: { name: 'ball-and-stick', params: { sizeFactor: 0.5, sizeAspectRatio: 0.5, alpha } },
+                type: { name: 'ball-and-stick', params: { sizeFactor: (params.size_factor ?? 1) * 0.5, sizeAspectRatio: 0.5, alpha, ignoreHydrogens: params.ignore_hydrogens } },
```
midlik, line 297: Here I'm thinking whether we should:
A) keep MVS `size_factor` aligned with Molstar `sizeFactor`, i.e. have `sizeFactor: params.size_factor` here with default 0.5 for `size_param`,
or 
B) let MVS `size_factor` be different from Molstar `sizeFactor`, i.e. have `sizeFactor: params.size_factor * 0.5` here with default 1 for `size_param` (current state).

The nice thing about B is that default is 1, so it's kinda intuitive for the user (>1 means bigger that default, <1 means smaller than default). Also the default value is the same for all repr types (in A it is different for different repr types). But it brings the complexity of having `size_factor` that means something different than Molstar's `sizeFactor`.

Also, does `sizeFactor: 1` mean that the balls have radius 1 Angstrom? If so, then maybe I would go for solution A so that `size_param` is also in Angstroms rather that arbitrary units. And maybe we could rename it to `ball_radius` so then we don't have different default values for the same parameter in different repr types.

What do you think @dsehnal ?

BTW there is nothing special about the original value `sizeFactor: 0.5`, I just thought it looked fairly good, when I was implementing this.

midlik (commented):
Thank you, looks nice, see comments. 
I will be creating some MVSs for our use-cases soon, so I will then add more params as needed.

PR #1421, opened 2025-01-20
add volume dot representation

arose:
# Description

Add volume dot representation and volume-value size theme

image

## Actions

### DIFF
--- a/src/mol-repr/volume/dot.ts
+++ b/src/mol-repr/volume/dot.ts
@@ -1,0 +1,274 @@
+/**
+ * Copyright (c) 2025 mol* contributors, licensed under MIT, See LICENSE file for more info.
+ *
+ * @author Alexander Rose <alexander.rose@weirdbyte.de>
+ */
+
+import { ParamDefinition as PD } from '../../mol-util/param-definition';
+import { Grid, Volume } from '../../mol-model/volume';
+import { VisualContext } from '../visual';
+import { Theme, ThemeRegistryContext } from '../../mol-theme/theme';
+import { Mesh } from '../../mol-geo/geometry/mesh/mesh';
+import { VolumeVisual, VolumeRepresentation, VolumeRepresentationProvider, VolumeKey } from './representation';
+import { VisualUpdateState } from '../util';
+import { RepresentationContext, RepresentationParamsGetter, Representation } from '../representation';
+import { PickingId } from '../../mol-geo/geometry/picking';
+import { EmptyLoci, Loci } from '../../mol-model/loci';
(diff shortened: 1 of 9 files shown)


src/mol-repr/volume/dot.ts
```diff
@@ -0,0 +1,274 @@
...
+
+    const p = Vec3();
+    const [xn, yn, zn] = space.dimensions;
+
+    const count = Math.ceil((xn * yn * zn) / 10);
+    const builder = SpheresBuilder.create(count, Math.ceil(count / 2), spheres);
+
+    for (let z = 0; z < zn; ++z) {
+        for (let y = 0; y < yn; ++y) {
+            for (let x = 0; x < xn; ++x) {
```
dsehnal, line 108: I wonder if we should add `Tensor.forEach` which would iterate based on actual axis ordering of the underlying data (slowest to fastest changing axis). This would prevent a lot of cache misses, although the drawback would be passing a callback for the inner body, which might erase the gains...

src/mol-theme/size/volume-value.ts
```diff
@@ -0,0 +1,62 @@
...
+            granularity: 'group',
+            size,
+            props,
+            description: Description
+        };
+    } else {
+        return {
+            factory: VolumeValueSizeTheme,
+            granularity: 'uniform',
+            size: () => 1 * props.scale,
```
dsehnal, line 47:
```suggestion
            size: () => props.scale,
```

src/mol-theme/size/volume-value.ts
```diff
@@ -0,0 +1,62 @@
...
+export function getVolumeValueSizeThemeParams(ctx: ThemeDataContext) {
+    return VolumeValueSizeThemeParams; // TODO return copy
+}
+
+export function VolumeValueSizeTheme(ctx: ThemeDataContext, props: PD.Values<VolumeValueSizeThemeParams>): SizeTheme<VolumeValueSizeThemeParams> {
+    if (ctx.volume) {
+        const { data } = ctx.volume.grid.cells;
+
+        const size: LocationSize = (location: Location): number => {
+            if (Volume.Cell.isLocation(location)) {
```
dsehnal, line 29:
```suggestion
        const isLocation = Volume.Cell.isLocation;
        const size: LocationSize = (location: Location): number => {
            if (isLocation(location)) {
```

dsehnal (approved):
Nice.

src/mol-repr/volume/dot.ts
arose: made  #1426

PR #1433, opened 2025-02-02
MVS: IHM support

dsehnal:
# Description

Adds better support for I/HM structure in MolViewSpec

- [x] Support `coarse` components
- [x] Support `spacefill` representation
- [x] Support `element_granularity` in component expressions
- [x] If no color theme is specified explicitly, use Mol*'s defaults instead
- [x] Add `atom.ihm.has-seq-id` symbol to the query language
- [x] Add ihm-restraints example

https://pdb-ihm.org/cif/8zz1.cif with crosslink restraints:

image

## Actions

### DIFF
--- a/src/extensions/mvs/helpers/selections.ts
+++ b/src/extensions/mvs/helpers/selections.ts
@@ -282,11 +282,7 @@ export function rowToExpression(row: MVSAnnotationRow): Expression {
 
     const residueTests: Expression[] = [];
     if (isDefined(row.label_seq_id)) {
-        if (row.element_granularity === 'coarse') {
-            residueTests.push(ihm.hasSeqId({ 0: row.label_seq_id }));
-        } else {
-            residueTests.push(eq([macromolecular.label_seq_id(), row.label_seq_id]));
-        }
+        residueTests.push(ihm.hasSeqId({ 0: row.label_seq_id }));
     }
     if (isDefined(row.auth_seq_id)) residueTests.push(eq([macromolecular.auth_seq_id(), row.auth_seq_id]));
     if (isDefined(row.pdbx_PDB_ins_code)) residueTests.push(eq([macromolecular.pdbx_PDB_ins_code(), row.pdbx_PDB_ins_code]));
--- a/src/extensions/mvs/helpers/selections.ts
+++ b/src/extensions/mvs/helpers/selections.ts
@@ -1,7 +1,8 @@
 /**
- * Copyright (c) 2023 mol* contributors, licensed under MIT, See LICENSE file for more info.
+ * Copyright (c) 2023-2025 mol* contributors, licensed under MIT, See LICENSE file for more info.
  *
  * @author Adam Midlik <midlik@gmail.com>
+ * @author David Sehnal <david.sehnal@gmail.com>
  */
 
 import { Column } from '../../../mol-data/db';
@@ -262,7 +263,7 @@ function matchesRange<T>(requiredMin: T | undefined | null, requiredMax: T | und
 export function rowToExpression(row: MVSAnnotationRow): Expression {
     const { and } = MS.core.logic;
     const { eq, gre: gte, lte } = MS.core.rel;
-    const { macromolecular } = MS.struct.atomProperty;
+    const { macromolecular, ihm } = MS.struct.atomProperty;
     const propTests: Partial<Record<string, Expression>> = {};
 
     if (isDefined(row.label_entity_id)) {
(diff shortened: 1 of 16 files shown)


src/extensions/mvs/load-helpers.ts
```diff
@@ -316,20 +322,22 @@ export function alphaForNode(node: MolstarSubtree<'representation'>): number {
...
+export function colorThemeForNode(node: MolstarSubtree<'color' | 'color_from_uri' | 'color_from_source' | 'representation'> | undefined, context: MolstarLoadingContext): StateTransformer.Params<StructureRepresentation3D>['colorTheme'] | undefined {
     if (node?.kind === 'representation') {
         const children = getChildren(node).filter(c => c.kind === 'color' || c.kind === 'color_from_uri' || c.kind === 'color_from_source') as MolstarNode<'color' | 'color_from_uri' | 'color_from_source'>[];
         if (children.length === 0) {
-            return {
-                name: 'uniform',
-                params: { value: decodeColor(DefaultColor) },
-            };
+            // If no color child node is specified, use Mol*'s default
+            return undefined;
```
midlik, line 330: Here we are making sure MVS is pretty much Molstar-specific and no one will ever implement it for any other viewer.

src/extensions/mvs/helpers/selections.ts
```diff
@@ -280,7 +281,13 @@ export function rowToExpression(row: MVSAnnotationRow): Expression {
...
 
     const residueTests: Expression[] = [];
-    if (isDefined(row.label_seq_id)) residueTests.push(eq([macromolecular.label_seq_id(), row.label_seq_id]));
+    if (isDefined(row.label_seq_id)) {
+        if (row.element_granularity === 'coarse') {
+            residueTests.push(ihm.hasSeqId({ 0: row.label_seq_id }));
+        } else {
+            residueTests.push(eq([macromolecular.label_seq_id(), row.label_seq_id]));
+        }
+    }
```
midlik, line 290: Does this mean that if `label_seq_id` is not present, it will select from both atomic and coarse parts of the structure?

midlik (commented):
Looks fine but some things are not clear to me - see comments.
Also, have you tested if the new selections work in all possible contexts (I mean, as inline component, as *_from_uri, as target for primitive...)?

src/extensions/mvs/load-helpers.ts
dsehnal: I disagree. You can still implement explicit colors. And this could be defined as expected behavior (= colors determined by viewer if not specified explicitly).

src/extensions/mvs/helpers/selections.ts
dsehnal: Actually, we can completely remove `element_granularity` in this case and just use the `ihm.hasSeqId`.

src/extensions/mvs/load-helpers.ts
dsehnal: I have an alternate solution to support "vendor extension" to do that. Thanks for the feedback.

dsehnal:
> Also, have you tested if the new selections work in all possible contexts (I mean, as inline component, as *_from_uri, as target for primitive...)?

My goal was to make this work with the primitives here. @midlik you wrote this code, so do I need to change anything else for things you mention to work or is this fine? Can also do this in a separate PR.

src/extensions/mvs/load-helpers.ts
midlik: I thought the original idea was that a MVS view should look roughly the same in all viewers. 
With this, a representation node without specified color can look very different in different viewers.
So as a user, I might think I have created a beautiful view where each chain has a different color, but then I load it in PyMOL it will be all magenta because default per-chain coloring was just Molstar's implementation detail.

src/extensions/mvs/helpers/selections.ts
midlik: Hmmm, if we can remove `element_granularity`, let's do it. One thing less for the user to worry about.

midlik:
> > Also, have you tested if the new selections work in all possible contexts (I mean, as inline component, as *_from_uri, as target for primitive...)?
> 
> My goal was to make this work with the primitives here. @midlik you wrote this code, so do I need to change anything else for things you mention to work or is this fine? Can also do this in a separate PR.

If we remove `element_granularity`, it should work in all context, i think.
If you want i can test it (just ping me once you remove `element_granularity`)

dsehnal:
> If we remove `element_granularity`, it should work in all context, i think.
> If you want i can test it (just ping me once you remove `element_granularity`)

@midlik 
Ready for re-review/testing. Thanks

midlik:
I'm getting errors when trying to use `beg_label_seq_id` and `end_label_seq_id`:
```
p.distance({ radius: 1, start: { label_asym_id: 'A', beg_label_seq_id: 607, end_label_seq_id: 609 }, end: { label_asym_id: 'A', label_seq_id: 58 } }); // atomic-atomic
```

even though the residues i'm targeting are atomic, not coarse.

I think this could be solved by implementing something like `MolScript.structureQuery.atomProperty.ihm.hasOverlapWithSeqIdRange` (pls give it a better name), which would take start and end argument (both can be omitted) instead of just one seqId.

midlik:
And getting the same kind of errors with any auth_* fields (auth_asym_id, auth_seq_id).

Not sure if these can be implemented for coarse models, but at least they should keep working for the atomic portion of the structure, while returning an empty selection from the coarse portion (+ print warning), IMHO.

midlik:
This completely breaks the repr, instead of just not coloring:
```
const repr = struct.component().representation({ type: 'spacefill' });
repr.color({ color: 'blue', selector: { auth_asym_id: 'C' } });
```

dsehnal:
> This completely breaks the repr, instead of just not coloring:
> 
> ```
> const repr = struct.component().representation({ type: 'spacefill' });
> repr.color({ color: 'blue', selector: { auth_asym_id: 'C' } }); // breaks everything
> ```
> 
> ```
> repr.color({ color: 'blue', selector: { label_asym_id: 'A', label_seq_id: 81 } }); // works nicely
> repr.color({ color: 'blue', selector: { label_asym_id: 'A', label_atom_id: 'XXXX' } }); // colors whole chain :(, should color nothing
> ```

This is expected behavior currently and would break in "native" Mol* code, will possibly adjust in the future.

PR #1416, opened 2025-01-17
Molstar Progressive Web Apps

agdturner:
# Description
Mol* things installable as Progressive Web Apps (PWAs).

This follows a discussion:
https://github.com/molstar/molstar/discussions/1390

PWAs are explained here:
https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps
https://en.wikipedia.org/wiki/Progressive_web_app

There are distinct PWAs for viewer, docking-viewer, mesoscale-explorer and some examples. Two scripts (pwa.js and pwa-clean.js) automate the code generation (pwa.js) and clean up (pwa-clean.js). For each PWA, an index-pwa.html is generated from the original index.html. The respective index.ts and index.tsx files were updated to add reference to these additional html files. The service worker and manifest files are (re)generated as part of the build.

Examples of the PWAs are linked below:
https://agdturner.github.io/molstar/build/viewer/index-pwa.html
https://agdturner.github.io/molstar/build/docking-viewer/index-pwa.html
https://agdturner.github.io/molstar/build/mesoscale-explorer/index-pwa.html
https://agdturner.github.io/molstar/build/examples/alpha-orbitals/index-pwa.html
https://agdturner.github.io/molstar/build/examples/basic-wrapper/index-pwa.html
https://agdturner.github.io/molstar/build/examples/lighting/index-pwa.html
https://agdturner.github.io/molstar/build/examples/proteopedia-wrapper/index-pwa.html

The above PWAs are served from GitHub Pages. Once installed by a user as a PWA, they can work offline to use some Mol* functionality.

## Actions

- [y] Added description of changes to the `[Unreleased]` section of `CHANGELOG.md`
- [n] Updated headers of modified files (I don't think there was anything that should have been modified...)
- [y] Added my name to `package.json`'s `contributors`
- [n] (Optional but encouraged) Improved documentation in `docs` (Maybe there is something wanted here, but this is essentially only a sort of wrapper around the existing offering...)

PR #1436, opened 2025-02-07
Pwa

agdturner:
# Description
This replaces the following Pull Request:
https://github.com/molstar/molstar/pull/1416/ for a PWA

It seemed better/easier to start afresh than remove all the unwanted parts and this has improved things further.

@arose is in the process of reviewing. I hope this has made things easier/better.

## Actions

- [x ] Added description of changes to the `[Unreleased]` section of `CHANGELOG.md`
- [x ] Updated headers of modified files
- [x ] Added my name to `package.json`'s `contributors`


arose (changes requested):
Thanks @agdturner, I finally found some time to play around with it.

I would like to see the manifest and service-worker (as TypeScript) file in src/apps/viewer instead it being generated with a script. Any code we might want to reuse we can move to src/mol-utils/ as needed. Also, it is OK to add pwa support to the index.html, no need for a separate file.

What are added icons used for - I did not see them in use - or can they be removed?

Regarding file_handlers, we need to add a consumer via launchQueue.setConsumer to handle the incoming file(s) see https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/How_to/Associate_files_with_your_PWA 

Using viewer.loadStructureFromData should work for structures, for other file type we need more viewer helper method but that can be done later.

agdturner:
> Thanks @agdturner, I finally found some time to play around with it.

:)
 
> I would like to see the manifest and service-worker (as TypeScript) file in src/apps/viewer instead it being generated with a script. Any code we might want to reuse we can move to src/mol-utils/ as needed. Also, it is OK to add pwa support to the index.html, no need for a separate file.

OK, I'll work on that... I created a separate index-pwa.html file as this was requested (https://github.com/molstar/molstar/discussions/1390#discussioncomment-11608841)...

> What are added icons used for - I did not see them in use - or can they be removed?

The icons are need for the PWA. They don't have to be as they are, but at least those resolutions are necessary for the PWA to function/be installable.

> Regarding file_handlers, we need to add a consumer via launchQueue.setConsumer to handle the incoming file(s) see https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/How_to/Associate_files_with_your_PWA

Ah, OK, I only added the following for file handling in the PWA manifest:
        "file_handlers": [
            {
                "action": "./index-pwa.html",
                "accept": {
                    "application/vnd.molstar": [".mol", ".mol2", ".sdf", ".pdb", ".mmcif", ".xyz", ".mmtf", ".bcif"],
                    "application/vnd.cif": [".cif"]
                }
            }
        ]

I'm thinking that in the new typescript file something along the following lines is wanted:

// Handle incoming files
if ('launchQueue' in window) {
    launchQueue.setConsumer((launchParams) => {
        if (!launchParams.files.length) {
            return;
        }
        for (const fileHandle of launchParams.files) {
            // Process the file(s) here
            console.log(`Received file: ${fileHandle.name}`);
        }
    });
}

> Using viewer.loadStructureFromData should work for structures, for other file type we need more viewer helper method but that can be done later.

OK

agdturner:
I have done the work @arose, please review again and let me know any further changes or merge this in.

I've tested via:
https://agdturner.github.io/molstar/build/viewer/index.html

agdturner:
Hold fire @arose, there is a problem: The Service Worker registration is not working as getting the version number from the package.json file is not working. Possible workarounds are to store the version number elsewhere or hard code it...

PR #1453, opened 2025-03-04
Use StructureElement.Bundle for measurements

### LINKED ISSUE #1451: Distance measurements for .cif files are 0
I may be missing something, but this seems like unexpected behavior.

I am trying to perform a distance measurement for PDB-IHM id 8ZZC. I believe this file is a .cif.
When selecting two items on the smallest granularity (atom/course elements), the distance measurement is always zero. 
From my observations, it seems like the measurement is calculated from the center of the model to itself, rather than between the selected elements.

dsehnal:
# Description

- [x] Add `MultiStructureSelectionFromBundle` transform
- [x] Use `StructureElement.Bundle` instead of expressions to serialize measurement elements
- [x] Fixes measurements not being supported for coarse models

Fixes #1451 

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,6 @@ Note that since we don't clearly distinguish between a public and private interf
 ## [Unreleased]
 
 - `ihm-restraints` example: show entity labels
-- Support bundles in `MultiStructureSelectionFromExpression`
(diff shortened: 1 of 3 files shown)


arose (approved):
thanks, lgtm

PR #1454, opened 2025-03-04
For viewer to be installable as a Progressive Web App with a Service …

agdturner:
…Worker transpiled from TypeScript

# Description
This replaces this pull request that I closed as I deleted the branch: https://github.com/molstar/molstar/pull/1436

@arose is in the process of reviewing.

I deleted and created another branch with the same name. Sorry for any confusion. I think this PR is ready to be merged.

Thanks,

Andy

## Actions

- [ x] Added description of changes to the `[Unreleased]` section of `CHANGELOG.md`
- [ x] Updated headers of modified files
- [ x] Added my name to `package.json`'s `contributors`


arose (commented):
Hey @agdturner, I took the liberty make changes to your PR, I hope you don't mind. I had to get familiar with PWAs and try things myself to see what I want in Mol*.

I mainly tried to keep the build process as simple and unobtrusive as possible. Given how simple the SW is, we can forego using TS as that would need an extra tsconfig.

I deployed it at https://molstar.org/dev/pwa/. I can install it with Chrome on Win11 and then double click files in the File Explorer to open them in the Mol* Viewer PWA. I can also install with Chrome on Android but opening files from other apps does not work.

We will need to add support to the new esbuild option to copy new file types around.

agdturner:
Thanks @arose.

Went on a bit of a wild goose chase before realising those exclusions and another tsconfig file was necessary to get the build to work! I think it is a good idea to use a TypeScript instead of JavaScript for the Service Worker, but understand about trying to keep the build process simple.

I had hoped you would change the icons/images. The ones I used were from the Mozilla tutorial I followed when learning about PWAs to use in another project.

Is there anything more for me to do with this PR?

arose:
> Is there anything more for me to do with this PR?

Testing would be great, I did not try other Browser or anything Apple.

src/apps/viewer/index.html
```diff
@@ -35,6 +35,7 @@
             }
         </style>
         <link rel="stylesheet" type="text/css" href="molstar.css" />
+        <link rel="manifest" href="./manifest.webmanifest">
```
dsehnal, line 38: Instead of having this in the default all, would it make sense to instead create a `sw.html` which supports this instead to make it optional?

src/apps/viewer/app.ts
```diff
@@ -560,6 +561,23 @@ export class Viewer {
         }
     }
 
+    async loadFiles(files: File[]) {
```
dsehnal, line 564: the `async` isn't needed as there is no await in the function.

webpack.config.common.js
```diff
@@ -13,31 +14,8 @@ class VersionFilePlugin {
...
-                ]
-            },
-            {
-                test: /\.(jpg)$/i,
-                type: 'asset/resource',
-            },
-        ]
-    },
-    plugins: [
+function getSharedConfig(copyPluginPatterns) {
```
dsehnal, line 17:
```suggestion
function getSharedConfig(options = {}) {
  // and use options.copyPluginPatterns
```

webpack.config.common.js
```diff
@@ -50,40 +28,73 @@ const sharedConfig = {
...
     };
 }
 
-function createEntryPoint(name, dir, out, library) {
+function createEntryPoint(name, dir, out, library, copyPluginPatterns) {
     return {
         entry: path.resolve(__dirname, `lib/${dir}/${name}.js`),
         output: { filename: `${library || name}.js`, path: path.resolve(__dirname, `build/${out}`), library: library || out, libraryTarget: 'umd', assetModuleFilename: 'images/[hash][ext][query]', 'publicPath': '' },
-        ...sharedConfig
+        ...getSharedConfig(copyPluginPatterns)
```
dsehnal, line 97: No need to modify `sharedConfig`. How about something like this?

```ts
const plugins = [...sharedConfig.plugins];
if (copyPluginPatterns) {
  plugins.push(....)
}
```

and 
```suggestion
        ...sharedConfig,
        plugins
```

webpack.config.common.js
```diff
@@ -50,40 +28,73 @@ const sharedConfig = {
...
         target: isNode ? 'node' : void 0,
         entry: path.resolve(__dirname, `lib/${src}.js`),
         output: { filename: `${outFilename}.js`, path: path.resolve(__dirname, `build/${outFolder}`) },
-        ...sharedConfig
+        ...getSharedConfig()
     };
 }
 
-function createEntryPoint(name, dir, out, library) {
+function createEntryPoint(name, dir, out, library, copyPluginPatterns) {
```
dsehnal, line 93:
```suggestion
function createEntryPoint(name, dir, out, library, options = {}) {
```

And use `options.copyPluginPatters`

webpack.config.viewer.js
```diff
@@ -1,6 +1,14 @@
 const common = require('./webpack.config.common.js');
+const VERSION = require('./package.json').version;
 const createApp = common.createApp;
-module.exports = [
-    createApp('viewer', 'molstar'),
-    createApp('mesoscale-explorer', 'molstar'),
-];
\ No newline at end of file
+module.exports = (env, argv) => {
+    return [
```
dsehnal, line 5: It's not enough to change this file. It is also required in:
-  `webpack.config.production.js`
- Since you are adding more import file types to `index.ts` for the viewer, can you please also adjust the recently added `build-dev.mjs` to support this?

dsehnal (commented):
Thanks for figuring out how to do this.

But at least for the initial release, can you make this only work on a specific `sw.html` file instead of on the `index.html` which everyone will use. I would prefer to test this out first at the very least.

arose:
> But at least for the initial release, can you make this only work on a specific `sw.html` file instead of on the `index.html` which everyone will use. I would prefer to test this out first at the very least.

Been thinking about this some more and now I believe it is best to just have this feature in the Viewer deployment on github pages. This would get rid of any build process changes and leave the html as is just add a comment like we do for the analytics code. Happy to make the changes later in the week. We can also note it in the docs to make it easier for other to enable PWA for the molstar app.

dsehnal:
> Been thinking about this some more and now I believe it is best to just have this feature in the Viewer deployment on github pages. This would get rid of any build process changes and leave the html as is just add a comment like we do for the analytics code. Happy to make the changes later in the week. We can also note it in the docs to make it easier for other to enable PWA for the molstar app.

- For development, will the PWA update automatically or will be version have to be incremented on every build?
- Is there some confirmation the user needs to click on installing the app? If I was using a web page for years and suddenly it started to ask me to install/confirm something, I would get suspicious.

arose:
Changed it so it is only added when deploying

> For development, will the PWA update automatically or will be version have to be incremented on every build?

Since it is only added when deploying, normal development is completely unaffected.

> Is there some confirmation the user needs to click on installing the app? If I was using a web page for years and suddenly it started to ask me to install/confirm something, I would get suspicious.

On desktop there is nothing, only a little extra icon in the address bar a user can click to install.

I don't get anything on mobile either, only an install option once I click "Add to Home screen" in the Browser menu

package-lock.json
dsehnal, line 1: We don't need this package lock rebuild as no packages have changed.

scripts/deploy.js
```diff
@@ -9,14 +9,23 @@ const path = require('path');
...
 const remoteUrl = "https://github.com/molstar/molstar.github.io.git";
+const dataDir = path.resolve(__dirname, '../data/');
 const buildDir = path.resolve(__dirname, '../build/');
 const deployDir = path.resolve(buildDir, 'deploy/');
 const localPath = path.resolve(deployDir, 'molstar.github.io/');
 
 const analyticsTag = /<!-- __MOLSTAR_ANALYTICS__ -->/g;
 const analyticsCode = `<!-- Cloudflare Web Analytics --><script defer src='https://static.cloudflareinsights.com/beacon.min.js' data-cf-beacon='{"token": "c414cbae2d284ea995171a81e4a3e721"}'></script><!-- End Cloudflare Web Analytics --><iframe src="https://web3dsurvey.com/collector-iframe.html" style="width: 1px; height: 1px;"></iframe>`;
 
+const manifestTag = /<!-- __MOLSTAR_MANIFEST__ -->/g;
```
dsehnal, line 23: I've only now realized that this deploy script added the tracking to the viewer. I've usually been doing this manually, meaning for some versions of the viewer the analytics wasn't present.

I think we should add a version of this script that just modifies the files, but doesn't do the automatic push to GH Pages. I my case, I sometimes only want to update a particular page or demo for example. I can have a look into this in a separate PR.

data/pwa/sw.js
```diff
@@ -0,0 +1,60 @@
...
+ * @author Alexander Rose <alexander.rose@weirdbyte.de>
+ */
+
+/** version from package.json, to be filled in during deployment */
+const VERSION = '__MOLSTAR_VERSION__';
+
+const CACHE_NAME = `molstar-viewer-${VERSION}`;
+
+// The static resources that the app needs to function.
+const APP_STATIC_RESOURCES = [
```
dsehnal, line 14: Should this cache the background as well or is that an overkill?

dsehnal (approved):
Tested this on Firefox, Chrome and mobile Safari, all work.

data/pwa/sw.js
arose: would keep it simple for now

package-lock.json
arose: it is only line ending changes, don't know how to fix that - doesn't matter

Issue #1458, opened 2025-03-05
Color interpolation not working with PDB uncertainity


luk27official:
I have a .pdb structure that has pLDDT scores in the b-factor fields (in a 0-100 range). I would like to color the residues based on these scores with a custom palette. I am able to do it this way:

```ts
const representation = await plugin.builders.structure.representation.addRepresentation(polymer, {
  type: "ball-and-stick",
  color: "uncertainty",
  colorParams: {
    domain: [0, 100],
    list: {
        kind: "set",
        colors: [Color(0x0053d6), Color(0x65cbf3), Color(0xffdb13), Color(0xff7d45)],
    },
  }
});
```

The problem is that for me, the `kind: "set"` parameter does not work when using `color: "uncertainity"`. It always interpolates the colors, regardless of the setting. This also happens to me when clicking the `Interpolated` setting in the state tree manually. All of this happens to me both when done programatically and manually in the online viewer.

I am attaching two screenshots showing that the colors have not changed, and also the PDB file (as a text file):

![Image](https://github.com/user-attachments/assets/bbebe368-d3f6-4a34-918e-5457da7621bb)
![Image](https://github.com/user-attachments/assets/2856f21a-0084-492a-ae0f-18d9363b5077)

[pdb_example.txt](https://github.com/user-attachments/files/19087304/pdb_example.txt)

Is there something I am doing wrong?

By the way, when playing around with the colors in the online viewer, when using `uncertainity` and I set `Interpolated` to off and I remove all colors from the list, the viewer crashes.

Thanks in advance :)

dsehnal:
So you would like the `set` to behave as categorical colors and assign 0-25 to the 1st color, 25-50 to the 2nd, etc?

> By the way, when playing around with the colors in the online viewer, when using uncertainity and I set Interpolated to off and I remove all colors from the list, the viewer crashes.

Thanks, will have a look if I get some free time.

dsehnal:
The thresholds can already be entered via `export type ColorListEntry = Color | [Color, number /** normalized value from 0 to 1 */]`, just need to normalize to the [0, 1] range, e.g., `list: [[c1, 0], [c2, 0.2], ...]`.

For he coloring, the `interpolate / set` is indeed ignored in https://github.com/molstar/molstar/blob/master/src/mol-theme/color/uncertainty.ts#L38

[...]

PR #1459, opened 2025-03-06
Implement ColorScale.createDiscrete, fix UI color palette bugs

luk27official:
# Description

- Implementation of `ColorScale.createDiscrete`, as discussed in #1458 
- `ColorScale.createDiscrete` used in the `uncertainty` color theme
- Fix color palette shown in the UI (for non-gradient palettes)
- Fix colors description in the UI (when using custom thresholds)
- Fix an edge case in the UI when the user deletes all colors from the color list when interpolation is disabled

Feel free to leave any further comments, I will be glad to improve the code if needed.

## Actions

### DIFF
--- a/src/mol-plugin-ui/controls/parameters.tsx
+++ b/src/mol-plugin-ui/controls/parameters.tsx
@@ -704,6 +704,9 @@ const colorGradientBanded = memoize1((colors: ColorListEntry[]) => {
         if (!off[0]) {
             return 'linear-gradient(to right, #000 0%, #000 100%)';
         }
+        if (off[0][1] > off[off.length - 1][1]) {
+            off.reverse();
+        }
         styles.push(`${Color.toStyle(off[0][0])} ${(100 * off[0][1]).toFixed(2)}%`);
         for (let i = 0, il = off.length - 1; i < il; ++i) {
             const [c0, o0] = off[i];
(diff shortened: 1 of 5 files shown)


src/mol-plugin-ui/controls/parameters.tsx
```diff
@@ -700,6 +701,13 @@ const colorGradientBanded = memoize1((colors: ColorListEntry[]) => {
     const hasOffsets = colors.every(c => Array.isArray(c));
     if (hasOffsets) {
         const off = colors as [Color, number][];
+        // 0 colors present
+        if (!off[0]) {
+            return 'linear-gradient(to right, #000 0%, #000 100%)';
+        }
+        if (off[0][1] > off[off.length - 1][1]) {
+            off.reverse();
```
dsehnal, line 709: This will mutate the `colors` argument. Create a copy in `off`?

dsehnal (commented):
First pass LGTM.

@arose  any objections?

src/mol-util/color/scale.ts
```diff
@@ -100,4 +101,82 @@ export namespace ColorScale {
             get legend() { return ScaleLegend(minLabel, maxLabel, colors); }
         };
     }
+
+    export function createDiscrete(props: ColorScaleProps): ColorScale {
```
arose, line 105: This has a lot of the same code as `ColorScale.create`. Can you find ways so they share more code?

dsehnal: I personally don't mind the copy-paste and small modification. General rule of thumb would be to do this at most 3 times though before refactoring.

luk27official: I have tried my best to make it more generic, check it out now, please.

I have also handled the case when the user enters the custom thresholds in a non-sorted order.

arose (approved):
thanks @luk27official this looks good to me. I wonder if this should be supported in other color themes as well but we can do that later.

PR #1462, opened 2025-03-08
Loadstyle

agdturner:
# Description
These changes allow users to load in colours. For example with 4ug0, you can load the attached file to colour in blue and orange the chains. To try this, download the structure 4ug0. Then in the Cartoon > Update 3D Representation > Chain Id ... > List 25 items > Colors 25 items > Scroll down the list to Browse... No file selected and action/click. Choose the file attached.

Currently the Json is written back out. This was because my colleagues use a different file format which can also optionally be loaded.

I suspect you might want a few changes in review, but thought this might be a useful contribution? I think I saw some discussion about others wanting functionality along these lines...   
[color_list.json](https://github.com/user-attachments/files/19136542/color_list.json)

## Actions

- [x ] Added description of changes to the `[Unreleased]` section of `CHANGELOG.md`
- [ x] Updated headers of modified files

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file, following t
 Note that since we don't clearly distinguish between a public and private interfaces there will be changes in non-major versions that are potentially breaking. If we make breaking changes to less used interfaces we will highlight it in here.
 
 ## [Unreleased]
-
+- Created a way to load a file of colors mapped to asym_id chain id values.
 - Fix PDBj structure data URL
 - Improve logic when to cull in renderer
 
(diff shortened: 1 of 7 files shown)


arose:
I would say this is too specific to be included here. Implementation-vise you should do this as an extension so you can have an unmodified copy of molstar plus your extension. Extension can also contain UI elements, see src\extensions\geo-export\ui.tsx

dsehnal:
It's a nice idea to be able to do this, but I second making this as an extension. The current implementation isn't specific to the colors even as you modify the `Object List` parameter control: this would affect all lists in the app, not just the color lists.

The extension can be a separate control in the "structures" panel that will list all represenations and allow the file upload/download.

PR #1464, opened 2025-03-11
fix: bond iterator starts from the same interbonds for every element

papillot:
# Description
Using the ElementBondIterator in the case of the attachment point of a covalent ligand (FMN) in a dimer (8Q5F), we noticed a discrepancy between the chain A (no issue) and the chain B: undefined element index due to an out of bound `otherIndex` for the iterator.

This is due to the fact that, for inter unit bonds, the iteration always begins at index 0 in the array of edges. The mapping to the edge indices is missing in the iteration loop.

## Actions


dsehnal (approved):
Thanks for the fix @papillot

PR #1466, opened 2025-03-14
Colors from user and file

corredD:
# Description
add support for color if present in the generic definition. Also add the option to load a json file to set the color at the root node level.

## Actions


src/apps/mesoscale-explorer/behavior/select.ts
```diff
@@ -73,7 +73,8 @@ export const MesoSelectLoci = PluginBehavior.create<MesoSelectLociProps>({
                         this.ctx.managers.interactivity.lociSelects.deselectAll();
                         return;
                     }
-                    const loci = Loci.normalize(current.loci, modifiers.control ? 'entity' : 'chain');
+                    const loci = Loci.normalize(current.loci, modifiers.control ? 'entity' : this.ctx.managers.interactivity.props.granularity);
+                    console.log(current.loci, loci, this.ctx.managers.interactivity.props.granularity);
```
dsehnal, line 77:
```suggestion
```

src/apps/mesoscale-explorer/ui/entities.tsx
```diff
@@ -881,12 +883,13 @@ export class GroupNode extends Node<{ filter: string }, { isCollapsed: boolean,
         const root = (isRoot && this.allGroups.length > 1) && <IconButton svg={BrushSvg} toggleState={false} disabled={disabled} small onClick={this.toggleRoot} />;
         const clip = <IconButton svg={ContentCutSvg} toggleState={false} disabled={disabled} small onClick={this.toggleClip} />;
         const visibility = <IconButton svg={state.isHidden ? VisibilityOffOutlinedSvg : VisibilityOutlinedSvg} toggleState={false} disabled={disabled} small onClick={this.toggleVisible} />;
-
+        const loadColorButton = (depth === 0) && <ColorLoaderControl />; // <IconButton svg={FileOutlineSvg} toggleState={false} disabled={disabled} small onClick={this.loadColorsFromFile} />;
```
dsehnal, line 886:
```suggestion
        const loadColorButton = (depth === 0) && <ColorLoaderControl />;
```

src/apps/mesoscale-explorer/ui/states.tsx
```diff
@@ -299,7 +300,45 @@ export const LoadModel = StateAction.build({
...
+    display: { name: 'Load Colors', description: 'Load color data for tree entities' },
+    params: {
+        file: PD.File({ accept: '.json' }),
+    },
+})(async ({ params }, ctx: PluginContext) => {
+    if (!params.file) {
+        ctx.log.error('No color file selected');
+        return;
+    }
+    const { data } = await ctx.builders.data.readFile({ file: params.file, isBinary: false });
```
dsehnal, line 314: This doesn't seem to be used in the state tree after getting the data out -- use `readFromFile` function directly? (no need to wrap the file as asset either)

src/apps/mesoscale-explorer/ui/states.tsx
```diff
@@ -337,6 +376,36 @@ export class ExampleControls extends PluginUIComponent {
     }
 }
 
+export class ColorLoaderControl extends PluginUIComponent {
```
dsehnal, line 379: Can we use functional components instead now? There might be time when react deprecates class components in the future, so the fewer we have the better.

src/apps/mesoscale-explorer/ui/states.tsx
```diff
@@ -299,7 +300,45 @@ export const LoadModel = StateAction.build({
     }
 }));
 
-//
+
+export const LoadCellPackColorsAction = StateAction.build({
```
dsehnal, line 304: Doing this as a state action only makes sense when this would be added to the UI via the plugin spec, which isn't the case here. This could be a simple function that accepts the `File` and `PluginContext`.

src/apps/mesoscale-explorer/ui/states.tsx
corredD: you mean like this : 
`
const data = await ctx.runTask(readFromFile(params.file.file, 'string'));
const colorData = JSON.parse(data);
`

src/apps/mesoscale-explorer/ui/states.tsx
corredD: in ui/states.tsx most are class extending PluginUIComponent, should they all be replace ? Also do you have a simple example of  functional components that can be used instead ?

src/apps/mesoscale-explorer/ui/states.tsx
corredD: Is there examples I could find that show how to do it more simply ?

src/apps/mesoscale-explorer/ui/states.tsx
dsehnal: For example https://github.com/molstar/molstar/blob/master/src/mol-plugin-ui/viewport/screenshot.tsx#L164

src/apps/mesoscale-explorer/ui/states.tsx
dsehnal: just make it a normal `function` which you call.

dsehnal (approved):

PR #1479, opened 2025-04-04
Big strings

midlik:
# Description

This is to solve issues with super big CIF files which cannot be parsed into a string due to Javascript engine MAX_STRING_LENGTH limitation (~500M chars).

Proposed solution:
- Replace type `string` by `StringLike` wherever big strings are expected to occur (e.g. in `PluginStateObject.Data.String` and in format parsers), `type StringLike = string | String | CustomString`
- Implement `CustomString` interface in `ChunkedBigString` class, using an array of constant-length strings shorter than MAX_STRING_LENGTH (only implement the most basic string methods, that are actually used in our code).
- When parsing UTF8 data, return regular `string` if possible. Only return `CustomString` when length limit exceeded.

Benchmarking: Accessing `ChunkedBigString` methods is roughly 25% slower than with normal string. 
But we only use `ChunkedBigString` when the data are really bigger than the limit, so with normal-size file, performance should not be affected.

## Actions


src/mol-io/common/string-like.ts
```diff
@@ -0,0 +1,572 @@
...
+    //  * @param searchString The substring to search for.
+    //  * @param position The index at which to begin searching. If omitted, the search begins at the end of the string.
+    //  */
+    // lastIndexOf(searchString: string, position?: number): number;
+
+    // /**
+    //  * Determines whether two strings are equivalent in the current locale.
+    //  * @param that String to compare to target string
+    //  */
+    // localeCompare(that: string): number;
```
dsehnal, line 61: For the final version I don't think we need to keep all these commented out functions.

src/mol-io/common/string-like.ts
```diff
@@ -0,0 +1,572 @@
...
+
+    constructor() {
+        return new Proxy(this, {
+            // overrides getting this[index]
+            get(self, index) {
+                if (typeof index === 'symbol') return self[index as any];
+                const i = Number.parseInt(index);
+                if (isNaN(i)) return self[index as any];
+                else return self.at(i);
+                // TODO See how this proxy thing affects performance (CIF parsing itself uses custom parseInt because Number.parseInt is slow)
```
dsehnal, line 394: The CIF parser relies on `charCodeAt`. We could have an `isASCII` flag, which would return this from the underying `Uint8Array` directly too when available.

dsehnal (commented):
I think this could work well if we manage to optimize it.

src/mol-io/common/string-like.ts
midlik: Yep. Removed.

src/mol-io/common/string-like.ts
midlik: In the meantime i removed support for [i], in favor of `charAt` and `charCodeAt`.

src/mol-io/common/string-like.ts
```diff
@@ -0,0 +1,291 @@
...
+            throw new Error(`Failed to convert StringLike object into string. This might be because the length ${str.length} exceeds maximum allowed string length ${MAX_STRING_LENGTH}. (${err})`);
+        }
+    },
+};
+
+
+/** Essential subset of `string` functionality. Add more string methods if needed. */
+interface CustomString {
+    /** Flag for recognizing `StringLike` objects */
+    _StringLike_: true,
```
dsehnal, line 33: `CustomString` is a strict subset of `string` (if we omit the `_StringLike_` magic flag which I don't think we need. So how about just doing this?

```suggestion
export const StringLike = {
    /** Try to convert `StringLike` to a primitive `string`. Might fail if the content is longer that max allowed string length. */
    asString(str: any): string {
        try {
            if (typeof str === 'string') return str;
            return str.toString();
        } catch (err) {
            throw new Error(`Failed to convert StringLike object into string. This might be because the length ${str.length} exceeds maximum allowed string length ${MAX_STRING_LENGTH}. (${err})`);
        }
    },
};


/** Essential subset of `string` functionality. Add more string methods if needed. */
export interface StringLike {
```

dsehnal (commented):
Very nice, good job! 

Before merning, please consider the suggestion bellow and since this is a big feature let's have @arose have a look too.

I think we should be able to use this (or perhaps a separate implementation of the StringLike interface) in https://github.com/molstar/molstar/pull/1413 too!

midlik:
> I think we should be able to use this (or perhaps a separate implementation of the StringLike interface) in #1413 too!

If the only issue in #1413 is too long strings, then this could potentially fix it out of the box 🤞

src/mol-io/common/string-like.ts
midlik: In many places in the code we need to check whether something is `StringLike` (mainly in functions that take `StringLike | Uint8Array`).
Originally there was `if (typeof obj === 'string')...` , which i had to change to `if (StringLike.is(obj))...`.

How would you achieve this without the `_StringLike_` magic flag?

dsehnal: How about check if it has a `charCodeAt` function and implement the .is function based on that?

dsehnal: Alternatively, the actual implementations of the interface could have the magic flag, but not make it part of the interface to make it trivially compatible with string.

The .is check could then check if the value is a builtin string object or has the magic flag.

midlik: > Alternatively, the actual implementations of the interface could have the magic flag, but not make it part of the interface to make it trivially compatible with string.
> 
> The .is check could then check if the value is a builtin string object or has the magic flag.

That does not force implementations to have the flag.

I think i can do the ducktyping approach (`typeof (obj as any).charCodeAt === 'function'`)

midlik: Changed

arose:
will review later this week

src/mol-io/common/string-like.ts
```diff
@@ -0,0 +1,287 @@
...
+
+    toString(): string {
+        try {
+            return this._chunks.join('');
+        } catch (err) {
+            throw new Error(`Failed to convert StringLike object into string. This might be because the length ${this.length} exceeds maximum allowed string length ${MAX_STRING_LENGTH}. (${err})`);
+        }
+    }
+
+    _repr(): string {
```
arose, line 284: what is this for?

src/mol-io/common/string-like.ts
```diff
@@ -0,0 +1,287 @@
...
+    /**
+     * Returns the substring at the specified location within a String object.
+     * @param start The zero-based index number indicating the beginning of the substring.
+     * @param end Zero-based index number indicating the end of the substring. The substring includes the characters up to, but not including, the character indicated by end.
+     * If end is omitted, the characters from start through the end of the original string are returned.
+     */
+    substring(start: number, end?: number): string;
+
+    /**
+     * Returns the position of the first occurrence of a substring.
```
arose, line 48:
```suggestion
     * Returns the position of the first occurrence of a substring or -1 if not found.
```

src/mol-io/common/string-like.ts
```diff
@@ -0,0 +1,287 @@
...
+
+    /**
+     * Returns the character at the specified index.
+     * @param pos The zero-based index of the desired character.
+     */
+    charAt(pos: number): string;
+
+    /**
+     * Returns the Unicode value of the character at the specified location.
+     * @param index The zero-based index of the desired character. If there is no character at the specified index, NaN is returned.
```
arose, line 35:
```suggestion
     * @param index The zero-based index of the desired character. If the specified index is out of range, NaN is returned.
```

src/mol-io/common/string-like.ts
```diff
@@ -0,0 +1,287 @@
...
+ * Add more string methods if needed.
+ */
+export interface StringLike {
+    /** Returns the length of a String object. */
+    readonly length: number;
+
+    // readonly [index: number]: string; // Avoided implementing this as it would require using Proxy, which drastically affects performance.
+
+    /**
+     * Returns the character at the specified index.
```
arose, line 22:
```suggestion
     * Returns the character at the specified index or `undefined` if not found. Supports relative indexing from the end of the string when passed a negative index.
```

src/mol-io/common/string-like.ts
```diff
@@ -0,0 +1,287 @@
...
+     * @param pos The zero-based index of the desired character.
+     */
+    at(index: number): string | undefined;
+
+    /**
+     * Returns the character at the specified index.
```
arose, line 28:
```suggestion
     * Returns the character at the specified index or an empty string if the index is out of range.
```

src/mol-io/common/string-like.ts
```diff
@@ -0,0 +1,287 @@
...
+        try {
+            return str.toString();
+        } catch (err) {
+            throw new Error(`Failed to convert StringLike object into string. This might be because the length ${str.length} exceeds maximum allowed string length ${MAX_STRING_LENGTH}. (${err})`);
+        }
+    },
+};
+
+
+/** Maximum allowed string length (might be bigger for some engines, but in Chrome and Node it is this). */
```
arose, line 91:
```suggestion
/** Maximum allowed string length (might be bigger for some engines, but in Chrome 136 and Node 22 it is this). */
```

src/mol-io/common/utf8.ts
```diff
@@ -106,6 +110,15 @@ export function utf8Read(data: Uint8Array, offset: number, length: number) {
     }
 }
 
+/** Decode UTF8 data, potentially exceeding MAX_STRING_LENGTH. Return as primitive `string` if possible; or as `ChunkedBigString` if the result is longer than MAX_STRING_LENGTH. */
+export function utf8ReadLong(data: Uint8Array, offset: number = 0, length: number = data.length): StringLike {
```
arose, line 114: ha, TIL you can reference args in defaults

arose (approved):
This is great, thank you @midlik! Just a few small suggestions and questions.

arose:
Also, tried to load the big petworld model (601 MB) without struct_conn records removed and it works nicely.

.gitignore
```diff
@@ -11,4 +11,5 @@ tsconfig.commonjs.tsbuildinfo
 *.sublime-workspace
 .idea
 
-.DS_Store
\ No newline at end of file
+.DS_Store
+tmp/
```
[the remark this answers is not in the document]
midlik: Oh, sometimes I put some data for manual testing in tmp/ .
Do you want me to remove this?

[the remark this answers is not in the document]
dsehnal: I think this is fine.

src/mol-io/common/string-like.ts
midlik: This was used in debugging. Yes, i can remove it.

src/mol-io/common/utf8.ts
midlik: I was also surprised.
Nice feature.

src/mol-io/common/string-like.ts
midlik: Removed.

midlik:
Thank you for the review. I think it's ready for merging now.

@arose i appreciate you adding details to the docstrings. I copied them from `string` documentation, it's sad that they don't include these details there.

PR #1483, opened 2025-04-13
added toggle for hiding screenshot controls

cmishra:
# Description

Added an extra config to control whether the screenshot tool pops up in the viewer or not. Seems reasonable to have given most of the other buttons have a toggle. 

Maybe it makes sense to also add one to Refresh / reset button? I don't need to hide it, but happy to add one for others if you all think it does. 

## Actions

### DIFF
--- a/src/mol-plugin-ui/viewport.tsx
+++ b/src/mol-plugin-ui/viewport.tsx
@@ -4,6 +4,7 @@
  * @author Alexander Rose <alexander.rose@weirdbyte.de>
  * @author David Sehnal <david.sehnal@gmail.com>
  * @author Adam Midlik <midlik@gmail.com>
+ * @author Chetan Mishra <chetan.s115@gmail.com>
  */
 
 import * as React from 'react';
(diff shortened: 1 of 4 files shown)


src/mol-plugin-ui/viewport.tsx
```diff
@@ -121,7 +122,7 @@ export class ViewportControls extends PluginUIComponent<ViewportControlsProps, V
                 </div>
                 <div>
                     <div className='msp-semi-transparent-background' />
-                    {this.icon(CameraOutlinedSvg, this.toggleScreenshotExpanded, 'Screenshot / State Snapshot', this.state.isScreenshotExpanded)}
+                    {this.plugin.config.get(PluginConfig.Viewport.ShowScreenshotControls) && this.icon(CameraOutlinedSvg, this.toggleScreenshotExpanded, 'Screenshot / State Snapshot', this.state.isScreenshotExpanded)}
```
dsehnal, line 125: This should be around the outside div since it's just a single control -- but I will fix that in a separate PR I am working on.

dsehnal (approved):
Thanks.

PR #1502, opened 2025-04-30
Avoid grid expansion when requiring unit cell on volume server

giagitom:
# Description

Avoids expanding the box when the whole unit cell is required. 

## Actions

### DIFF
--- a/src/servers/volume/server/version.ts
+++ b/src/servers/volume/server/version.ts
@@ -1,2 +1,2 @@
-export const VOLUME_SERVER_VERSION = '0.9.6';
+export const VOLUME_SERVER_VERSION = '0.9.7';
 export const VOLUME_SERVER_HEADER = `VolumeServer ${VOLUME_SERVER_VERSION}, (c) 2018-2024, Mol* contributors`;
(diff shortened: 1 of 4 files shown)


dsehnal:
The code LGTM. Were you able to test this locally too?

giagitom:
> The code LGTM. Were you able to test this locally too?

Not yet, which is the best way of doing this? Is there some documentation?

dsehnal:
https://molstar.org/docs/data-access-tools/volume-server/ should have the important info.

giagitom:
Ok thanks, so I tried to test it locally and it seems to work as expected. Now when using cell mode there isn't a grid expansion.

Also I have changed the version, and used the volume server specific changelog instead of the main one.

src/servers/volume/server/version.ts
```diff
@@ -1,2 +1,2 @@
-export const VOLUME_SERVER_VERSION = '0.9.6';
+export const VOLUME_SERVER_VERSION = '0.9.7';
 export const VOLUME_SERVER_HEADER = `VolumeServer ${VOLUME_SERVER_VERSION}, (c) 2018-2024, Mol* contributors`;
```
dsehnal, line 2:
```suggestion
export const VOLUME_SERVER_HEADER = `VolumeServer ${VOLUME_SERVER_VERSION}, (c) 2018-2025, Mol* contributors`;
```

dsehnal (approved):

PR #1504, opened 2025-05-01
Add format selection option to image export UI

dsehnal:
# Description

- [x] Add format selection option to image export UI (PNG, WebP, JPEG)

image

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,11 +5,12 @@ Note that since we don't clearly distinguish between a public and private interf
 
 
 ## [Unreleased]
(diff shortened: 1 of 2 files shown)


arose (changes requested):
webp is not supported in Safari (see https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toBlob#browser_compatibility) so should disable it in those browsers since Browsers will revert to png if a non-supported format is requested which would mean we return a webp file that contains a png.

arose (approved):

PR #1505, opened 2025-05-01
MVS: support updating transform states

dsehnal:
# Description

In response to https://github.com/molstar/mol-view-spec/issues/63

- [x] Add `StateBuilder.To.updateState`
- [x] MVS: Support updating transform states
- [x] MVS: Add support for `molstar_is_hidden` custom state as an extension 
- [x] MVS: Add `queryMVSRef` and `createMVSRefMap` utility functions

Extension creation docs https://github.com/molstar/mol-view-spec-docs/pull/3

## Actions


src/extensions/mvs/load-extensions/is-hidden-custom-state.ts
```diff
@@ -0,0 +1,18 @@
...
+ * Copyright (c) 2025 mol* contributors, licensed under MIT, See LICENSE file for more info.
+ *
+ * @author David Sehnal <david.sehnal@gmail.com>
+ */
+
+import { MolstarLoadingExtension } from '../load';
+
+
+export const IsHiddenCustomStateExtension: MolstarLoadingExtension<{}> = {
+    id: 'mvs/is-hidden-custom-state',
```
midlik, line 11: I conceived the prefix of the id as a "scope", something like name of the organization that maintains the extension or name of the project it is used for. `mvs/` makes little sense here, as all extensions will be for MVS. 

Also, perhaps `-custom-state` is superfluous as all these extensions are kinda expected to work with custom state.

So maybe `wwpdb/is-hidden` (or `wwpdb/hide-nodes' or similar)?

midlik (commented):
Thank you @dsehnal , looks nice. There is just one thing i would change (plus naming doubts :D), see comments.

PR #1507, opened 2025-05-02
Update dependencies in package.json and package-lock.json

paullewallencom:
# 🧪 Add Vitest Integration, Separate Jest Config, and Restructure Test Directory

## Summary

This PR introduces a series of improvements to the testing infrastructure:

- ✅ **Moved Jest config** from `package.json` into a standalone `jest.config.js` file using `ts-jest`.
- ✅ **Added `vitest.config.cjs`** to begin integrating Vitest as a faster alternative test runner.
- ✅ **Created recursive folder-generation scripts** in PowerShell and Bash to automate test directory setup.
- ✅ **Established new convention** for test folder placement using `__tests__` subdirectories across the project.
- ✅ **Created minimal test files** for both Jest and Vitest to benchmark performance differences.
- ✅ **Updated `package.json`** to include separate scripts for running tests via Jest or Vitest:
  - `npm run test:jest`
  - `npm run test:vitest`
  - `npm run test:vitest:watch`
  - `npm run test:vitest:coverage`

## Rationale

- 🧭 Separating test configuration files improves modularity and clarity.
- 🚀 **Vitest** offers dramatically faster test execution, which encourages frequent test runs and boosts test coverage.
- 🧰 Shell and PowerShell scripts allow quick setup of a consistent test folder structure.
- 🧪 Some test cases may still require Jest's feature set, so both frameworks are supported side by side for flexibility.

## Installation Notes

I ran the following to update/install dependencies:

### Jest Setup
```bash
npm install --save-dev jest ts-jest @types/jest
npx ts-jest config:init
```

### Vitest Setup
```bash
npm install --save-dev vitest
npm install --save-dev @types/node
```

## Next Steps

This PR does not migrate existing tests to the new structure but lays the groundwork. I recommend migrating tests incrementally in future PRs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file, following t
 Note that since we don't clearly distinguish between a public and private interfaces there will be changes in non-major versions that are potentially breaking. If we make breaking changes to less used interfaces we will highlight it in here.
 
 ## [Unreleased]
+- Fix transparency rendering with occlusion in NodeJS
 
 ## [v4.14.1] - 2025-05-09
 - Do not raise error when creating duplicate state transformers and print console warning instead
--- a/src/mol-canvas3d/passes/draw.ts
+++ b/src/mol-canvas3d/passes/draw.ts
@@ -314,7 +314,7 @@ export class DrawPass {
                     if (!this.packedDepth) {
                         this.depthTextureOpaque.attachFramebuffer(this.transparentColorTarget.framebuffer, 'depth');
                     } else {
-                        this.colorTarget.depthRenderbuffer?.detachFramebuffer(this.transparentColorTarget.framebuffer);
+                        this.colorTarget.depthRenderbuffer?.attachFramebuffer(this.transparentColorTarget.framebuffer);
                     }
                 }
 


tsconfig.json
```diff
@@ -1,24 +1,25 @@
...
+        "target": "ES2020",
         "alwaysStrict": true,
         "noImplicitAny": true,
         "noImplicitThis": true,
         "sourceMap": false,
         "noUnusedLocals": true,
         "strictNullChecks": true,
         "strictFunctionTypes": true,
-        "module": "esnext",
+        "module": "CommonJS",
```
dsehnal, line 12: Why this change? We won't be able to make this change as the `lib` is expected to be esnext modules. There is a separate `tsconfig.commonjs.json` for commonjs build of the library.

src/mol-util/__tests__/jest/array.test.ts
```diff
@@ -0,0 +1,12 @@
+import { arrayMax } from '../../array';
```
dsehnal, line 1: Could we just include the `import { describe, it, expect } from 'vitest';` in all existing `_spec` files instead and move to vitest?

01-create-__tests__-folders.ps1
```diff
@@ -0,0 +1,9 @@
+# 01-create-__tests__-folders.ps1
```
dsehnal, line 1: Why do we need these scripts?

tsconfig.json
paullewallencom: After I initiated the pull request the build kept failing and so additioinal commits were needed to to pass the build.

src/mol-util/__tests__/jest/array.test.ts
paullewallencom: In some cases jest is superior to vitest in mocking, so that's why it seemed that in tandem it would be more flexible.

01-create-__tests__-folders.ps1
paullewallencom: As the code coverage increases, I'm concerned that test files will be disorganized. I included the scripts because they create __tests__ directories recursively, so it will keep the test files organized.

tsconfig.json
dsehnal: I am afraid we won't be able to make this change currently though because this builds the esnext modules in `lib` folder which are distributed.

01-create-__tests__-folders.ps1
dsehnal: But you create the test file manually, no? What is the value of just creating the `__test__` folder everywhere?

src/mol-util/__tests__/jest/array.test.ts
dsehnal: I am not sure that having two test frameworks is the best way to go. I would just pick one and stick with it.

01-create-__tests__-folders.ps1
dsehnal: I would also prefer we kept the `_spec` naming convention for folders with tests.

PR #1513, opened 2025-05-09
Fix transparency rendering with occlusion in NodeJS

midlik:
# Description

This attempts to fix transparency rendering bug which occurs when occlusion is on, transparencyMode is blended, and depthTexture extension is not available (this is the case when running in NodeJS).

Before:
1hda_2_B_Pfam_PF00042_image-800x800

After:
1hda_2_B_Pfam_PF00042_image-800x800

Please, @arose and/or @giagitom  check if the fix makes sense. I have no idea what the code does but it looked suspicious to have `attach` in `if` and `detach` in `else`, and changing it seems to solve my issue 🤣 

## Actions


giagitom (approved):
Yeah, this seems definitely a bug. Thanks for the fix!

arose (approved):
thanks

PR #1514, opened 2025-05-11
Remove unnecessary files as per review feedback

paullewallencom:
# 🧪 Replaced Jest w/ Vitest Integration

## Summary

This PR introduces a series of improvements to the testing infrastructure:

- ✅ **Removed Jest config** from `package.json` .
- ✅ **Added `vitest`** to begin integrating Vitest as a faster alternative test runner.
- ✅ **Updated `package.json`** to include separate scripts for running tests via Vitest:
  - `npm run test:vitest`
  - `npm run test:vitest:watch`
  - `npm run test:vitest:coverage`

## Rationale

- 🚀 **Vitest** offers dramatically faster test execution, which encourages frequent test runs and boosts test coverage.

## Installation Notes

I ran the following to update/install dependencies:

### Vitest Setup
```bash
npm install --save-dev vitest
npm install --save-dev @types/node
```

## Next Steps

This PR does not migrate existing tests to the new structure but lays the groundwork. I recommend migrating tests incrementally in future PRs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,8 +16,10 @@
   "scripts": {
     "lint": "eslint .",
     "lint-fix": "eslint . --fix",
-    "test": "npm install --no-save \"gl@^6.0.2\" && npm run lint && jest",
-    "jest": "jest",
+    "test": "npm install --no-save \"gl@^6.0.2\" && npm run lint && nps vitest",
+    "test:vitest": "vitest",
+    "test:vitest:watch": "vitest --watch",
+    "test:vitest:coverage": "vitest run --coverage --coverage.reporter=text-summary --coverage.dir=./coverage",
     "build": "npm run build-tsc && npm run build-extra && npm run build-webpack",
     "clean": "node ./scripts/clean.js",
     "rebuild": "npm run clean && npm run build",
@@ -74,22 +76,30 @@
     "ignore": [],
     "delay": "2500"
   },
-  "jest": {
-    "moduleFileExtensions": [
-      "ts",
-      "js"
-    ],
-    "transform": {
-      "\\.ts$": "ts-jest"
-    },
-    "moduleDirectories": [
-      "node_modules",
-      "lib"
-    ],
-    "testEnvironmentOptions": {
-      "url": "http://localhost/"
-    },
-    "testRegex": "\\.spec\\.ts$"
+  "vitest": {
+    "test": {
+      "globals": true,
+      "include": ["**/*.spec.ts"],
+      "environment": "jsdom",
+      "coverage": {
+        "enabled": true,
+        "reportsDirectory": "./coverage",
+        "reporter": ["text-summary"]
+      },
+      "setupFiles": [],
+      "resolveSnapshotPath": {
+        "lib": "./lib"
+      },
+      "alias": {
+        "lib": "./lib",
+        "@": "./src"
+      },
+      "environmentOptions": {
+        "jsdom": {
+          "url": "http://localhost/"
+        }
+      }
+    }
   },
   "author": "Mol* Contributors",
   "contributors": [
@@ -132,7 +142,6 @@
   "devDependencies": {
     "@types/cors": "^2.8.17",
     "@types/gl": "^6.0.5",
-    "@types/jest": "^29.5.14",
     "@types/pngjs": "^6.0.5",
     "@types/react": "^18.3.18",
     "@types/react-dom": "^18.3.5",
@@ -150,7 +159,6 @@
     "file-loader": "^6.2.0",
     "fs-extra": "^11.3.0",
     "http-server": "^14.1.1",
-    "jest": "^29.7.0",
     "jpeg-js": "^0.4.4",
     "mini-css-extract-plugin": "^2.9.2",
     "path-browserify": "^1.0.1",
@@ -162,8 +170,9 @@
     "simple-git": "^3.27.0",
     "stream-browserify": "^3.0.0",
     "style-loader": "^4.0.0",
-    "ts-jest": "^29.2.5",
     "typescript": "^5.7.3",
+    "vite": "^4.4.1",
+    "vitest": "^0.34.3",
     "webpack": "^5.97.1",
     "webpack-cli": "^6.0.1"
   },
--- a/package-lock.json
+++ b/package-lock.json
@@ -48,7 +48,6 @@
       "devDependencies": {
         "@types/cors": "^2.8.17",
         "@types/gl": "^6.0.5",
-        "@types/jest": "^29.5.14",
         "@types/pngjs": "^6.0.5",
         "@types/react": "^18.3.18",
         "@types/react-dom": "^18.3.5",
@@ -66,7 +65,6 @@
         "file-loader": "^6.2.0",
         "fs-extra": "^11.3.0",
         "http-server": "^14.1.1",
-        "jest": "^29.7.0",
         "jpeg-js": "^0.4.4",
         "mini-css-extract-plugin": "^2.9.2",
         "path-browserify": "^1.0.1",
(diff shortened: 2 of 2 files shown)


paullewallencom:
I having some trouble with examples.spec.ts. Are there any suggestions?

dsehnal (commented):
This looks good, though still need to fix the transpilers test:

image


@arose before we merge this any objections?

vitest.config.ts
arose, line 1: could this just be a js file? why is it included in tsconfig? is it actually needed with the other config inside of package.json?

package.json
```diff
@@ -74,22 +76,34 @@
...
+      "coverage": {
+        "enabled": true,
+        "reportsDirectory": "./coverage",
+        "reporter": [
+          "text-summary"
+        ]
+      },
+      "setupFiles": [],
+      "resolveSnapshotPath": {
+        "lib": "./lib"
```
arose, line 95: We only want to test things in `src/`, not in `lib/`, which only exists after a build and does not contain different test than what is in `src/`.

arose (changes requested):
Thanks for this effort @paullewallencom!

Currently I use the "firsttris.vscode-jest-runner" (which is in the recomended extensions in `.vscode\extensions.json`) in VSCode for test development. Can you place replace that with the official vitest extension and make sure it runs, thanks.

package.json
paullewallencom: Task complete

vitest.config.ts
paullewallencom: Task complete

package.json
```diff
@@ -16,8 +16,10 @@
   "scripts": {
     "lint": "eslint .",
     "lint-fix": "eslint . --fix",
-    "test": "npm install --no-save \"gl@^6.0.2\" && npm run lint && jest",
-    "jest": "jest",
+    "test": "npm install --no-save \"gl@^6.0.2\" && npm run lint && npm run vitest",
+    "test:vitest": "vitest",
+    "test:vitest:watch": "vitest --watch",
+    "test:vitest:coverage": "vitest run --coverage --coverage.reporter=text-summary --coverage.dir=./coverage",
```
dsehnal, line 22: Remove the `test:` prefix or fix the `test` command (it only says `vitest`, not `test:vitest`)

package.json
```diff
@@ -74,22 +76,34 @@
...
-    "testEnvironmentOptions": {
-      "url": "http://localhost/"
-    },
-    "testRegex": "\\.spec\\.ts$"
+  "vitest": {
+    "test": {
+      "globals": true,
+      "include": [
+        "**/*.spec.ts"
+      ],
```
dsehnal, line 84: Thinking about this, we should actually exclude the `spec` files from being compiled at all in `tsconfig.json` so they don't get included in the NPM package. @paullewallencom Can you have a look into this please or want us to do this?

package.json
```diff
@@ -74,22 +76,34 @@
...
+        ]
+      },
+      "setupFiles": [],
+      "resolveSnapshotPath": {
+        "lib": "./src"
+      },
+      "alias": {
+        "lib": "./lib",
+        "@": "./src"
+      },
```
dsehnal, line 100: We don't use these aliases in the source code. I did experiment with the `@` alias, but there is an issue with this as the tsc compiler doesn't rewrite the paths and it would involve more work.

package.json
```diff
@@ -162,9 +174,10 @@
     "simple-git": "^3.27.0",
     "stream-browserify": "^3.0.0",
     "style-loader": "^4.0.0",
-    "ts-jest": "^29.3.4",
-    "typescript": "^5.8.3",
-    "webpack": "^5.99.8",
+    "typescript": "^5.7.3",
+    "vite": "^4.4.1",
```
dsehnal, line 178: `vitest` requires importing `vite` too? Because we already use `esbuild` (and I have some plans to get rid of webpack for the production builds too).

package.json
```diff
@@ -183,6 +196,7 @@
     "immer": "^10.1.1",
     "immutable": "^5.1.2",
     "io-ts": "^2.2.22",
+    "jsdom": "^26.1.0",
```
dsehnal, line 199: Why is `jsdom` not a dev dependency?

vitest.config.js
```diff
@@ -0,0 +1,8 @@
+const { defineConfig } = require('vitest/config');
```
dsehnal, line 1: Do we need this file if you already specified a config in `package.json`? Perhaps choose just 1 location?

tsconfig.json
```diff
@@ -20,6 +20,6 @@
         "rootDir": "src",
         "outDir": "lib"
     },
-    "include": [ "src/**/*" ],
+    "include": [ "src/**/*", "vitest.config.ts" ],
```
dsehnal, line 23: There is no `vitest.config.ts`

.eslintrc.json
```diff
@@ -117,6 +117,12 @@
                 "@typescript-eslint/func-call-spacing": ["error"],
                 "@typescript-eslint/keyword-spacing": ["error"]
             }
+        },
+        {
+            "files": ["vitest.config.ts"],
```
dsehnal, line 122: This file doesn't exist (and perhaps should remain the .js file you currently include)

package.json
arose: they are already excluded because we have a `_spec` pattern in .npmignore

package.json
arose: why do we need vite here?

package.json
arose: Same question, why not a dev dep? also, we probably don't really need as we can use the defaultvitest environment "node". Mol* runs in both, the browser and node. Any particular reason to use jsdom here?

package.json
```diff
@@ -193,7 +202,7 @@
   },
   "peerDependencies": {
     "@google-cloud/storage": "^7.14.0",
-    "canvas": "^2.11.2",
+    "canvas": "^3.1.0",
```
arose, line 205: why did you update it? should probably be done separately

tsconfig.json
```diff
@@ -21,6 +21,6 @@
         "rootDir": "src",
         "outDir": "lib"
     },
-    "include": [ "src/**/*" ],
-    // "exclude": [ "src/servers/**/*", "src/perf-tests/*", "src/cli/**/*" ]
+    "include": ["src/**/*"],
+    "exclude": [ "**/*.spec.ts", "src/servers/**/*", "src/perf-tests/*", "src/cli/**/*"]
```
arose, line 25: please revert that or do you have a specific reason?

tsconfig.commonjs.json
```diff
@@ -21,5 +21,5 @@
         "rootDir": "src",
         "outDir": "lib/commonjs"
     },
-    "include": [ "src/**/*" ]
+    "include": [ "src/**/*" , "vitest.config.ts" ]
```
arose, line 24: there is no "vitest.config.ts" anymore

.vscode/launch.json
```diff
@@ -11,6 +11,19 @@
...
             "sourceMaps": true,
             "webRoot": "${workspaceFolder}"
-        }
+        },
+        {
+            "type": "node",
+            "request": "launch",
+            "name": "Debug Vitest",
+            "program": "${workspaceFolder}/node_modules/vitest/vitest.mjs",
+            "args": ["run", "--inspect", "--singleThread", "src/mol-script/transpilers/_spec/examples.spec.ts"],
```
arose, line 20: that seems to be arbitrarily specific, do we need that config here?

package.json
```diff
@@ -163,10 +172,9 @@
     "simple-git": "^3.27.0",
     "stream-browserify": "^3.0.0",
     "style-loader": "^4.0.0",
-    "ts-jest": "^29.3.4",
-    "typescript": "^5.8.3",
-    "webpack": "^5.99.8",
-    "webpack-cli": "^6.0.1"
+    "typescript": "^5.7.3",
+    "vite": "^4.4.1",
+    "vitest": "^0.34.3"
```
arose, line 177: that is an ancient vitest version and it does not work with the vscode testing integration - updating it makes it work which is very nice

PR #1520, opened 2025-05-17
WebGL ContextLost handling improvements

arose:
- Fix missing framebuffer & drawbuffer re-attachments
- Fix missing cube texture re-initialization
- Fix missing extensions reset
- Fix timer clearing edge case
- Add reset support for geometry generated on the GPU

Test by running in debug mode, then hold shift+ctrl+alt and mouse left click. This will trigger a context loss.


dsehnal (approved):
lgtm

PR #1524, opened 2025-05-20
MVS single state loading

midlik:
# Description

This changes the way single-state MVS files are loaded. A single-state MVS is now treated the same as a multi-state MVS with one snapshot.

Related changes:
- Merged `loadMVS` options `keepCamera` and `keepSnapshotCamera`, as `keepCamera` lost meaning 
- Added `loadMVS` option `appendSnapshots` ("do not remove current list of snapshots")
- Removed `loadMVS` option `replaceExisting` (is now default, state trees are never combined)
- Drag-and-dropping multiple MVSs loads concatenation of their snapshots (instead of combining state trees into one)

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ Note that since we don't clearly distinguish between a public and private interf
 
 ## [Unreleased]
 - Load potentially big text files as `StringLike` to bypass string size limit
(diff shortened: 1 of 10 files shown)


dsehnal (approved):
LGTM

PR #1525, opened 2025-05-20
Fix camera interpolation during animation export

dsehnal:
# Description
https://github.com/molstar/molstar/pull/1520 introduced a bug in MP4 export. This should fix it.

## Actions


src/mol-canvas3d/canvas3d.ts
```diff
@@ -865,7 +868,7 @@ namespace Canvas3D {
         }
 
         const contextLostSub = contextLost?.subscribe(() => {
-            pause(true);
+            isContextLost = true;
```
dsehnal, line 871: @arose  we need to use this (or a similar approach that doesn't reset the animation loop), because the plugin animation is managed by the `PluginAnimationLoop`

arose (approved):

Issue #1488, opened 2025-04-16
Runtime Error with Next.js turbopack


MattBurn:
When using Next.js (15.3.0, although I have experienced this through several versions of next) turbopack, any page that imports from molstar (4.13.0 although it has been an issue with several versions of molstar) I get a runtime error with the following stacktrace:
```
Runtime Error

Error: Cannot read properties of undefined (reading 'Provider')

Call Stack 52

[project]/node_modules/molstar/lib/mol-model-formats/structure/mmcif.js [app-client] (ecmascript)
node_modules/molstar/lib/mol-model-formats/structure/mmcif.js (29:15)
...
[project]/node_modules/molstar/lib/mol-model/structure/query/queries/modifiers.js [app-client] (ecmascript)
node_modules/molstar/lib/mol-model/structure/query/queries/modifiers.js (14:1)
...
[project]/node_modules/molstar/lib/mol-model/structure/query.js [app-client] (ecmascript) <locals>
node_modules/molstar/lib/mol-model/structure/query.js (10:1)
...
[project]/node_modules/molstar/lib/mol-model-formats/structure/property/assembly.js [app-client] (ecmascript)
node_modules/molstar/lib/mol-model-formats/structure/property/assembly.js (9:1)
...
[project]/node_modules/molstar/lib/mol-model-formats/structure/property/symmetry.js [app-client] (ecmascript)
node_modules/molstar/lib/mol-model-formats/structure/property/symmetry.js (10:1)
...
[project]/node_modules/molstar/lib/mol-model/structure/structure/unit.js [app-client] (ecmascript)
node_modules/molstar/lib/mol-model/structure/structure/unit.js (21:1)
...
[project]/node_modules/molstar/lib/mol-model/structure/structure/properties.js [app-client] (ecmascript)
node_modules/molstar/lib/mol-model/structure/structure/properties.js (7:1)
...
[project]/node_modules/molstar/lib/mol-model/structure/query/queries/internal.js [app-client] (ecmascript)
node_modules/molstar/lib/mol-model/structure/query/queries/internal.js (8:1)
...
[project]/node_modules/molstar/lib/mol-model/structure/query/context.js [app-client] (ecmascript)
node_modules/molstar/lib/mol-model/structure/query/context.js (9:1)
...
InnerLayoutRouter
node_modules/next/src/client/components/layout-router.tsx (415:5)
OuterLayoutRouter
node_modules/next/src/client/components/layout-router.tsx (615:19)
```

[...]

PR #1533, opened 2025-06-01
isolatedModules and fix turbopack build

dsehnal:
# Description

- [x] `isolatedModules` tsconfig flag (should streamline compilation/packaging)
- [x] Fix turbopack build with ES6 modules (fixes https://github.com/molstar/molstar/issues/1488)
  - [x] There was some out-of-order code execution. Interestingly, this only happens together with turbopack and ES6 modules were using (CommonJS was fine), and wasn't happening with other builds such as webpack, rollup, or esbuild 

## Actions

### DIFF
--- a/src/mol-gl/_spec/gl.shim.ts
+++ b/src/mol-gl/_spec/gl.shim.ts
@@ -786,5 +786,5 @@ export function createGl(width: number, height: number, contextAttributes: WebGL
         stencilOp: function () { },
         stencilOpSeparate: function () { },
         unpackColorSpace: 'srgb',
-    } as unknown as WebGLRenderingContext;
+    };
 }
--- a/src/mol-gl/_spec/gl.shim.ts
+++ b/src/mol-gl/_spec/gl.shim.ts
@@ -787,5 +787,5 @@ export function createGl(width: number, height: number, contextAttributes: WebGL
         stencilOpSeparate: function () { },
         // Caused TSC build error with isolated modules
         // unpackColorSpace: 'srgb',
-    };
+    } as unknown as WebGLRenderingContext;
 }
--- a/src/mol-gl/_spec/gl.shim.ts
+++ b/src/mol-gl/_spec/gl.shim.ts
@@ -785,7 +785,6 @@ export function createGl(width: number, height: number, contextAttributes: WebGL
         stencilMaskSeparate: function () { },
         stencilOp: function () { },
         stencilOpSeparate: function () { },
-        // Caused TSC build error with isolated modules
-        // unpackColorSpace: 'srgb',
+        unpackColorSpace: 'srgb',
     } as unknown as WebGLRenderingContext;
 }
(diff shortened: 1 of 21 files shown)


zachcp:
+100 for improved ES6 module compatability.

src/mol-model/structure/query/queries/modifiers.ts
```diff
@@ -576,16 +584,14 @@ export function surroundingLigands({ query, radius, includeWater }: SurroundingL
...
-    return _entity_type(l) === 'water';
-}
-
 function getPrdAsymIdx(structure: Structure) {
     const model = structure.models[0];
     const ids = new Set<string>();
-    if (!MmcifFormat.is(model.sourceData)) return ids;
-    const { _rowCount, asym_id } = model.sourceData.data.db.pdbx_molecule;
+
+    // Need to do this menually to prevent a cyclical import causing
```
arose: How about splitting src\mol-model-formats\structure\mmcif.ts into mmcif/format.ts and mmcif/data.ts or so? The cyclic dependency issue can creep up over and over again and thios would fix it.

src/mol-gl/_spec/gl.shim.ts
```diff
@@ -785,6 +785,7 @@ export function createGl(width: number, height: number, contextAttributes: WebGL
         stencilMaskSeparate: function () { },
         stencilOp: function () { },
         stencilOpSeparate: function () { },
-        unpackColorSpace: 'srgb',
-    };
+        // Caused TSC build error with isolated modules
```
arose: why would that be the case? I don't see an error on my machine

arose (changes requested):
this is great, thanks

src/mol-model/structure/query/queries/modifiers.ts
dsehnal: Where would we import the data (the part with the inline code execution that registers the custom property providers)? ... because that's the part that causes the issues.

dsehnal: In general, these are like 200 places where we do cyclic imports. It only causes an issue where there is the inline code execution though.

dsehnal: Another solution would be to move where we do the format provider registration, e.g., to the files with the providers.

src/mol-gl/_spec/gl.shim.ts
dsehnal: image

But it's enough to do `as unknown as WebGLRenderingContext` to fix this (before I had just the comment out part)

src/mol-model/structure/query/queries/modifiers.ts
dsehnal: Don't know how to change this. I tried splitting the inline code execution into a separate file and importing it from the `mmcif.ts`, but that didn't work.

I think overall we will be fine and if this creeps up again, we do a similar fix. But perhaps I am missing something obvious...

PR #1535, opened 2025-06-07
Support sampling from arbitrary planes

arose:
Support sampling from arbitrary planes for structure plane and volume slice representations

# Description

## Actions


src/mol-geo/geometry/image/image.ts
```diff
@@ -71,6 +71,8 @@ interface Image {
 
     /** Bounding sphere of the image */
     boundingSphere: Sphere3D
```
dsehnal:
```suggestion
    readonly boundingSphere: Sphere3D
```

src/mol-geo/geometry/image/image.ts
```diff
@@ -127,6 +129,10 @@ namespace Image {
                 }
                 return boundingSphere;
             },
+            setBoundingSphere(sphere: Sphere3D) {
+                Sphere3D.copy(boundingSphere, sphere);
+                currentHash = hashCode(image);
```
dsehnal, line 134: I don't understand this. The hash only depends on the `image.cornerBuffer.ref.version`, so why re-assign it here?

src/mol-repr/structure/visual/plane-image.ts
```diff
@@ -34,15 +34,20 @@ export const PlaneImageParams = {
...
+    axis: PD.Select('c', PD.arrayToOptions(['a', 'b', 'c'] as const), { isEssential: true, hideIf: p => p.mode !== 'frame' }),
     rotation: PD.Group({
         axis: PD.Vec3(Vec3.create(1, 0, 0), {}, { description: 'Axis of rotation' }),
         angle: PD.Numeric(0, { min: -180, max: 180, step: 1 }, { immediateUpdate: true, description: 'Axis rotation angle in Degrees' }),
-    }, { isExpanded: true }),
+    }, { isExpanded: true, hideIf: p => p.mode !== 'frame' }),
+    plane: PD.Group({
+        point: PD.Vec3(Vec3.create(0, 0, 0), {}, { description: 'Plane point' }),
+        normal: PD.Vec3(Vec3.create(1, 0, 0), {}, { description: 'Plane normal' }),
+    }, { isExpanded: true, hideIf: p => p.mode !== 'plane' }),
```
dsehnal, line 47: This needs 6 numbers... options:
- use `Vec4` to represent it as `ax + by + cz + d = 0`. Would require adding `Vec4` param definition + UI too
- change `point` to `constant` and we will get the same interface as `Plane3D`

dsehnal (commented):
Nice

giagitom (approved):
Great, I tested it and it is working well on my side

src/mol-geo/geometry/image/image.ts
arose: Because it does not directly get reassigned when cornerBuffer changes, only when the boundingSphere getter is called and when we assign our own boundingSphere we don't want to trigger to calculation based on cornerBuffer.

src/mol-geo/geometry/image/image.ts
arose: thanks

src/mol-repr/structure/visual/plane-image.ts
arose: I'd like to keep normal & point form here as it is easier to put in/reason about manually.

PR #1530, opened 2025-05-27
MVS generic color themes

midlik:
# Description

Support for generic data-driven color themes in MVS (enables mapping data from a CIF column to color). This is implemented by `color_from_source` and `color_from_uri` nodes having a new optional parameter `palette`, which can be 
- `{ kind: 'categorical', ... }` for categorical variables, e.g. `label_atom_id`
- `{ kind: 'discrete', ... }` for numeric variables colored by fixed ranges, e.g. pLDDT
- `{ kind: 'continuous', ... }` for numeric variables colored by continuous scale, e.g. `label_seq_id`
- `null` (default) for variables containing the color string directly (current behavior)

Also adds `field_remapping` param to all `*_from_*` node types to allow loading annotations with non-standard column names.

Open questions:
1. We support explicit color lists and dictionaries, plus named "presets". List presets contain all palettes from D3.js (https://observablehq.com/@d3/color-schemes) plus 'Chainbow'. Dictionary presets contain 'ElementSymbol', 'ResidueName', 'ResidueProperties'. **Can you think about any other useful presets we should add?**
2. Our dictionary presets are taken from various sources. **Should we keep simple naming or be more explicit?** (JmolElementSymbol, JmolResidueName, ClustalResidueProperties)
3. Having a single object-typed `palette` parameter is ugly because 
  a) defaults for nested parameters cannot be defined directly in parameter definition and added automatically as for top-level parameters, 
  b) display of complex object types is very hard to read (in auto-generated docs from `mvs-print-schema` and in validation error messages), 
  c) nested parameters cannot have their own description which would be shown in the auto-generated docs. 
  We should either flatten the parameter hierarchy (will be messy, meh), or implement some kind of nested parameter handling (more work for me, meh).

Notes (@arose , please confirm if these changes are OK):
- I modified Molstar's internal color lists to align with standard names and avoid confusion:
  - Renamed 'inferno' -> 'inferno-no-black', 'magma' -> 'magma-no-black', 'turbo' -> 'turbo-no-black'
  - Added 'inferno', 'magma', 'turbo', synchronized with D3.js (include black region in the beginning)
  - Renamed 'rainbow' -> 'simple-rainbow' (only 5 colors: blue-green-yellow-orange-red)
  - Added 'rainbow', synchronized with D3.js 
  - 'sequence-id' color theme had 'turbo' as default, changed to 'turbo-no-black' (i.e. no change except for renaming)
[...]

PR #1538, opened 2025-06-09
MolViewSpec as Standalone / Publishable package

zachcp:
## Background 

This PR is a follow up on a thread with @dsehnal where we briefly  discuss how to easily [expose the Typescript version of MolViewSpec](https://github.com/molstar/mol-view-stories/issues/5#issuecomment-2927329406).  This PR is intended as a proof of principle that can be further developed if the idea/approach makes sense.

## The Problem Statement

Molstar is extraordinarily powerful but complicated, whereas MolViewSpec is almost as powerful but  far more understandable/teachable.  Can we provide TS/JS developers a tool to accelerate their use of MolViewSpec by providing a simple, standalone tool that they can import with similar ergonomics as the fantastic MVS Python package?  

## Technical Considerations

- prefer ES6 import syntax
- prefer zero dependencies and tight scope - e.g. its only role is to generate valid MVSJ.
- prefer living in the molstar repo (@dsehnal ) and sharing tests/release cycle etc.
- prefer identical interface to the python molviewspec library - ( ideally they would share the same interfaces )
- explore literate/API documentation of [JSR](https://jsr.io/docs/introduction)

## This PR

- create `packages/molviewspec` folder for the MVSData code
- updates top-level package.json to now about that folder.
- uses ES6/import maps and no build step.
- creates a README.md mirrored on the python lib.

### Package Layout

- move extractable portion of `src/extensions/mvs` to `packages/molviewspec`
- create package-specific metatdata / Readme

```bash
tree -L3 packages
packages
└── molviewspec
    ├── package.json
    ├── Readme.md
    └── src
        ├── index.ts
        ├── mvs-data.ts
        ├── tree
        └── util
```

### Example Usage.

```ts
// use in browser or anywhere.
import { MVSData } from 'molviewspec';
const mvsjString = '{"root": {...}, "metadata": {...}}';
const mvsData = MVSData.fromMVSJ(mvsjString);
```

### DIFF
--- a/src/extensions/mvs/camera.ts
+++ b/src/extensions/mvs/camera.ts
@@ -5,28 +5,28 @@
  * @author David Sehnal <david.sehnal@gmail.com>
  */
 
-import { Camera } from '../../mol-canvas3d/camera';
-import { Canvas3DParams, Canvas3DProps } from '../../mol-canvas3d/canvas3d';
-import { Vec3 } from '../../mol-math/linear-algebra';
+import { Camera } from "../../mol-canvas3d/camera";
+import { Canvas3DParams, Canvas3DProps } from "../../mol-canvas3d/canvas3d";
+import { Vec3 } from "../../mol-math/linear-algebra";
 import {
-    getFocusSnapshot,
-    getPluginBoundingSphere,
-} from '../../mol-plugin-state/manager/focus-camera/focus-object';
-import { PluginCommands } from '../../mol-plugin/commands';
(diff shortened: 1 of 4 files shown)


dsehnal:
Thanks @zachcp, I think something like this could be very useful and going forward more smaller packages like this could 
/should be added.

@midlik currently has quite big MVS PR #1530 up that adds more coloring options... so will need to sync this.

Regarding the imports, I think few are fine. For example, I think the schema validation should be a part of this.

@midlik what do you think about this?

Issue #1552, opened 2025-06-26
issue with isosurface using transformed volumes


giagitom:
I'm having this issue when applying  a transformation matrix on a volume. Many normals are wrong. 

Any idea why? 

I am using this to apply the transformation:
```ts
function (volume: Volume, mat: Mat4) {
        const originMat = Grid.getGridToCartesianTransform(volume.grid);
        return Object.assign({}, volume, {
            grid: {
                transform: {
                    kind: 'matrix',
                    matrix: Mat4.mul(Mat4(), mat, originMat)
                },
                cells: volume.grid.cells,
                stats: volume.grid.stats
            }
        });
    }
```

![Image](https://github.com/user-attachments/assets/fd8a7264-8600-4375-a0fd-ecae7756f6d3)

giagitom:
> Does that happen even when `mat` is identity?

No, with identity it is working well

giagitom:
Oh.. It seems a shader issue than, it is only visible when using gpu. I can reproduce on both firefox and chrome (Ubuntu OS)

> Also is you transform just rotation+translation? 

Yes, it should be only translation + rotation. The matrix is derived from a structure alignment win an other structure.

PR #1554, opened 2025-06-28
Fix isosurface compute shader normals when transformation matrix is applied to volume 

giagitom:
…adjoint

# Description

Fix for #1552

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Note that since we don't clearly distinguish between a public and private interf
   - Canvas node: support custom properties `molstar_enable_outline`, `molstar_enable_shadow`, `molstar_enable_ssao`
 - Renamed some color schemes ('inferno' -> 'inferno-no-black', 'magma' -> 'magma-no-black', 'turbo' -> 'turbo-no-black', 'rainbow' -> 'simple-rainbow')
 - Added new color schemes, synchronized with D3.js ('inferno', 'magma', 'turbo', 'rainbow', 'sinebow', 'warm', 'cool', 'cubehelix-default', 'category-10', 'observable-10', 'tableau-10')
+- Fix isosurface compute shader normals when transformation matrix is applied to volume 
 
 ## [v4.18.0] - 2025-06-08
 - MolViewSpec extension:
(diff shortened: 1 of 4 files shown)


dsehnal (commented):
How does a rotation+translation transform cause this singularity?

Isn't there something wrong with your xform matrix?

dsehnal (commented):
Also, the adjoint is computed only once: would it make more sense to do this just once on the CPU?

CC @arose

giagitom:
> How does a rotation+translation transform cause this singularity?
> 
> Isn't there something wrong with your xform matrix?

I am using this to calculate the transformation matrix https://github.com/giagitom/molstar/blob/385357e72c6474f4ff30c4b5393c0b1a2f9790ed/src/mol-model/structure/structure/util/superposition.ts#L37

This should be correct since it is working well in aligning the structures. But I just noticed that `Grid.getGridToCartesianTransform(volume.grid);` that I am using to retrieve the transformation matrix for the volume should apply some scaling factor when 'spacegroup' kind transform is present. Maybe is that the cause?

dsehnal:
> Maybe is that the cause?

But with your matrix being identity you didn't observe the issue, right?

giagitom:
You are right, I discovered that adjoint was not the problem.

I just had to fix the multiplication in the isosurface calculation shader and changed 

**this**
```ts
gl_FragData[2].xyz *= adjoint(uGridTransform); 
```
**to**

```ts
gl_FragData[2].xyz = adjoint(uGridTransform) * gl_FragData[2].xyz;
```

dsehnal:
And since you are working on this, can you compute the adjoint on the CPU and store it as an uniform? Seems like a worthwhile optimization

giagitom:
> And since you are working on this, can you compute the adjoint on the CPU and store it as an uniform? Seems like a worthwhile optimization

Ok good, did this in the last commits

arose (approved):
thanks, lgtm

arose:
> You are right, I discovered that adjoint was not the problem.
> 
> I just had to fix the multiplication in the isosurface calculation shader and changed
> 
> **this**
> 
> ```ts
> gl_FragData[2].xyz *= adjoint(uGridTransform); 
> ```
> 
> **to**
> 
> ```ts
> gl_FragData[2].xyz = adjoint(uGridTransform) * gl_FragData[2].xyz;
> ```

I wonder if this is a gpu driver bug or

PR #1557, opened 2025-06-30
instanced volume

arose:
# Description

Add support for instanced volume.
- add Volume.instances
- add Volume.InstanceIndex and Volume.SegmentIndex types
- volume loci improvements
- add volume-instance color theme

## Actions

### DIFF
--- a/src/mol-model/volume/volume.ts
+++ b/src/mol-model/volume/volume.ts
@@ -2,6 +2,7 @@
  * Copyright (c) 2020-2025 mol* contributors, licensed under MIT, See LICENSE file for more info.
  *
  * @author Alexander Rose <alexander.rose@weirdbyte.de>
+ * @author David Sehnal <david.sehnal@gmail.com>
  */
 
 import { Grid } from './grid';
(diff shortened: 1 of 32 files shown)


src/mol-model/volume/volume.ts
```diff
@@ -146,40 +153,41 @@ export namespace Volume {
...
         grid: Grid.One,
+        instances: [],
         sourceData: { kind: '', name: '', data: {} },
         customProperties: new CustomProperties(),
         _propertyData: Object.create(null),
     };
 
     export function areEquivalent(volA: Volume, volB: Volume) {
-        return Grid.areEquivalent(volA.grid, volB.grid);
+        return Grid.areEquivalent(volA.grid, volB.grid) && deepEqual(volA.instances, volB.instances);
```
dsehnal: @arose needed to add this change otherwise the visuals would not dynamically update. Is there a way to change the instances without re-creating everything?

dsehnal (approved):
Cool, thanks for adding this. Added a state transform support this too.

src/mol-model/volume/volume.ts
arose: handled with [064d572](https://github.com/molstar/molstar/pull/1557/commits/064d572d6fcf7ebf60ce82baf478fd2428c1aa49)

PR #1561, opened 2025-07-02
MVS operator_name

midlik:
# Description

Support for `instance_id` in inline selectors and MVS annotations. This is used to select individual symmetry instances in crystal symmetry structures and assembly structures.

This goes together with https://github.com/molstar/mol-view-spec/pull/84.

This PR also changes the way symmetry instances are named (SymmetryOperator.name) and introduces alternative identifier (SymmetryOperator.canonicalName).

#### SymmetryOperator.name
- Stick to symmetry operators like `1_555` described in [mmCIF dictionary](https://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v50.dic/Items/_struct_conn.ptnr1_symmetry.html):
- Any translation index <0 or >9 will always be enclosed in parenthesis:
    - e.g. `1_(11)15`, `1_1(11)5`, `1_11(15)` instead of ambiguous `1_1115`
    - e.g. `1_(-1)1(-1)`
    - e.g. `2_454` (stays unchanged)

#### SymmetryOperator.canonicalName
- Identical to SymmetryOperator.name

#### SymmetryOperator.name
- Stay as is, e.g. `ASM_1`, `ASM_2`... `ASM_1680` (for compatibility reasons)

#### SymmetryOperator.canonicalName
- Where only one operator is applied to create the symmetry instance, the name of the symmetry instance will be `ASM-` plus the operator identifier ([pdbx_struct_oper_list.id](https://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v50.dic/Items/_pdbx_struct_oper_list.id.html))
  - e.g. `ASM-1`, `ASM-2`, `ASM-3`, `ASM-4` from generator expression `1,2,3,4`
  - e.g. `ASM-1`, `ASM-2`, `ASM-3`, `ASM-4`, `ASM-5` from generator expression `(1-5)`
- Where multiple operators are applied, the name of the symmetry instance will be `ASM-` plus a dash-separated list of operator identifiers. The order of the operators is the same as in the [generator expression](https://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v50.dic/Items/_pdbx_struct_assembly_gen.oper_expression.html), i.e. rightmost operator is applied first.
  - e.g. `ASM-X0-1`, `ASM-X0-2`... `ASM-X0-20` from generator expression `(X0)(1-20)`
  - e.g. `ASM-1-61`, `ASM-1-62`... `ASM-2-61`, `ASM-2-61`... `ASM-60-88` from generator expression `(1-60)(61-88)`

### Inline selectors

(component, color, label, tooltip, primitives)

- example-operator_name-inline-assembly-1m4x.mvsj
- example-operator_name-inline-crystal-1ejg.mvsj
- example-operator_name-inline-crystal-1ejg-big.mvsj

### MVS annotations

(component_from_uri, color_from_uri, label_from_uri, tooltip_from_uri)

- example-operator_name-in-annotation-1tqn.mvsx

[testing-files-v2.zip](https://github.com/user-attachments/files/21143737/testing-files-v2.zip)

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ Note that since we don't clearly distinguish between a public and private interf
   - Representation node: support custom property `molstar_reprepresentation_params`,
   - Canvas node: support custom properties `molstar_enable_outline`, `molstar_enable_shadow`, `molstar_enable_ssao`
   - `clip` node support for structure and volume representations
-  - Inline selectors and MVS annotations support `operator_name`
+  - Inline selectors and MVS annotations support `instance_id`
 - [Breaking] Renamed some color schemes ('inferno' -> 'inferno-no-black', 'magma' -> 'magma-no-black', 'turbo' -> 'turbo-no-black', 'rainbow' -> 'simple-rainbow')
 - Added new color schemes, synchronized with D3.js ('inferno', 'magma', 'turbo', 'rainbow', 'sinebow', 'warm', 'cool', 'cubehelix-default', 'category-10', 'observable-10', 'tableau-10')
 - Snapshot Markdown improvements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,8 +26,8 @@ Note that since we don't clearly distinguish between a public and private interf
   - Indicate external links with ⤴
 - Avoid calculating rings for coarse-grained structures
 - Fix isosurface compute shader normals when transformation matrix is applied to volume
-- Change symmetry operator naming for assemblies (ASM_1, ASM_2 -> 1, X0-1) - TODO update
 - Symmetry operator naming for spacegroup symmetry - parenthesize multi-character indices (1_111-1 -> 1_(11)1(-1))
+- Add canonical operator name (e.g. ASM-1, ASM-X0-1 for assemblies, 1_555, 1_(11)1(-1) for crystals)
 - [Breaking] `PluginContext.initViewer/initContainer/mount` are now async and have been renamed to include `Async` postfix
 - Mol2 Reader
     - Fix column count parsing
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Note that since we don't clearly distinguish between a public and private interf
   - Annotation field remapping (`field_remapping` parameter for color_from_* nodes)
   - Representation node: support custom property `molstar_reprepresentation_params`,
   - Canvas node: support custom properties `molstar_enable_outline`, `molstar_enable_shadow`, `molstar_enable_ssao`
+  - Inline selectors and MVS annotations support `operator_name`
 - Renamed some color schemes ('inferno' -> 'inferno-no-black', 'magma' -> 'magma-no-black', 'turbo' -> 'turbo-no-black', 'rainbow' -> 'simple-rainbow')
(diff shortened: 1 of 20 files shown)


dsehnal (commented):
Looks good overall. Will you do a corresponding Python PR too?

src/mol-model-formats/structure/property/assembly.ts
```diff
@@ -118,7 +118,8 @@ function getAssemblyOperators(matrices: Matrices, operatorNames: string[][], sta
             Mat4.mul(m, m, matrices.get(op[i])!);
         }
         index++;
-        operators[operators.length] = SymmetryOperator.create(`ASM_${index}`, m, { assembly: { id: assemblyId, operId: index, operList: op } });
+        const opName = op.join('-');
```
dsehnal: I wonder: should we still include the `ASM_` prefix?

CHANGELOG.md
```diff
@@ -13,11 +13,14 @@ Note that since we don't clearly distinguish between a public and private interf
   - Annotation field remapping (`field_remapping` parameter for color_from_* nodes)
   - Representation node: support custom property `molstar_reprepresentation_params`,
   - Canvas node: support custom properties `molstar_enable_outline`, `molstar_enable_shadow`, `molstar_enable_ssao`
+  - Inline selectors and MVS annotations support `operator_name`
 - Renamed some color schemes ('inferno' -> 'inferno-no-black', 'magma' -> 'magma-no-black', 'turbo' -> 'turbo-no-black', 'rainbow' -> 'simple-rainbow')
 - Added new color schemes, synchronized with D3.js ('inferno', 'magma', 'turbo', 'rainbow', 'sinebow', 'warm', 'cool', 'cubehelix-default', 'category-10', 'observable-10', 'tableau-10')
 - Avoid calculating rings for coarse-grained structures
 - Fix isosurface compute shader normals when transformation matrix is applied to volume 
 - Breaking: `PluginContext.initViewer/initContainer/mount` are now async and have been renamed to include `Async` postfix
+- Change symmetry operator naming for assemblies (ASM_1, ASM_2 -> 1, X0-1)
```
dsehnal: This could be considered a "minor breaking" change

dsehnal:
@midlik: @arose  pointed out that this would break rcsb-molstar. I found it being used here: https://github.com/molstar/rcsb-molstar/blob/1ba9b9adea968173032d8b4009bcb28536ca5f91/src/viewer/helpers/selection.ts#L102

It seems like the only change needed there should be to return `assembly.name` instead, right?

arose:
This is to a) have better names and b) to have a stricter way to do mol script expressions, right?

I am all for better names. For assemblies we probably still want to have some prefix. The operator names is used in the UI.

Apart from operator.key, operator.name is the only way to do mol script selection by operator. So changing operator.name will break existing mol script expressions (in states/session and in code) - unfortunately, silently. 

I am not sure encoding custom expression in the operator name is a good thing and we probably should not have used it this way in the first place. There are a few properties in SymmetryOperator that we should expose to mol script instead I think.

If we really want a string-based encoding of the 'canonical' (assembly, ncs, hkl) SymmetryOperator properties it should be a new property.

dsehnal:
@arose 
> This is to a) have better names and b) to have a stricter way to do mol script expressions, right?
The main motivation for this was to be able to reference assembly instances from MolViewSpec.

src/mol-model-formats/structure/property/assembly.ts
midlik: I'm including `ASM-` prefix, as we discussed. Name and canonicalName can be easily distinguished by the prefix (`ASM_` vs `ASM-`)

PR #1583, opened 2025-07-26
fix async buffer issues

arose:
# Description

- mark pick-helper dirty when async pick failed
- add pixel-pack buffer wrapper
- recover pixel-pack buffer after context loss (pick buffer, hi-z pass)

## Actions

### DIFF
--- a/src/mol-canvas3d/passes/pick.ts
+++ b/src/mol-canvas3d/passes/pick.ts
@@ -16,6 +16,7 @@ import { Renderbuffer } from '../../mol-gl/webgl/renderbuffer';
 import { Texture } from '../../mol-gl/webgl/texture';
 import { Vec3 } from '../../mol-math/linear-algebra';
 import { isDebugMode, isTimingMode } from '../../mol-util/debug';
+import { now } from '../../mol-util/now';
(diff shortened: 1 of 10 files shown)


src/mol-gl/webgl/buffer.ts
```diff
@@ -258,4 +266,63 @@ export function createElementsBuffer(gl: GLRenderingContext, array: ElementsType
...
+
+        _type,
+        _format,
+        _bpe,
+
+        read,
+        getSubData,
+
+        reset: () => {
+            _buffer = getBuffer(gl);
```
dsehnal, line 320: Should this delete the previous one? Something like

```ts
if (!destroyed) gl.deleteBuffer(_buffer);
destroyed = false;
```

--------

Also, there is one more place where the reset just re-creates the buffer object.

src/mol-canvas3d/passes/pick.ts
```diff
@@ -377,32 +388,24 @@ export class PickBuffers {
             this.fenceSync = null;
             this.lag = 0;
             this.ready = false;
-            return true;
+            return AsyncPickStatus.Failed;
```
dsehnal, line 398: If you call check too many times too fast, it will fail. Should there be a one frame time tolerance like I added in the other PR?

PR #1585, opened 2025-07-27
Scene scale

arose:
# Description

Adds a camera state property `scale` that will scale the scene in "GPU land". This needed for XR to scale the scene to fit into the viewers world, fit with its perception. We do it only in "GPU land" so the application level does not need to know about it.

The scale is exposed as a parameter in this PR but that is just to allow easy testing before actual XR support lands. Changing it should essentially not change anything visually (only later when in XR).

## Actions


giagitom (approved):
@arose, I haven't noted any issues except that when scale is set to 0.001 occlusion seems to disappear.

arose:
> @arose, I haven't noted any issues except that when scale is set to 0.001 occlusion seems to disappear.

I see that too, seems to start around 0.1. Not sure why. I will make an issue for now.

Issue #1572, opened 2025-07-16
Molstar 5.0 dev release?


zachcp:
Hi Team, 

I love the work you've been doing. Any chance we could get a 5.0 dev release? 

zach cp

dsehnal:
<img width="624" height="304" alt="Image" src="https://github.com/user-attachments/assets/373a75b4-2994-4feb-85d5-12b25a06c1f2" />

Will do one soon, likely during the weekend. Wanna get few more MVS releated features in.

dsehnal:
https://github.com/molstar/molstar/releases/tag/v5.0.0-dev.0

dsehnal:
Done

zachcp:
Just testing this today. Hitting the same issue. with `patch.1.` 

```sh
error: [ERR_MODULE_NOT_FOUND] Cannot find module 'file:///Users/zcpowers/Library/Caches/deno/npm/registry.npmjs.org/io-ts/2.2.22/lib/PathReporter' imported from 'file:///Users/zcpowers/Library/Caches/deno/npm/registry.npmjs.org/molstar/5.0.0-dev.1/lib/extensions/mvs/tree/generic/field-schema.js'
Did you mean to import with the ".js" extension?
```

dsehnal:
@zachcp  see https://github.com/molstar/molstar/pull/1595
Is there a way to test this locally in Deno?

dsehnal:
Would including the .ts files in the NPM package help?

And generally what's the issue with this? The relative imports?

zachcp:
> Would including the .ts files in the NPM package help?

Yes. Actually this would help. Right now I am using the npm lib after the build step. But we already have TS files so that would be a better format. The main current blocker there is......

> And generally what's the issue with this? The relative imports?

The relative imports is the biggest immediate issue.  When we [added  `tsc-alias`](https://github.com/molstar/molstar/blob/master/package.json#L26) we fixed the imports after the build but a better fix is to update the imports in each file.  At that point I think that the  `src` directory should then be ECMAScript module-compatible

zachcp:
Note: i wrote up a longer discussion  in #1599

PR #1586, opened 2025-07-27
Allow tsc-alias to modify all files under src/

zachcp:
# Description

Attempt to address the issue in raised in  #1572 where aliases have not been fixed for ESM imports in deno/stypescript environments.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -22,7 +22,6 @@
         "outDir": "lib"
     },
     "include": [ "src/**/*" ],
-    "exclude": [ "src/**/_spec/*" ],
     "tsc-alias": {
         "resolveFullPaths": true,
         "verbose": true


dsehnal:
This doesn't change anything about the spec files, it just includes the tests in the /lib folder.

zachcp:
Ah. yes I was too hasty. this is outside of the tsc-alias call. 


```json
    "include": [ "src/**/*" ],
    "exclude": [ "src/**/_spec/*" ],
    "tsc-alias": {
        "resolveFullPaths": true,
        "verbose": true
    }
```

 But you're right. The issue is incompatability with ESM imports in a dependent library. `tsc-alias` wouldn't work there anyway. Maybe they will accept a patch.

zachcp:
Trying again in #1587

Issue #1559, opened 2025-07-01
Dot representation issue with negative isovalues


giagitom:
It seems there is something wrong with dot representation at negative isovalues. Here is a comparison at -3.0 isovalue between isosurface and dot. PDB ID 2E4A

**isosurface**
![Image](https://github.com/user-attachments/assets/30058ebd-ed4c-475d-b944-e44558b418c2)

**dot**
![Image](https://github.com/user-attachments/assets/d8dd2825-9e9f-459a-8f1e-d31e1387ee07)

PR #1589, opened 2025-07-28
Dot volume representation improvements

giagitom:
# Description

### Dot volume representation improvements

- Add positional perturbation to avoid camera artifacts
- Fix handling of negative isoValues by considering only volume cells with values lower than isoValue (#1559)
- Fix volume-value size theme

**Previous version**
image

**New version**
image

**Uniformed to isosurface the behavior for negative isoValues** 
image
 

## Actions

### DIFF
--- a/src/mol-repr/volume/dot.ts
+++ b/src/mol-repr/volume/dot.ts
@@ -128,12 +128,10 @@ export function createVolumeSphereImpostor(ctx: VisualContext, volume: Volume, k
     const count = Math.ceil((xn * yn * zn) / 10);
     const builder = SpheresBuilder.create(count, Math.ceil(count / 2), spheres);
 
-    const { perturbPositions } = props;
-    const l = Volume.Cell.Location(volume);
     const invert = isoVal < 0;
 
     // Precompute basis vectors and largest cell axis length
-    const basis = perturbPositions ? getBasis(gridToCartn) : undefined;
+    const basis = props.perturbPositions ? getBasis(gridToCartn) : undefined;
 
     for (let z = 0; z < zn; ++z) {
         for (let y = 0; y < yn; ++y) {
@@ -142,12 +140,10 @@ export function createVolumeSphereImpostor(ctx: VisualContext, volume: Volume, k
                 if (!invert && value < isoVal || invert && value > isoVal) continue;
 
                 const cellIdx = space.dataOffset(x, y, z);
-                if (perturbPositions) {
-                    l.cell = cellIdx as Volume.CellIndex;
-
+                if (basis) {
                     Vec3.set(p, x, y, z);
                     Vec3.transformMat4(p, p, gridToCartn);
-                    const offset = getRandomOffsetFromBasis(basis!);
+                    const offset = getRandomOffsetFromBasis(basis);
                     Vec3.add(p, p, offset);
                 } else {
                     Vec3.set(p, x, y, z);
@@ -177,13 +173,12 @@ export function createVolumeSphereMesh(ctx: VisualContext, volume: Volume, key: 
     const vertexCount = count * sphereVertexCount(detail);
     const builderState = MeshBuilder.createState(vertexCount, Math.ceil(vertexCount / 2), mesh);
 
-    const { perturbPositions } = props;
     const l = Volume.Cell.Location(volume);
     const themeSize = theme.size.size;
     const invert = isoVal < 0;
 
     // Precompute basis vectors and largest cell axis length
-    const basis = perturbPositions ? getBasis(gridToCartn) : undefined;
+    const basis = props.perturbPositions ? getBasis(gridToCartn) : undefined;
 
     for (let z = 0; z < zn; ++z) {
         for (let y = 0; y < yn; ++y) {
@@ -194,10 +189,10 @@ export function createVolumeSphereMesh(ctx: VisualContext, volume: Volume, key: 
                 const cellIdx = space.dataOffset(x, y, z);
                 l.cell = cellIdx as Volume.CellIndex;
                 const size = themeSize(l) * sizeFactor;
-                if (perturbPositions) {
+                if (basis) {
                     Vec3.set(p, x, y, z);
                     Vec3.transformMat4(p, p, gridToCartn);
-                    const offset = getRandomOffsetFromBasis(basis!);
+                    const offset = getRandomOffsetFromBasis(basis);
                     Vec3.add(p, p, offset);
                 } else {
                     Vec3.set(p, x, y, z);
@@ -251,12 +246,10 @@ export function createVolumePoint(ctx: VisualContext, volume: Volume, key: numbe
     const count = Math.ceil((xn * yn * zn) / 10);
     const builder = PointsBuilder.create(count, Math.ceil(count / 2), points);
 
-    const { perturbPositions } = props;
-    const l = Volume.Cell.Location(volume);
     const invert = isoVal < 0;
 
     // Precompute basis vectors and largest cell axis length
-    const basis = perturbPositions ? getBasis(gridToCartn) : undefined;
+    const basis = props.perturbPositions ? getBasis(gridToCartn) : undefined;
 
     for (let z = 0; z < zn; ++z) {
         for (let y = 0; y < yn; ++y) {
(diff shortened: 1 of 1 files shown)


arose (changes requested):
The positional perturbation is a nice idea and I am in favor of it. However it should be optional, especially when you use an orthographic camera you may want to see the dots lined up.

src/mol-repr/volume/dot.ts
```diff
@@ -26,9 +27,11 @@ import { addSphere } from '../../mol-geo/geometry/mesh/builder/sphere';
 import { sphereVertexCount } from '../../mol-geo/primitive/sphere';
 import { Points } from '../../mol-geo/geometry/points/points';
 import { PointsBuilder } from '../../mol-geo/geometry/points/points-builder';
+import { SizeTheme } from '../../mol-theme/size';
 
 export const VolumeDotParams = {
     isoValue: Volume.IsoValueParam,
+    perturbatePositions: PD.Boolean(false)
```
dsehnal:
```suggestion
    perturbPositions: PD.Boolean(false)
```

src/mol-repr/volume/dot.ts
```diff
@@ -103,14 +110,42 @@ export function createVolumeSphereImpostor(ctx: VisualContext, volume: Volume, k
     const count = Math.ceil((xn * yn * zn) / 10);
     const builder = SpheresBuilder.create(count, Math.ceil(count / 2), spheres);
 
+    const origin = Vec3.transformMat4(Vec3(), Vec3.create(0, 0, 0), gridToCartn);
+    const cs = count >= 4 ? Vec3.create( // Getting volume cell size. Any better way of doing this?
+        Vec3.distance(origin, Vec3.transformMat4(p, Vec3.set(p, 1, 0, 0), gridToCartn)),
+        Vec3.distance(origin, Vec3.transformMat4(p, Vec3.set(p, 0, 1, 0), gridToCartn)),
+        Vec3.distance(origin, Vec3.transformMat4(p, Vec3.set(p, 0, 0, 1), gridToCartn))
+    ) : Vec3.create(0, 0, 0);
```
dsehnal: Maybe add a helper function for this to `Grid` instead, since you use it multiple times?

Also, this could contain a shear component and the cell will not be orthogonal -- will your method work as well then?

arose:
will review later this week

src/mol-repr/volume/dot.ts
giagitom: > Maybe add a helper function for this to Grid instead, since you use it multiple times?

Ok, done.

>Also, this could contain a shear component and the cell will not be orthogonal -- will your method work as well then?

@dsehnal I tried to handle this by shrinking each dimension by `sin(axisAngle)`. It seems to work as expected in the tests I have done so far.

dsehnal: Alternatively, can you use the vectors as a basis for the coordinate system instead of taking just the magnitude and assuming orthogonal space? The pertubation would then be a random vector in that space, so you would essentially have `B x [rand(), rand(), rand()]` as the offset?

giagitom: >Alternatively, can you use the vectors as a basis for the coordinate system instead of taking just the magnitude and assuming orthogonal space? The pertubation would then be a random vector in that space, so you would essentially have B x [rand(), rand(), rand()] as the offset?

Should be done with last commit

src/mol-repr/volume/dot.ts
```diff
@@ -91,26 +97,68 @@ export function VolumeSphereMeshVisual(materialId: number): VolumeVisual<VolumeS
...
+    const scaleFactor = Math.max(0, (cellMax - dotSize) / cellMax);
+    if (scaleFactor <= 0) return Vec3.set(offset, 0, 0, 0);
+
+    const rx = (rand() - 0.5) * scaleFactor;
+    const ry = (rand() - 0.5) * scaleFactor;
+    const rz = (rand() - 0.5) * scaleFactor;
+
+    Vec3.scale(offset, a, rx);
+    Vec3.add(offset, offset, Vec3.scale(Vec3(), b, ry));
+    Vec3.add(offset, offset, Vec3.scale(Vec3(), c, rz));
```
dsehnal: You should not allocate a new vector here.

src/mol-repr/volume/dot.ts
```diff
@@ -91,26 +97,69 @@ export function VolumeSphereMeshVisual(materialId: number): VolumeVisual<VolumeS
...
+    const scaleFactor = Math.max(0, (cellMax - dotSize) / cellMax);
+    if (scaleFactor <= 0) return Vec3.set(offset, 0, 0, 0);
+
+    const rx = (rand() - 0.5) * scaleFactor;
+    const ry = (rand() - 0.5) * scaleFactor;
+    const rz = (rand() - 0.5) * scaleFactor;
+
+    Vec3.scale(offset, a, rx);
+    Vec3.add(offset, offset, Vec3.scale(tmpVec, b, ry));
+    Vec3.add(offset, offset, Vec3.scale(tmpVec, c, rz));
```
arose:
```suggestion
    Vec3.scaleAndAdd(offset, offset, b, ry);
    Vec3.scaleAndAdd(offset, offset, c, rz);
```

src/mol-repr/volume/dot.ts
```diff
@@ -91,26 +97,69 @@ export function VolumeSphereMeshVisual(materialId: number): VolumeVisual<VolumeS
...
+const rand = Math.random;
+const offset = Vec3();
+const tmpVec = Vec3();
+function getRandomOffsetFromBasis(a: Vec3, b: Vec3, c: Vec3, cellMax: number, dotSize: number): Vec3 {
+    const scaleFactor = Math.max(0, (cellMax - dotSize) / cellMax);
+    if (scaleFactor <= 0) return Vec3.set(offset, 0, 0, 0);
+
+    const rx = (rand() - 0.5) * scaleFactor;
+    const ry = (rand() - 0.5) * scaleFactor;
+    const rz = (rand() - 0.5) * scaleFactor;
```
arose: I think I prefer this. Even with a scaleFactor of cellMax * 0.7 I still see dots lining up (1tqn volume with uniform dot size of 0.1).

Not sure about using the dot size here. I find the perturbed position are a nice effect with bigger dots too.

```suggestion
function getRandomOffsetFromBasis(a: Vec3, b: Vec3, c: Vec3, cellMax: number): Vec3 {
    const rx = (rand() - 0.5) * cellMax;
    const ry = (rand() - 0.5) * cellMax;
    const rz = (rand() - 0.5) * cellMax;
```

src/mol-repr/volume/dot.ts
```diff
@@ -130,24 +179,46 @@ export function createVolumeSphereMesh(ctx: VisualContext, volume: Volume, key:
...
     const vertexCount = count * sphereVertexCount(detail);
-    const builderState = MeshBuilder.createState(vertexCount, vertexCount / 2, mesh);
+    const builderState = MeshBuilder.createState(vertexCount, Math.ceil(vertexCount / 2), mesh);
 
+    const perturbateGeometry = props.perturbPositions;
     const l = Volume.Cell.Location(volume);
     const themeSize = theme.size.size;
+    const invert = isoVal < 0;
+
+    // Precompute basis vectors and largest cell axis length
```
arose, line 180: How about a helper

```ts
function extractBasis(m: Mat4) {
    return {
        a: Vec3.create(m[0], m[1], m[2]),
        b: Vec3.create(m[4], m[5], m[6]),
        c: Vec3.create(m[8], m[9], m[10]),
        maxScale: Mat4.getMaxScaleOnAxis(m)
    };
}
```

src/mol-theme/size/volume-value.ts
```diff
@@ -26,9 +27,10 @@ export function VolumeValueSizeTheme(ctx: ThemeDataContext, props: PD.Values<Vol
         const { data } = ctx.volume.grid.cells;
 
         const isLocation = Volume.Cell.isLocation;
+        const abs = Math.abs;
         const size: LocationSize = (location: Location): number => {
             if (isLocation(location)) {
-                return data[location.cell] * props.scale;
+                return abs(data[location.cell]) * props.scale;
```
arose: Please note that behavior in the theme `Description`

src/mol-theme/size/volume-value.ts
```diff
@@ -26,9 +27,10 @@ export function VolumeValueSizeTheme(ctx: ThemeDataContext, props: PD.Values<Vol
         const { data } = ctx.volume.grid.cells;
 
         const isLocation = Volume.Cell.isLocation;
+        const abs = Math.abs;
```
arose: I don't think this is needed for `Math` (or other built-ins)

arose (changes requested):

giagitom:
I should have applied all suggestions

arose (approved):
thanks

PR #1590, opened 2025-07-28
WebXR

arose:
# Description

Works for browsers with immersive AR or VR support. Only tested with a Quest 3 in standalone mode and tethered to a PC. In standalone mode you want to change the default "auto" quality to "medium", "low", ... if looking larger models like ribosomes, to stay at 72 or 90 fps.

Aim is to work as a supplement to non-XR. At any point you can jump into XR and out of it and see the same scene. Only exception is that there is no near/far clipping in XR and for performance reasons all post-processing effects are off by default (there are XR params to change that).

Controls
- B: exit session
- A: toggle pass-through (when available)
- Y: next snapshot
- X: previous snapshot
- stick forward/backward: move scene closer/away
- trigger gesture: scale scene (on session start the scene is scaled to fit 1/3m)
- trigger click: focus
- trigger drag: rotate
- squeeze press: highlight (generally too expensive and also maybe too distracting to have it always on)

Demos
- https://molstar.org/dev/webxr/
- https://molstar.org/dev/webxr/me/

For related issues and PRs, see https://github.com/orgs/molstar/projects/4

## Actions


src/mol-canvas3d/canvas3d.ts
```diff
@@ -445,6 +451,63 @@ namespace Canvas3D {
...
+            fenceSync = null;
+            resizeRequested = true;
+            pause(true);
+            if (xrManager.session) {
+                saveNonXRProps();
+                setXRProps();
+            } else {
+                loadNonXRProps();
+            }
+            animate();
```
dsehnal: This will not work together with the plugin context-driven animation loop (`PluginAnimationLoop`).

I think we will need a `suspend` property in addition to `pause/animate` for this to work.

src/mol-gl/webgl/context.ts
```diff
@@ -355,6 +373,78 @@ export function createContext(gl: GLRenderingContext, props: Partial<{ pixelScal
...
+                    depth: true,
+                    framebufferScaleFactor: options?.resolutionScale ?? 1,
+                });
+                await value.updateRenderState({ baseLayer });
+
+                console.log('fixed foveation', baseLayer.fixedFoveation);
+                baseLayer.fixedFoveation = 1;
+
+                console.log('supportedFrameRates', value.supportedFrameRates);
+                console.log('frameRate', value.frameRate);
```
dsehnal: hide the logs behind debug flag?

src/mol-gl/webgl/context.ts
```diff
@@ -355,6 +373,78 @@ export function createContext(gl: GLRenderingContext, props: Partial<{ pixelScal
...
+                xrSession = value;
+                isContextLost = false;
+                console.log('xrLayer', xrLayer.framebufferWidth, xrLayer.framebufferHeight);
+
+                xrSession.addEventListener('end', () => {
+                    console.log('xrSession ended');
+                    xrLayer = undefined;
+                    xrSession = undefined;
+                    xrSessionChanged.next();
+                });
```
dsehnal: will these need to be unsubscribed?

src/mol-canvas3d/canvas3d.ts
arose: handled in ea17902aa6be6ad503ae7c8e0f0ce97cbe438c74

src/mol-gl/webgl/context.ts
arose: will just clean this up

src/mol-gl/webgl/context.ts
arose: will be cleaned up

arose:
@corredD ready to merge now from my side.

Anyone else want/can test this?

PR #1592, opened 2025-07-30
MVS: Fix MVSInlinePrimitiveData param type

midlik:
# Description

Fixes a Typescript issue for dependants.

## Actions

### DIFF
--- a/src/extensions/mvs/components/primitives.ts
+++ b/src/extensions/mvs/components/primitives.ts
@@ -39,8 +39,9 @@ import { ParamDefinition as PD } from '../../../mol-util/param-definition';
 import { capitalize } from '../../../mol-util/string';
 import { rowsToExpression, rowToExpression } from '../helpers/selections';
 import { collectMVSReferences, decodeColor, isDefined } from '../helpers/utils';
-import { MolstarNode, MolstarSubtree } from '../tree/molstar/molstar-tree';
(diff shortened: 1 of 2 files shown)


src/extensions/mvs/components/primitives.ts
```diff
@@ -97,14 +98,27 @@ export const MVSDownloadPrimitiveData = MVSTransform({
...
     name: 'mvs-inline-primitive-data',
     display: { name: 'MVS Primitives' },
     from: [SO.Root, SO.Molecule.Structure],
     to: MVSPrimitivesData,
     params: {
-        node: PD.Value<MolstarSubtree<'primitives'>>(undefined as any, { isHidden: true }),
+        node: PD.Value<PrimitivesSubtree>({
+            kind: 'primitives',
+            params: addParamDefaults(MVSTreeSchema.nodes.primitives.params, {}),
+        }, { isHidden: true }),
```
dsehnal, line 121:
```suggestion
        node: PD.Value<PrimitivesSubtree>(undefined as any, { isHidden: true }),
```

A value will always be passed here, no need for a default

midlik: It's just nicer to have a valid default instead of `undefined as any`

dsehnal: All this does in this particular case is introduce more things to do by the state reconciler.

PR #1600, opened 2025-08-04
Update TS imports

zachcp:
# Description

This is an draft PR for TS compatibility with Deno. There are 2 commits. 

- The first is to create a `deno.json` file, the deno equivalent of `package.json` 
- the second is to run `deno lint --fix` in the repo.  This results in import changes in over a thousand files.
- this is a followup some comments in #1572  and discussion #1599 .
- As this is a potentially disruptive PR I wanted to bring it to @dsehnal s attention but i don't have an expectation of acceptance.
- `deno.json` could be removed before merging.  
- In another branch that [I was exploring](https://github.com/zachcp/molstar/tree/20250801-deno-test), the deno lints start to add up  - there are thousands of lint warnings for the default lint rules which I can largely eliminate using the following lints. I bring this up only to point out that `deno` has quite an opinionated take on how to write TypeScript and I don't know how many of these changes/lints you would agree with. So I left out the fixes and the lint ignore rules.
- you can checkout this branch and run `deno lint` and `deno format` yourself if you are interested. 

```json
  "lint": {
    "rules": {
      "tags": ["recommended"],
      "exclude": [
        "no-explicit-any",
        "no-namespace",
        "ban-types",
        "no-unused-vars",
        "no-case-declarations",
        "prefer-const",
        "getter-return",
        "require-await",
        "no-fallthrough",
        "jsx-key",
        "no-extra-boolean-cast"
        "prefer-as-const",
        "jsx-void-dom-elements-no-children",
        "jsx-no-useless-fragment",
        "jsx-button-has-type",
        "no-cond-assign",
        "no-empty",
        "no-empty-interface",
        "no-extra-non-null-assertion"
        "no-irregular-whitespace",
        "no-this-alias"
      ]
    },
```


dsehnal:
In general I am ok with doing this. Have you tested if the cli apps work and the changes dont break the commonjs build?

zachcp:
- Attempting to satisfy the desire to include TS extensions in the source without breaking the build.  
- removed the ts changes from `deno lint` to make it easier to see the changes
- Current workflow: 

```sh
git restore . && deno lint --fix
npm run build
//  look for any issues
git add deno.json && git commit -m '<>'
```

- will need to update the tsconfig.json settings to 1) allow TS in the SRC code; remove it from esbuild step; correct links in commonjs.

PR #1606, opened 2025-08-09
MVS: Animations

dsehnal:
# Description

- [x] Hash-based `StateTransform.version`
- [x] More robust animation loop that prevents animation skips during shader compilation
- [x] Add `Hsl` color space
- [x] Support `StateAnimation` in `PluginState`
  - [x] Data model
  - [x] Animation loop
  - [x] Autoplay
  - [x] UI
- [x] MVS Animation node (controls duration, camera behavior, autoplay, ...)
- [x] MVS Transition node
- [x] MVS Animation logic
  - [x] Interpolate scalar
  - [x] Interpolate vector
  - [x] Interpolate rotation
  - [x] Interpolate transform matrix
  - [x] Interpolate color
  - [x] Easing functions
- [x] MVS Changes
  - [x] Add `transform.rotation_center` param
  - [x] Make `canvas.background_color` optional parameter
- [x] MVS Stories example

Python MVS PR: https://github.com/molstar/mol-view-spec/pull/89

https://github.com/user-attachments/assets/ef41ad24-e985-4b58-957a-4c2f6e3cf77d

## Actions


zachcp:
This looks amazing!

dsehnal:
@corredD I had to rename `from/to` to `start/end` because `from` is a reserved keyword in Python.

corredD:
> > the kind: 'transform_matrix' is really great, can I use it to do a spin animation with multiple turn ?
> 
> @corredD I've added `frequency` and `alternate_direction` params to support this (see the update code/video)

this is excellent!  Can you show me how to use it for a full turn rotation ? I am sorry I can't figure out how to do it. I couldn't make  0 to 2pi to be interpolated. If I trick with 2pi - 0.1 it will interpolate on the shortest path.

corredD:
@dsehnal I found a solution that worked for me, I attached the patch file 
[patchrotation.patch](https://github.com/user-attachments/files/21743954/patchrotation.patch)

src/examples/mvs-stories/stories/animation.ts
```diff
@@ -0,0 +1,306 @@
...
+                ref: 'prims-opacity',
+                target_ref: 'prims',
+                duration_ms: 1000,
+                property: 'label_opacity',
+                end: 1,
+            });
+
+
+            // Uncomment this to make 2nd frame render much faster
+            // It will cause shader compilation to happen during the 1st snapshot
```
midlik, line 52: By chance, is there any way of detecting which shaders will be needed during the whole animation and pre-compiling them, so the users don't need to use hacks like this?

I understand this might be too much to want, just bringing up the idea.

src/extensions/mvs/load-generic.ts
```diff
@@ -46,7 +46,7 @@ export function loadTreeVirtual<TTree extends Tree, TContext>(
 ) {
     const updateRoot: UpdateTarget = UpdateTarget.create(plugin, options?.replaceExisting ?? false);
     loadTreeInUpdate(updateRoot, tree, loadingActions, context, options);
-    const stateTree: StateTree = updateRoot.update.getTree();
+    const stateTree: StateTree = updateRoot.update.getTree({ useHashVersion: true });
```
midlik, line 49: Is there a reason for keeping non-useHashVersion implementation?

src/extensions/mvs/tree/mvs/mvs-animation.ts
```diff
@@ -0,0 +1,128 @@
...
+const _Noise = {
+    noise_magnitude: OptionalField(float, 0, 'Magnitude of the noise to apply to the interpolated value.')
+    // support cummulative noise?
+};
+
+const _Common = {
+    target_ref: RequiredField(str, 'Reference to the node.'),
+    property: RequiredField(union(str, list(union(str, int))), 'Value accessor.'),
+    start_ms: OptionalField(float, 0, 'Start time of the transition in milliseconds.'),
+    duration_ms: RequiredField(float, 'End time of the transition in milliseconds.'),
```
midlik: Is it really end time or is it end time - start time?

src/extensions/mvs/tree/mvs/mvs-animation.ts
```diff
@@ -0,0 +1,128 @@
...
+const ScalarInterpolation = {
+    ..._Common,
+    start: OptionalField(nullable(float), null, 'Start value. If unset, parent state value is used.'),
+    end: OptionalField(nullable(float), null, 'End value. If unset, only noise is applied.'),
+    ..._Noise,
+};
+
+const Vec3Interpolation = {
+    ..._Common,
+    start: OptionalField(nullable(list(float)), null, 'Start value. If unset, parent state value is used. Must be array of length 3N.'),
```
midlik: We have a type `Vector3`, you can use it instead of `list(float)`

src/extensions/mvs/tree/mvs/mvs-animation.ts
```diff
@@ -0,0 +1,128 @@
...
+    end: OptionalField(nullable(list(float)), null, 'End value. Must be array of length 3N. If unset, only noise is applied.'),
+    spherical: OptionalField(bool, false, 'Whether to use spherical interpolation.'),
```
midlik: Do i understand correctly that spherical interpolation mean interpolation via the surface of a sphere rather than via a straight line?
What is the use-case for this?
Is it sufficient to have the center of the sphere always [0,0,0], or does the use-case require parametrization for this?

src/extensions/mvs/tree/mvs/mvs-animation.ts
```diff
@@ -0,0 +1,128 @@
...
+    start: OptionalField(nullable(Matrix), null, 'Start value. If unset, parent state value is used.'),
+    end: OptionalField(nullable(Matrix), null, 'End value. If unset, only noise is applied.'),
+    ..._Noise,
+};
+
+const TransformationMatrixInterpolation = {
+    target_ref: RequiredField(str, 'Reference to the node.'),
+    property: RequiredField(union(str, list(union(str, int))), 'Value accessor.'),
+    start_ms: OptionalField(float, 0, 'Start time of the transition in milliseconds.'),
+    duration_ms: RequiredField(float, 'End time of the transition in milliseconds.'),
```
midlik: Use `..._Common`?

src/extensions/mvs/tree/mvs/mvs-animation.ts
```diff
@@ -0,0 +1,128 @@
...
+    scale_frequency: OptionalField(int, 1, 'Determines how many times the scale interpolation loops. Current T = frequency * t mod 1.'),
+    scale_alternate_direction: OptionalField(bool, false, 'Whether to alternate the direction of the interpolation for frequency > 1.'),
+};
+
+const ColorInterpolation = {
+    ..._Common,
+    palette: RequiredField(union(DiscretePalette, ContinuousPalette), 'Palette to sample colors from.'),
+};
+
+export const MVSAnimationSchema = TreeSchema({
```
midlik: This file defines a tree schema which is completely independent from the MVS tree schema, right?
Then it shouldn't be in `mvs` subdirectory. It should be
`src/extensions/mvs/tree/animation/animation-tree.ts`

src/extensions/mvs/tree/mvs/mvs-builder.ts
```diff
@@ -89,6 +93,37 @@ export class Root extends _Base<'root'> implements FocusMixin, PrimitivesMixin {
     focus = bindMethod(this, FocusMixinImpl, 'focus');
     primitives = bindMethod(this, PrimitivesMixinImpl, 'primitives');
     primitives_from_uri = bindMethod(this, PrimitivesMixinImpl, 'primitives_from_uri');
+
+    animation(params: MVSAnimationNodeParams<'animation'> & CustomAndRef = {}): Animation {
```
midlik, line 97: Would be nice to have `animation` method on each builder node, not just the root.

This wouldn't change the shape of underlying MVS format; it would be just the builder's syntax sugar, removing the necessity to use refs, and allowing to keep animations next to their node in the source code.

The method would auto-generate a ref for the current node (if not provided) and add animation with matching ref.

midlik: Sorry, i guess my comment applies to `interpolation`, not `animation`.

src/extensions/mvs/tree/mvs/mvs-tree.ts
```diff
@@ -57,6 +57,8 @@ const TransformParams = SimpleParamsSchema({
     rotation: OptionalField(Matrix, [1, 0, 0, 0, 1, 0, 0, 0, 1], 'Rotation matrix (3x3 matrix flattened in column major format (j*3+i indexing), this is equivalent to Fortran-order in numpy). This matrix will multiply the structure coordinates from the left. The default value is the identity matrix (corresponds to no rotation).'),
     /** Translation vector, applied to the structure coordinates after rotation. The default value is the zero vector (corresponds to no translation). */
     translation: OptionalField(Vector3, [0, 0, 0], 'Translation vector, applied to the structure coordinates after rotation. The default value is the zero vector (corresponds to no translation).'),
+    /** Point to rotate the object around. Can be either a 3D vector or dynamically computed object centroid. */
+    rotation_center: OptionalField(nullable(union(Vector3, literal('centroid'))), null, 'Point to rotate the object around. Can be either a 3D vector or dynamically computed object centroid.'),
```
midlik, line 61: thumbs up

src/extensions/mvs/tree/mvs/mvs-tree.ts
```diff
@@ -322,7 +324,7 @@ export const MVSTreeSchema = TreeSchema({
             parent: ['root'],
             params: SimpleParamsSchema({
                 /** Color of the canvas background. Can be either an X11 color name (e.g. `"red"`) or a hexadecimal code (e.g. `"#FF0011"`). */
-                background_color: RequiredField(ColorT, 'Color of the canvas background. Can be either an X11 color name (e.g. `"red"`) or a hexadecimal code (e.g. `"#FF0011"`).'),
+                background_color: OptionalField(nullable(ColorT), null, 'Color of the canvas background. Can be either an X11 color name (e.g. `"red"`) or a hexadecimal code (e.g. `"#FF0011"`).'),
```
midlik: What does `null` mean here? 
Is it a viewer-specific default (i.e. white/black depending on initial settings in case of Molstar)?
Or is it just white? (is such case, we can use `background_color: OptionalField(ColorT, 'white', ...)`)

src/mol-plugin-state/transforms/helpers.ts
```diff
@@ -49,18 +49,48 @@ export function getPlaneDataFromStructureSelections(s: ReadonlyArray<PluginState
...
 
-export function getTransformFromParams(src:
-    | { name: 'matrix', params: { data: Mat4, transpose?: boolean } }
-    | { name: 'components', params: { translation: Vec3, axis: Vec3, angle: number } }
-) {
+const GetTransformState = {
+    center: Vec3(),
+    rotation: Mat4(),
+    translationToCenter: Mat4(),
+    translationFromCenter: Mat4(),
```
midlik, line 56: I guess this is still to be changed to `rotationCenter`

src/mol-plugin/animation-loop.ts
```diff
@@ -11,7 +11,16 @@ import { PluginAnimationManager } from '../mol-plugin-state/manager/animation';
...
 
+const MaxProperFrameDelta = 1000 / 30;
+
 export class PluginAnimationLoop {
+    private lastTickT: number = 0;
+    // Proper time is used to prevent animations from skipping
+    // if there is a blocking operation, e.g., shader compilation
+    // The drawback of this is that sometimes the animation will take
+    // longer than intended, but hopefully that's a reasonable tradeoff
+    private properTimeT: number = 0;
```
midlik, line 22: I think this is fair enough in the UI. Freezing seems better then freezing+skipping.
But when i export animation, the timing will be exact, right?

src/mol-util/color/color.ts
```diff
@@ -116,6 +116,19 @@ export namespace Color {
         return ((r << 16) | (g << 8) | b) as Color;
     }
 
+    const interpolateHcl1 = Hcl.zero();
+    const interpolateHcl2 = Hcl.zero();
+
+    /** Linear interpolation between two colors in HCL space */
+    export function interpolateHcl(c1: Color, c2: Color, t: number): Color {
```
midlik: I don't see this used anywhere, so i guess it's more for the future.
I'm wondering if interpolation in HCL is superior to RGB.

Anyway, this implementation is wrong, as hue is a cyclical property. See ugly case below (#ff8800-#ff0088):

Screenshot 2025-08-13 at 13 47 04

midlik: To be honest, i'm very sceptical about HCL in general.
It tries to tell me that all colors in the first row are the same hue. The same for the second row. 
And also that colors in each column have the same CL and only differ in H.

Screenshot 2025-08-13 at 13 40 09

(first rows  = hcl(270, 100, 0-100), second row =  hcl(300, 100, 0-100))

midlik (commented):
This is super cool.

It's a big chunk of changes. The part that i was able to understand, looks good. 
See a few comments.

src/examples/mvs-stories/stories/animation.ts
dsehnal: We discussed this with Alex. In theory it might be possible to compile them ahead of time.

src/extensions/mvs/load-generic.ts
dsehnal: Well -- I originally wrote this back in 2017-18 so this is old code and I didn't want to go though the exercise of figuring out edge-cases.

src/extensions/mvs/tree/mvs/mvs-animation.ts
dsehnal: This supports flat lists of vectors, so you can pass 6 numbers to it for example.

src/extensions/mvs/tree/mvs/mvs-animation.ts
dsehnal: it's this https://en.wikipedia.org/wiki/Slerp

midlik:
I'm wondering: how does easing work in combination with alternate_direction? E.g. does quad-in become quad-out on the way back?

src/extensions/mvs/tree/mvs/mvs-builder.ts
dsehnal: I thought about it, but this would not work in general because some of the builder methods return `this` instead of a wrapper of the new node.

We could support this later if needed.

src/extensions/mvs/tree/mvs/mvs-tree.ts
dsehnal: It's same as if you don't define the `.canvas()` node at all. 

I made this change to support just doing `.canvas({ custom: { ... } })`

src/mol-plugin-state/transforms/helpers.ts
dsehnal: No, this is just a helper state, it's just missing the word `rotation` in the name.

src/extensions/mvs/tree/mvs/mvs-animation.ts
midlik: Then pls don't call it `Vec3Interpolation`.

If we can interpolate custom-length vectors, that means we can interpolate mesh shapes, which will look cool :)
(not sure about a use-case though)

src/mol-plugin/animation-loop.ts
dsehnal: Yes, this is just for the UI-driven animations.

src/mol-util/color/color.ts
dsehnal: Maybe add `HSL` implementation as well then? Perhaps that could be better?

src/extensions/mvs/load-generic.ts
midlik: 🤣  fair enough

src/extensions/mvs/tree/mvs/mvs-builder.ts
midlik: Oh, good point.
Yeah, it's not necessary for now.

src/extensions/mvs/tree/mvs/mvs-tree.ts
midlik: I checked on this: In the current implementation, not having .canvas() node at all will result in using DefaultCanvasBackgroundColor (white), regardless of what is the current color of the background.
So we can use `background_color: OptionalField(ColorT, 'white', ...))`, as `null` would effectively mean `'white'`.

src/mol-plugin-state/transforms/helpers.ts
midlik: ok.

src/mol-util/color/color.ts
dsehnal: Added HSL.

src/extensions/mvs/tree/mvs/mvs-animation.ts
dsehnal: It's still vec3 interpolation, it can just interpolate multiple vectors at the same time.

dsehnal:
Gonna merge this now and do fixes in a separate PR if needed.

PR #1607, opened 2025-08-12
fix shadow

corredD:
# Description
fix the shadow shader to work on both eye


arose:
thanks, changed it so the bounds are always considered, the shadows where wrong whenever the we were not rendering to the full viewport

PR #1609, opened 2025-08-14
MVS Stories and related updates

dsehnal:
# Description

- [x] Add "Open in Mol*" link
- [x] Add "Edit in MolViewStories" link for story states (will work after https://github.com/molstar/mol-view-stories/pull/41 is merged)
- [x] Better snapshot playback if a transition is present
- [x] Add `createMVSX` helper function to MVS extension
- [x] Support trackball animation in MVS
- [x] Breaking: change camera spin speed to mean "rotations per second"

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,7 @@ Note that since we don't clearly distinguish between a public and private interf
 - Add `StructureInstances` transform
 - `mvs-stories` app
   - Add `story-id` URL arg support
+  - Add `story-session-url` URL arg support
   - Add "Download MVS State" link
   - Add "Open in Mol*" link
   - Add "Edit in MolViewStories" link for story states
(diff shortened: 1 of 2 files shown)


src/examples/mvs-stories/stories/animation.ts
```diff
@@ -37,7 +37,14 @@ A story showcasing MolViewSpec animation capabilities.`,
...
 
-            const anim = builder.animation();
+            const anim = builder.animation({
+                custom: {
+                    molstar_trackball: {
+                        name: 'rock',
+                        params: { speed: 1.5 },
+                    }
+                }
+            });
```
dsehnal, line 47: @corredD one more option for you to play with :)

corredD: this is excellent! Could it be trigger by the markdown ? And can we use duration_ms and start_ms for it ?

dsehnal: The duration and start no. But we could add "rotate vector" interpolation kind and you would just interpolate the position on camera node.

Writing a command in markdown to trigger the animation is quite easy (completely separate from MVS).

corredD (approved):
Nice work! Regarding the rock and spin features, why cant we slow down the spin to 0.1 like the rock ?

PR #1625, opened 2025-08-23
Trackball & Snapshot handling tweaks

arose:
# Description

- add SnapshotControls behavior
- don't save behaviors in mesoscale explorer snapshots (@corredD we will need to remove the "behaviors" from the existing ME session state.json files)
- add canvas3d/trackball attribs that are configurable but are not saved in the state like props

In the WebXR branch I will add bindings for controllers

## Actions

### DIFF
--- a/src/mol-plugin/behavior/dynamic/state.ts
+++ b/src/mol-plugin/behavior/dynamic/state.ts
@@ -37,7 +37,7 @@ export const SnapshotControls = PluginBehavior.create<SnapshotControlsProps>({
     ctor: class extends PluginBehavior.Handler<SnapshotControlsProps> {
         register(): void {
             this.subscribeObservable(this.ctx.behaviors.interaction.keyReleased, ({ code, modifiers, key }) => {
-                if (!this.ctx.canvas3d) return;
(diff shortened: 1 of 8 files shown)


src/mol-plugin/behavior/dynamic/state.ts
```diff
@@ -0,0 +1,70 @@
...
+};
+type SnapshotControlsProps = PD.Values<typeof SnapshotControlsParams>
+
+export const SnapshotControls = PluginBehavior.create<SnapshotControlsProps>({
+    name: 'snapshot-controls',
+    category: 'interaction',
+    ctor: class extends PluginBehavior.Handler<SnapshotControlsProps> {
+        register(): void {
+            this.subscribeObservable(this.ctx.behaviors.interaction.keyReleased, ({ code, modifiers, key }) => {
+                if (!this.ctx.canvas3d) return;
```
dsehnal: I think we should we add the `!isBusy` check here again. This was to prevent messing up the state with too fast updates.

```suggestion
                if (!this.ctx.canvas3d || this.ctx.isBusy) return;
```

arose: excellent, missed that

PR #1631, opened 2025-08-26
MVS: Animation improvements

dsehnal:
# Description

- [x] Support multiple interpolations of a single property within the same snapshot
- [x] Update transition slider to use time instead of frame index
- [x] Do not create new transition frame if it was the same as the previous one


corredD (approved):
Excellent addition, it makes a lot of sense to be able to do theses kind sequential animation in one snapshots.

PR #1635, opened 2025-08-28
Fix renderer transparency check

### LINKED ISSUE #1634: Outline over transparancy
I am not sure if it's the expect behavior, but since the outline is a post-process it doesn't consider the opacity of the geometry. Its more visible with Labels for which even at opacity 0 they still appear in the outline pass.

Image

giagitom:
# Description

This should fix #1634

## Actions

### DIFF
--- a/src/mol-gl/renderer.ts
+++ b/src/mol-gl/renderer.ts
@@ -483,13 +483,13 @@ namespace Renderer {
             const alpha = clamp(r.values.alpha.ref.value * r.state.alphaFactor, 0, 1);
             const xrayShaded = r.values.dXrayShaded?.ref.value === 'on' || r.values.dXrayShaded?.ref.value === 'inverted';
             return (
-                (alpha < 1 && alpha !== 0) ||
+                alpha !== 0 && (alpha < 1 ||
                 r.values.transparencyAverage.ref.value > 0 ||
                 r.values.dGeometryType.ref.value === 'directVolume' ||
                 r.values.dPointStyle?.ref.value === 'fuzzy' ||
                 r.values.dGeometryType.ref.value === 'text' ||
                 r.values.dGeometryType.ref.value === 'image' ||
-                xrayShaded
+                xrayShaded)
             );
         };
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,6 +100,7 @@ Note that since we don't clearly distinguish between a public and private interf
 - Add plugin config item ShowReset (shows/hides "Reset Zoom" button)
 - Fix transform params not being normalized when used together with param hash version
 - Replace `immer` with `mutative`
+- Fix renderer transparency check
 
 ## [v4.18.0] - 2025-06-08
 - MolViewSpec extension:


src/mol-gl/renderer.ts
```diff
@@ -483,13 +483,13 @@ namespace Renderer {
             const alpha = clamp(r.values.alpha.ref.value * r.state.alphaFactor, 0, 1);
             const xrayShaded = r.values.dXrayShaded?.ref.value === 'on' || r.values.dXrayShaded?.ref.value === 'inverted';
             return (
-                (alpha < 1 && alpha !== 0) ||
+                alpha !== 0 && (alpha < 1 ||
```
dsehnal, line 486: Won't this fix it only for the alpha=0 case and still appear weird for say alpha =0.01?

giagitom: > Won't this fix it only for the alpha=0 case and still appear weird for say alpha =0.01?

Hmm yes I know.. How about fading outlines to transparency?
image

dsehnal: Better. But how about giving the outline less transparency? Eg outline alpha = max(3 * base, 1)?

dsehnal: Ok you actually did that but with factor of 2. That could work too.

dsehnal (approved):

src/mol-gl/shader/postprocessing.frag.ts
```diff
@@ -169,13 +169,25 @@ void main(void) {
...
+                    float outlineAlpha = clamp(isTransparentOutline * 2.0, 0.0, 1.0);
+                    vec3 outlineColor = mix(uOutlineColor, uFogColor, fogFactor);                    
+                    color.a = color.a + outlineAlpha * (1.0 - color.a); 
+                    color.rgb = mix(color.rgb, outlineColor, outlineAlpha);
+                } else {
+                    color.a = 1.0 - fogFactor;
+                    color.rgb = mix(uOutlineColor, vec3(0.0), fogFactor);
+                }                
             }
             #ifdef dBlendTransparency
```
corredD: I would propose to move all the #ifdef dBlendTransparancy block before the outline. It avoid blending the outline with the background when opactity are low, if you try opacity 0.04, you will actually see a background color outline...
before 
image
after
image

corredD (commented):
looks really great! can you look at the blending part ? I think it looks better if the blending happen before the outline.

src/mol-gl/shader/postprocessing.frag.ts
giagitom: > I would propose to move all the #ifdef dBlendTransparancy block before the outline. It avoid blending the outline with the background when opactity are low, if you try opacity 0.04, you will actually see a background color outline...

Can't do this directly since It would not work with outlines from opaque objects that would be rendered in front of transparency. 
But I have made some improvements that should solve the issue

giagitom: image
image

corredD (approved):

giagitom:
I'm still working on further improvements. I'll make an other pr when done

dsehnal:
Ok, leave that be then.

PR #1638, opened 2025-08-29
Vol gradient

corredD:
# Description
New representation for volume, show gradient field line as lines or cylinders.

image

image

## Actions

### DIFF
--- a/src/mol-repr/volume/gradient.ts
+++ b/src/mol-repr/volume/gradient.ts
@@ -37,12 +37,12 @@ export const VolumeLinesParams = {
     ...Lines.Params,
     ...Cylinders.Params,
     ...VolumeGradientParams,
-    useCylinder: PD.Boolean(true),
+    useCylinder: PD.Boolean(false),
     detail: PD.Numeric(0, { min: 0, max: 3, step: 1 }, BaseGeometry.CustomQualityParamInfo),
-    seedDensity: PD.Numeric(8, { min: 1, max: 100, step: 1 }, { description: 'Seeds per dimension for gradient lines.' }),
-    maxSteps: PD.Numeric(100, { min: 1, max: 1000, step: 1 }, { description: 'Maximum number of steps for gradient lines.' }),
-    stepSize: PD.Numeric(0.5, { min: 0.01, max: 10, step: 0.01 }, { description: 'Step size for gradient lines.' }),
-    minSpeed: PD.Numeric(0.001, { min: 0, max: 1, step: 0.0001 }, { description: 'Minimum speed for gradient lines.' }),
+    seedDensity: PD.Numeric(20, { min: 1, max: 100, step: 1 }, { description: 'Seeds per dimension for gradient lines.' }),
+    maxSteps: PD.Numeric(1000, { min: 1, max: 2000, step: 1 }, { description: 'Maximum number of steps for gradient lines.' }),
+    stepSize: PD.Numeric(0.35, { min: 0.01, max: 10, step: 0.01 }, { description: 'Step size for gradient lines.' }),
+    minSpeed: PD.Numeric(0.001, { min: 0, max: 1, step: 1e-6 }, { description: 'Minimum speed for gradient lines.' }),
 };
 export type VolumeLinesParams = typeof VolumeLinesParams
 export type VolumeLinesProps = PD.Values<VolumeLinesParams>
@@ -63,7 +63,11 @@ export function VolumeCylindersImpostorVisual(materialId: number): VolumeVisual<
         eachLocation: eachGradient,
         setUpdateState: (state: VisualUpdateState, volume: Volume, newProps: PD.Values<VolumeLinesParams>, currentProps: PD.Values<VolumeLinesParams>, newTheme: Theme, currentTheme: Theme) => {
             state.createGeometry = (
-                !Volume.IsoValue.areSame(newProps.isoValue, currentProps.isoValue, volume.grid.stats)
+                !Volume.IsoValue.areSame(newProps.isoValue, currentProps.isoValue, volume.grid.stats)||
+                newProps.seedDensity !== currentProps.seedDensity ||
+                newProps.maxSteps !== currentProps.maxSteps ||
+                newProps.stepSize !== currentProps.stepSize ||
+                newProps.minSpeed !== currentProps.minSpeed
             );
         },
         geometryUtils: Cylinders.Utils,
@@ -84,9 +88,10 @@ export function VolumeLinesVisual(materialId: number): VolumeVisual<VolumeLinesP
         setUpdateState: (state: VisualUpdateState, volume: Volume, newProps: PD.Values<VolumeLinesParams>, currentProps: PD.Values<VolumeSphereParams>, newTheme: Theme, currentTheme: Theme) => {
             state.createGeometry = (
                 !Volume.IsoValue.areSame(newProps.isoValue, currentProps.isoValue, volume.grid.stats) ||
-                // newProps.perturbPositions !== currentProps.perturbPositions ||
-                newProps.sizeFactor !== currentProps.sizeFactor ||
-                newProps.detail !== currentProps.detail
+                newProps.seedDensity !== currentProps.seedDensity ||
+                newProps.maxSteps !== currentProps.maxSteps ||
+                newProps.stepSize !== currentProps.stepSize ||
+                newProps.minSpeed !== currentProps.minSpeed
             );
         },
         geometryUtils: Lines.Utils,
@@ -170,11 +175,13 @@ export function createVolumeLinesMesh(ctx: VisualContext, volume: Volume, key: n
     // const { min, max } = stats;
     const [nx, ny, nz] = space.dimensions as Vec3;
 
-    // Streamline parameters
-    const seedDensity = props.seedDensity || 8; // seeds per dimension
-    const maxSteps = props.maxSteps || 100;
-    const stepSize = props.stepSize || 0.5;
-    const minSpeed = props.minSpeed || 0.001;
+    const { hx, hy, hz } = getCellSize(gridToCartn);
+
+    // seeds
+    const seedDensity = props.seedDensity ?? 8;
+    const maxSteps = props.maxSteps ?? 2000; // more steps
+    const ds = (props.stepSize ?? 0.35); // in index units; 0.2–0.5 works well
+    const eps = props.minSpeed ?? 1e-6; // MUCH smaller for APBS
 
     // Create color scale
     // const colorScale = ColorScale.create({
@@ -199,32 +206,37 @@ export function createVolumeLinesMesh(ctx: VisualContext, volume: Volume, key: n
     for (let z = seedStep; z < nz - seedStep; z += seedStep) {
         for (let y = seedStep; y < ny - seedStep; y += seedStep) {
             for (let x = seedStep; x < nx - seedStep; x += seedStep) {
-                // Trace streamline from this seed point
-                const streamline = traceStreamline(
-                    space, data, Vec3.create(x, y, z), 
-                    maxSteps, stepSize, minSpeed
+
+                const pos = Vec3.create(x, y, z);
+
+                // Filtrage par potentiel et champ
+                const phi = getInterpolatedValue(space, data, pos);
+                const g = getInterpolatedGradient(space, data, pos, hx, hy, hz);
+                const mag = Vec3.magnitude(g);
+
+                // seuils à ajuster selon tes cartes APBS
+                if (Math.abs(phi) > 0.2 || mag < 1e-6) continue;
+
+                // si ok, on trace la streamline
+                const streamline = traceStreamlineBothDirs(
+                    space, data, pos,
+                    maxSteps, ds, eps, hx, hy, hz
                 );
 
                 if (streamline.length < 2) continue;
                 const cellIdx = space.dataOffset(x, y, z);
 
-                // Convert to cartesian coordinates and add to builder
                 for (let i = 0; i < streamline.length - 1; i++) {
                     const start = Vec3();
                     const end = Vec3();
-
                     Vec3.transformMat4(start, streamline[i].position, gridToCartn);
                     Vec3.transformMat4(end, streamline[i + 1].position, gridToCartn);
-
-                    // Color based on volume value at start point
-                    // const color = colorScale.color(streamline[i].value);
-                    // const group = Color.toRgb(color)[0] * 255; // Use red component as group
-
                     builder.addVec(start, end, cellIdx);
                 }
             }
         }
     }
+
 
     const pt = builder.getLines();
     pt.setBoundingSphere(Volume.Isosurface.getBoundingSphere(volume, props.isoValue));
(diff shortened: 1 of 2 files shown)


src/mol-repr/volume/gradient.ts
```diff
@@ -0,0 +1,436 @@
...
+    const builder = LinesBuilder.create(count, Math.ceil(count / 2), points);
+
+    // Precompute basis vectors and largest cell axis length
+    // const basis = getBasis(gridToCartn);
+
+    for (let z = seedStep; z < nz - seedStep; z += seedStep) {
+        for (let y = seedStep; y < ny - seedStep; y += seedStep) {
+            for (let x = seedStep; x < nx - seedStep; x += seedStep) {
+
+                const pos = Vec3.create(x, y, z);
```
dsehnal: create the vector outside the loop and call `Vec3.set` here

src/mol-repr/volume/gradient.ts
```diff
@@ -0,0 +1,436 @@
...
+
+                // Filtrage par potentiel et champ
```
dsehnal: English please :)

src/mol-repr/volume/gradient.ts
```diff
@@ -0,0 +1,436 @@
...
+                    space, data, pos,
+                    maxSteps, ds, eps, hx, hy, hz
+                );
+
+                if (streamline.length < 2) continue;
+                const cellIdx = space.dataOffset(x, y, z);
+
+                for (let i = 0; i < streamline.length - 1; i++) {
+                    const start = Vec3();
+                    const end = Vec3();
```
dsehnal: No vector allocation inside a loop again

src/mol-repr/volume/gradient.ts
```diff
@@ -0,0 +1,436 @@
...
+
+        const g = getInterpolatedGradient(space, data, p, hx, hy, hz);
+        const m = Vec3.magnitude(g);
+        if (!(m > eps)) break;
+
+        // direction only, signed
+        const v = Vec3.scale(Vec3(), g, dirSign / m);
+
+        // RK4 with constant arc-length ds
+        const k1 = Vec3.scale(Vec3(), v, ds);
```
dsehnal: Move all the vector allocations outside the function (e.g. use `TraceState = { v1: Vec3() }` and use `Vec3.set(TraceState.v1, ...)` etc.

Doing this many allocations inside a loop is a performance killer.

src/mol-repr/volume/gradient.ts
```diff
@@ -0,0 +1,436 @@
...
+}
+
+function traceStreamlineBothDirs(
+    space: Tensor.Space, data: ArrayLike<number>, seed: Vec3,
+    maxSteps: number, ds: number, eps: number,
+    hx: number, hy: number, hz: number
+): StreamlinePoint[] {
+    const back = traceOneDirection(space, data, seed, maxSteps, ds, eps, hx, hy, hz, -1);
+    const fwd = traceOneDirection(space, data, seed, maxSteps, ds, eps, hx, hy, hz, +1);
+    return back.reverse().concat(fwd.length ? fwd.slice(1) : []);
```
dsehnal: It should be possible to do this without allocations too.

src/mol-repr/volume/gradient.ts
```diff
@@ -0,0 +1,436 @@
...
+}
+
+function getInterpolatedGradient(space: Tensor.Space, data: ArrayLike<number>, pos: Vec3, hx: number, hy: number, hz: number): Vec3 {
+    const x = Math.floor(pos[0]), y = Math.floor(pos[1]), z = Math.floor(pos[2]);
+    const fx = pos[0] - x, fy = pos[1] - y, fz = pos[2] - z;
+
+    const grads: Vec3[] = [];
```
dsehnal: This could also be allocated just once with the correct length and instead of `grads.push` use direct assignment to that slot.

src/mol-repr/volume/gradient.ts
```diff
@@ -0,0 +1,436 @@
...
+    for (let dz = 0; dz <= 1; dz++) {
+        for (let dy = 0; dy <= 1; dy++) {
+            for (let dx = 0; dx <= 1; dx++) {
+                grads.push(calculateGradient(space, data, x + dx, y + dy, z + dz, hx, hy, hz));
+            }
+        }
+    }
+
+    const out = Vec3();
```
dsehnal: use `out: Vec3` as param of this function and allocate only 1 Vec3 for this.

src/mol-repr/volume/gradient.ts
```diff
@@ -0,0 +1,436 @@
...
+    const xi = Math.max(1, Math.min(x, (space.dimensions[0] as number) - 2));
+    const yi = Math.max(1, Math.min(y, (space.dimensions[1] as number) - 2));
+    const zi = Math.max(1, Math.min(z, (space.dimensions[2] as number) - 2));
+
+    const gx = (data[space.dataOffset(xi + 1, yi, zi)] - data[space.dataOffset(xi - 1, yi, zi)]) / (2 * hx);
+    const gy = (data[space.dataOffset(xi, yi + 1, zi)] - data[space.dataOffset(xi, yi - 1, zi)]) / (2 * hy);
+    const gz = (data[space.dataOffset(xi, yi, zi + 1)] - data[space.dataOffset(xi, yi, zi - 1)]) / (2 * hz);
+
+    // Electric field E = -∇φ
+    return Vec3.create(-gx, -gy, -gz);
```
dsehnal: `out: Vec3` in the params again

src/mol-repr/volume/gradient.ts
```diff
@@ -0,0 +1,436 @@
...
+import { Interval, OrderedSet } from '../../mol-data/int';
+import { RepresentationContext, RepresentationParamsGetter, Representation } from '../representation';
+
+export const VolumeGradientParams = {
+    isoValue: Volume.IsoValueParam,
+    seedDensity: PD.Numeric(20, { min: 1, max: 100, step: 1 }, { description: 'Seeds per dimension for gradient lines.' }),
+    maxSteps: PD.Numeric(1000, { min: 1, max: 2000, step: 1 }, { description: 'Maximum number of steps for gradient lines.' }),
+    stepSize: PD.Numeric(0.35, { min: 0.01, max: 10, step: 0.01 }, { description: 'Step size for gradient lines.' }),
+    minSpeed: PD.Numeric(0.001, { min: 0, max: 1, step: 1e-6 }, { description: 'Minimum speed for gradient lines.' }),
+};
```
dsehnal, line 58: Which of these parameters would be used to animate the visual like in this one you shared in the notes?

Image

corredD: I think paraView use a particle system that travel along the lines. https://www.kitware.com/new-animated-stream-lines-representation-for-paraview-5-3/
Its an open question to decide how animate it.

corredD: disk305x302

corredD:
> > @giagitom do you have some time to help get this in?
> 
> @dsehnal I'll try my best but I know nothing about the algorithm itself.. I'll try to look into this during next days
> 
> Anyway @corredD this seems a super nice addition!

Thanks @giagitom ! maybe you can help with the color theme ? it seems that the external volume work better than the volume-value theme. I tried to fix it, not sure I did it right. Did you got proper coloring with the Dot representation ?

arose:
Thanks @corredD this looks nice, will review later.

Regarding "pymol algo". Did you copy any code? Because PyMOL is GPL last I checked and we do not want GPL code here.

The isovalue param is not used, right?

corredD:
@arose thanks for looking at it! The open-source parts of PyMOL are under a [BSD-style license](https://github.com/schrodinger/pymol-open-source/blob/a550aa709f45f5a57146f3bd57a8b6dcdc4db21c/LICENSE). No copy/paste has been done here — this is an independent reimplementation, with algorithmic inspiration only. But for sake of transparancy I added a comment 'Algorithm based on PyMOL’s isosurface gradient tracing (Schrödinger LLC)'. 
Code was written from the Dot representation with the help of GPT-5, which generates MIT-licensed code by default ( hope its ok with you ). I removed the isoValue param which is not used. I also changed the color-theme, but I am not sure I did it correctly...

src/mol-repr/volume/gradient.ts
```diff
@@ -0,0 +1,726 @@
...
+    return written;
+}
+
+function reverseSegmentInPlace<T>(arr: T[], start: number, end: number) {
+    for (let i = start, j = end; i < j; i++, j--) { const tmp = arr[i]; arr[i] = arr[j]; arr[j] = tmp; }
+}
+
+function traceStreamlineBothDirs(space: Tensor.Space, data: ArrayLike<number>, seed: Vec3, maxSteps: number, dsWorld: number, eps: number, hx: number, hy: number, hz: number, writeStride: number, order: 2 | 4): StreamlinePoint[] {
+    const line: StreamlinePoint[] = [];
+    const S = { p: Vec3(), g: Vec3(), g2: Vec3(), g3: Vec3(), g4: Vec3(), v1: Vec3(), v2: Vec3(), v3: Vec3(), v4: Vec3(), k1: Vec3(), k2: Vec3(), k3: Vec3(), k4: Vec3(), dp: Vec3(), t1: Vec3(), t2: Vec3(), t3: Vec3() };
```
dsehnal: This should be declared outside the function.

src/mol-repr/volume/gradient.ts
```diff
@@ -0,0 +1,726 @@
...
+        return (l < lo || l > hi) ? { f, l, done: true } : { f, l, done: false };
+    }
+
+    // emit segments similar to PyMOL bookkeeping but directly into out as separate polylines
+    const pushNewPolyline = () => { out.push([]); return out.length - 1; };
+
+    for (let a = 0; a < rangeSize; a++) {
+        const activeCells: number[] = [];
+
+        const walk = (sign: 1 | -1) => {
```
dsehnal: A closure inside a loop is pretty bad for perf.

src/mol-repr/volume/gradient.ts
```diff
@@ -0,0 +1,726 @@
...
+                let r = normalizeFrac(fx, li, i0, i1 - 2); fx = r.f; li = r.l; if (r.done) break;
+                r = normalizeFrac(fy, lj, j0, j1 - 2); fy = r.f; lj = r.l; if (r.done) break;
+                r = normalizeFrac(fz, lk, k0, k1 - 2); fz = r.f; lk = r.l; if (r.done) break;
+
+                if (flag[flagIndex(li, lj, lk)]) break;
+
+                const level = interpolateValueFrac(space, data, li, lj, lk, fx, fy, fz);
+                if (level < props.minLevel || level > props.maxLevel) break;
+
+                const gW = Vec3();
```
dsehnal: No vector allocations inside a loop.

src/mol-repr/volume/gradient.ts
```diff
@@ -0,0 +1,726 @@
...
+    }
+}
+
+// ---------------------------------------------------------------------------------------
+// Field sampling (clamped + world-aware gradients)
+// ---------------------------------------------------------------------------------------
+
+function clampIndex(space: Tensor.Space, x: number, y: number, z: number) {
+    const nx = space.dimensions[0] as number, ny = space.dimensions[1] as number, nz = space.dimensions[2] as number;
+    return [Math.max(0, Math.min(nx - 1, x)), Math.max(0, Math.min(ny - 1, y)), Math.max(0, Math.min(nz - 1, z))] as [number, number, number];
```
dsehnal: `clampIndex` seems to be called inside a loop -- this should get an `out: [number, number, number]` vector instead as a param too.

src/mol-repr/volume/gradient.ts
```diff
@@ -0,0 +1,726 @@
...
+                newProps.mid_interpolation !== currentProps.mid_interpolation ||
+                newProps.algorithm !== currentProps.algorithm
+            );
+        },
+        geometryUtils: Lines.Utils,
+        mustRecreate: (_: VolumeKey, __: PD.Values<VolumeLinesParams>, webgl?: WebGLContext) => !!webgl,
+    }, materialId);
+}
+
+export function VolumeCylindersImpostorVisual(materialId: number): VolumeVisual<VolumeCylindersLinesParams> {
```
arose, line 255: When we have a Cylinder geometry we should have a mesh version as a fallback in case glFragDepth is not supported. Same for the Sphere geometry.

I am not too sure about the cylinder geometry because it will always have ugly overlaps if you use the geomStride=1. Using a tube mesh would avoid that. Maybe we could start just with a lines visual and add other visuals in later PRs?

src/mol-repr/volume/gradient.ts
```diff
@@ -0,0 +1,726 @@
...
+
+export const VolumeGradientParams = {
+    seedDensity: PD.Numeric(10, { min: 1, max: 100, step: 1 }, { description: 'Seeds per dimension for streamlines.' }),
+    maxSteps: PD.Numeric(1000, { min: 1, max: 2000, step: 1 }, { description: 'Maximum number of steps for streamlines.' }),
+    stepSize: PD.Numeric(0.35, { min: 0.01, max: 10, step: 0.01 }, { description: 'Step size (Å) for streamlines.' }),
+    minSpeed: PD.Numeric(0.001, { min: 0, max: 1, step: 1e-6 }, { description: 'Minimum |E| (in 1/Å) to continue tracing.' }),
+    minLevel: PD.Numeric(-5.0, { min: -10, max: 10, step: 0.1 }, { description: 'Minimum map value (φ) allowed when tracing.' }),
+    maxLevel: PD.Numeric(5.0, { min: -10, max: 10, step: 0.1 }, { description: 'Maximum map value (φ) allowed when tracing.' }),
+    algorithm: PD.Select('simple', PD.arrayToOptions(['simple', 'advanced'] as const), { description: 'Streamline algorithm to use.' }),
+    mid_interpolation: PD.Boolean(true, { description: 'Use RK2 (midpoint) instead of RK4 for the simple tracer.' }),
```
arose:
```suggestion
    midInterpolation: PD.Boolean(true, { description: 'Use RK2 (midpoint) instead of RK4 for the simple tracer.', hideIf: (o) => o.algorithm !== 'simple' }),
```

src/mol-repr/volume/gradient.ts
```diff
@@ -0,0 +1,726 @@
...
+// ---------------------------------------------------------------------------------------
+// Loci plumbing
+// ---------------------------------------------------------------------------------------
+
+function getLoci(volume: Volume, props: VolumeGradientProps) {
+    const instances = Interval.ofLength(volume.instances.length as Volume.InstanceIndex);
+    return Volume.Loci(volume, instances);
+}
+
+function getGradientLoci(pickingId: PickingId, volume: Volume, _key: number, props: VolumeGradientProps, id: number) {
```
arose, line 514: We should cache/safe the streamlines (keyed by their params) as a custom volume property. Then we can reference individual streamlines in the loci (will need a new Volume.Streamline.Loci). We need to be careful to cleanup cached streamlines when params change and when the visual gets deleted.

Also, please move the streamlines algo to mol-math/volume/streamlines (and simplify the input so you are not importing things from mol-model or mol-repr in mol-math).

src/mol-theme/color/volume-value.ts
```diff
@@ -89,11 +93,18 @@ export function VolumeValueColorTheme(ctx: ThemeDataContext, props: PD.Values<Vo
                 const value = getTrilinearlyInterpolated(location.position);
                 if (isNaN(value)) return props.defaultColor;
 
-                return (clamp((value - domain[0]) / (domain[1] - domain[0]), 0, 1) * ColorTheme.PaletteScale) as Color;
+                if (usePalette) {
```
arose: Hmm, something seems wrong with either the palette or non-palette version. I will have a look.

src/mol-repr/volume/gradient.ts
```diff
@@ -0,0 +1,726 @@
...
+function createVolumeLinesGeometry(ctx: VisualContext, volume: Volume, _key: number, _theme: Theme, props: VolumeLinesProps, lines?: Lines): Lines {
+    const streamLines = collectStreamlines(volume, props);
+    const gridToCartn = Grid.getGridToCartesianTransform(volume.grid);
+
+    const segCount = countSegments(streamLines, props.geomStride);
+    const builder = LinesBuilder.create(segCount, Math.ceil(segCount / 2), lines);
+
+    writeSegmentsToBuilder(streamLines, gridToCartn, props.geomStride, (a, b, id) => builder.addVec(a, b, id));
+
+    const g = builder.getLines();
```
arose, line 378: we can set the boundingsphere from the volume so it does not need to calculated from the geometry. Should save something like 10%.

src/mol-repr/volume/gradient.ts
```diff
@@ -0,0 +1,726 @@
...
+                newProps.minLevel !== currentProps.minLevel ||
+                newProps.maxLevel !== currentProps.maxLevel ||
+                newProps.writeStride !== currentProps.writeStride ||
+                newProps.geomStride !== currentProps.geomStride||
+                newProps.mid_interpolation !== currentProps.mid_interpolation ||
+                newProps.algorithm !== currentProps.algorithm
+            );
+        },
+        geometryUtils: Lines.Utils,
+        mustRecreate: (_: VolumeKey, __: PD.Values<VolumeLinesParams>, webgl?: WebGLContext) => !!webgl,
```
arose: not needed

src/mol-repr/volume/gradient.ts
```diff
@@ -0,0 +1,726 @@
...
+};
+export type VolumeGradientParams = typeof VolumeGradientParams
+export type VolumeGradientProps = PD.Values<VolumeGradientParams>
+
+
+// LINES
+export const VolumeLinesParams = {
+    ...Lines.Params,
+    ...VolumeGradientParams,
+    detail: PD.Numeric(0, { min: 0, max: 3, step: 1 }, BaseGeometry.CustomQualityParamInfo),
```
arose: ???

src/mol-repr/volume/gradient.ts
```diff
@@ -0,0 +1,726 @@
+/**
+ * Copyright (c) 2025 mol* contributors, licensed under MIT, See LICENSE file for more info.
+ *
+ * @author Ludovic Autin <autin@scripps.edu>
+ * @maintainer ChatGPT5
```
arose: nope, just you Ludo :)

arose (changes requested):

src/mol-repr/volume/gradient.ts
corredD: I added the mesh fallback for spheres and cylinders, I am actually fine with the current cylinder looks. But the tube mesh would definitely look better. I will look into it.

src/mol-repr/volume/gradient.ts
corredD: I moved the streamline in mol-math, but do you have example of cache/safe system ?

src/mol-repr/volume/gradient.ts
corredD: in the dot representation it used pt.setBoundingSphere(Volume.Isosurface.getBoundingSphere(volume, props.isoValue)); How would you do it for the streamline ?

src/mol-repr/volume/gradient.ts
arose: I think we could do this fully in the shader by a) having a global time and b) a per vertex attribute value that defines the position on a stream line. With that we can make parts of the stream line transparent based on the global time. We also need a way in the renderer to continuously render not just on changes.

Let's do this after the publications.

src/mol-repr/volume/gradient.ts
arose:
```ts
const gridDimension = volume.grid.cells.space.dimensions as Vec3;
const transform = Grid.getGridToCartesianTransform(volume.grid);
const bbox = getBoundingBox(gridDimension, transform);
```

src/mol-repr/volume/gradient.ts
arose: Something like `VolumeIsosurfaceTexture` in src\mol-repr\volume\isosurface.ts. It stores things in volume._propertyData. But the streamlines need to be a bit different as they depend on parameters and you can have multiple set of streamlines per volume at the same time.

src/mol-repr/volume/gradient.ts
corredD: I used g.setBoundingSphere(Sphere3D.fromDimensionsAndTransform(Sphere3D(), gridDimension, gridToCartn));

src/mol-repr/volume/gradient.ts
corredD: I put some caching system in the representation file, but how would you do the Volume.Streamline.Loci ? And whats the issue with the current default one ?

src/mol-theme/color/volume-value.ts
arose: Fixed it in 0e968ae59c08b967413ef755bcc301f29758e53d. You can undo your changes.

src/mol-repr/volume/gradient.ts
arose: You currently use a streamline index with a Cell Loci which makes no sense. You would need to implement a Volume.Streamline.Loci, similar to Volume.Cell.Loci but using streamline index not cell index. There are also a bunch of place where support for a new Loci needs to be added, best to use an existing Loci as an example.

giagitom:
> I tried to fix it, not sure I did it right. Did you got proper coloring with the Dot representation ?

@corredD in my implementation I am using these parameters (without specifying the offset for each color) and it is working accordingly with volume-value size theme

 ```ts 
       { 
          name : 'volume-value',
          params : {
            colorList : {
              colors : this.palette.colors.map(c => parseInt(c, 16)),
              kind : this.palette.kind
            }
          }
        }
```

src/mol-repr/volume/gradient.ts
corredD: that would be really great to have some animation accessible directly in shaders but not urgently and the current molviewstory builder works really well.
streamlines(1)

PR #1640, opened 2025-09-01
VolumeServer: Default to P1 spacegroup

dsehnal:
# Description

This fixes volume server not working with undefined spacegroup in the CCP4 header

## Actions


arose:
hmm, don't we generally want to handle such missing data downstream?

dsehnal:
I guess -- wanted a quick fix yesterday evening, but will revisit this before merging.

PR #1641, opened 2025-09-01
Trying to fix fuzzy point postprocessing when alpha is 1

giagitom:
# Description

Triyng to fix fuzzy point when alpha is 1 (still not working properly). Something changes but postprocessing isn't working properly. Outlines are strange and ssao is nearly invisible.  Not sure what is causing this. 

image

## Actions


giagitom:
Now outlines seems to work well, but strangely ssao isn't..

Image

Issue #1623, opened 2025-08-22
Unable to highlight cells in Volume Representations when hovering/clicking over canvas (when using volume data derived from https://files.wwpdb.org/pub/emdb/structures)


giagitom:
I tried with a few maps and all of them don't allow highlighting in the canvas in basically all spots (some rare groups of cells are correctly highlighted and can be focused though). Tested with isosurface and dot representation, both presents the issue. Here is an example url presenting the issue https://files.wwpdb.org/pub/emdb/structures/EMD-46521/map/emd_46521.map.gz

arose:
Yeah, that is because there are more than 2^24 groups, i.e., cells in those volumes. Maybe that can eventually be raised to 2^32 but will be some work and not sure if it would work with webgl1.

What we can probably easily do is support picking the whole instance.

arose:
For context, 2^24 is the max-safe-integer of floats and we use floats for group ids. Additionally there are place where we only use 24 bits (rgb part of a texture) for group ids.

With webgl2 we can make use of integer attributes (not implemented in engine at present) and write integer textures. 

Another option would be to split-up large volumes.

For me it is an edge use-case and not high priority.

PR #1651, opened 2025-09-06
improve picking of objects with many groups

arose:
# Description

Improve picking of objects with many groups. If there are too many groups (currently >=2^24-2) pick whole instance/object instead of picking nothing.

Partly addresses #1623.

## Actions


giagitom (approved):
Thanks @arose ,
It seems to be working as expected. When the picked ID is less than 2^24, a group is selected; otherwise, the whole instance is selected.

PR #1655, opened 2025-09-09
Adaptive parallel shader compilation

arose:
# Description

- Split shader compilation into linking and finalizing
- Start linking as early as possible and wait with finalizing to avoid blocking main thread
    - Use of `KHR_parallel_shader_compile` extension when available to check status
- Compile shaders when needed based on `Canvas3D` params and `Scene` content
- Draw `Scene` only when shaders are ready

## Actions

### DIFF
--- a/src/mol-canvas3d/canvas3d.ts
+++ b/src/mol-canvas3d/canvas3d.ts
@@ -744,9 +744,9 @@ namespace Canvas3D {
         let drawPaused = false;
         let isContextLost = false;
 
-        function draw(options?: { force?: boolean, xrFrame?: XRFrame }) {
+        function draw(options?: { force?: boolean, isSynchronous?: boolean, xrFrame?: XRFrame }) {
             if (drawPaused || isContextLost) return;
-            if (!webgl.resources.finalizePrograms(requiredShaderVariants)) {
+            if (!webgl.resources.finalizePrograms(requiredShaderVariants, options?.isSynchronous)) {
                 forceNextRender = true;
                 return;
             }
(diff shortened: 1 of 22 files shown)


src/mol-gl/webgl/render-item.ts
```diff
@@ -218,6 +218,8 @@ export function createRenderItem<T extends string>(ctx: WebGLContext, drawMode:
             if (drawCount === 0 || instanceCount === 0) return;
 
             const program = programs[variant];
+            if (!program.isReady()) return;
```
dsehnal: I don't think this will work together with the animation generation. 

I've pushed https://github.com/molstar/molstar/pull/1655/commits/bdd1805620f0e524874c79deaa534d88a530a268 with a suggestion to move this into the commit queue -- although that still has some drawbacks and would have to solve the init of the post-processing shaders separately.

Alternatively, we can update the renderable "lifecycle" to support splitting the compilation of shaders into a separate step?

src/mol-canvas3d/canvas3d.ts
```diff
@@ -715,6 +746,10 @@ namespace Canvas3D {
 
         function draw(options?: { force?: boolean, xrFrame?: XRFrame }) {
             if (drawPaused || isContextLost) return;
+            if (!webgl.resources.finalizePrograms(requiredShaderVariants)) {
```
dsehnal: If this ever returns false, it will break the animation (and possibly offscreen screenshot) rendering.

Can we pass `isSynchronous` from `tick()` and run this call with "force" parameter?

dsehnal: Added it.

dsehnal (approved):
Tested on Mac in FF, Chrome and Safari and works well.

PR #1663, opened 2025-09-17
Allow full-screen (not just full-page) mode

prihoda:
# Description

Enable true fullscreen mode using `requestFullscreen()`.

This is useful in two cases:

- When people want to enable users to see molstar in complete fullscreen mode
- When molstar is being used through an iframe, since full-page just means full-iframe in that case. When the iframe has `allow="fullscreen"`, this behavior will enable showing the viewer in fullscreen mode even if it sits inside the iframe.

## TODO

- [ ] Add a config attribute that controls this behavior and only run requestFullscreen if enabled. This is beyond my capabilities 😄  @dsehnal could you help with this if you agree with the PR?

## Actions

### DIFF
--- a/src/mol-plugin/layout.ts
+++ b/src/mol-plugin/layout.ts
@@ -115,7 +115,11 @@ export class PluginLayout extends StatefulPluginComponent<PluginLayoutStateProps
                 }
 
                 if (!hasExp) head.appendChild(this.expandedViewport);
-
+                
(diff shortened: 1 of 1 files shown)


luk27official:
I believe it would be great if one could specify this when creating the plugin instance, e.g. in `PluginSpec.config` :)

prihoda:
Lukas continues here: https://github.com/molstar/molstar/pull/1666

PR #1666, opened 2025-09-30
ExpandToFullscreen option in PluginConfig

luk27official:
# Description

> OK, feel free to add "ExpandToFullsceen" option to the plugin config list

As discussed in #1663, I have implemented the `PluginConfig.General.ExpandToFullscreen` option to enable fullscreen mode (but I don't have push rights for David's repository).

A new `fullscreenchange` handler was added as well - the user might want to get out of the fullscreen mode using Escape and not by using the expand button.

> referencing it in the layout manager, supporting it in the Viewer, the URL params, etc.

I have tried changing the configuration for the `docking-viewer` app which is working, referencing the value in the `Viewer` class works as well. Not sure about the URL params, though.

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ Note that since we don't clearly distinguish between a public and private interf
 
 ## [Unreleased]
 
+- Add `PluginConfig.General.ExpandToFullscreen` to enable fullscreen instead of just expanding the viewer
+
 ## [v5.0.0] - 2025-09-28
 - [Breaking] Renamed some color schemes ('inferno' -> 'inferno-no-black', 'magma' -> 'magma-no-black', 'turbo' -> 'turbo-no-black', 'rainbow' -> 'simple-rainbow')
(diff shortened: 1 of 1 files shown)


src/mol-plugin/layout.ts
```diff
@@ -199,7 +209,22 @@ export class PluginLayout extends StatefulPluginComponent<PluginLayoutStateProps
...
+        if (expandToFullscreen) {
+            const fullscreenChangeHandler = () => {
+                // In case the user exits fullscreen mode by pressing ESC, treat it as an expand toggle
+                if (!document.fullscreenElement) {
+                    this.updateProps({ isExpanded: false });
+                }
+            };
+
+            document.addEventListener('fullscreenchange', fullscreenChangeHandler);
+            document.addEventListener('webkitfullscreenchange', fullscreenChangeHandler);
```
dsehnal, line 226: This could cause a memory leak, so will need to add a `dispose` method too which unregisters the listeners during `plugin.dispose`.

PR #1683, opened 2025-10-25
Fullscreen mode improvements

dsehnal:
# Description

Follow-up to #1666

- [x] Update fullscreen toggle logic
- [x] Add UI support to `viewport.tsx`
- [x] Add `show-toggle-fullscreen` Viewer URL option

## Actions

### DIFF
--- a/src/mol-plugin/layout.ts
+++ b/src/mol-plugin/layout.ts
@@ -10,7 +10,6 @@ import { ParamDefinition as PD } from '../mol-util/param-definition';
 import { StatefulPluginComponent } from '../mol-plugin-state/component';
 import { PluginCommands } from './commands';
 import { PluginContext } from './context';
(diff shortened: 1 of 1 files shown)


src/mol-plugin/layout.ts
```diff
@@ -72,6 +71,13 @@ export class PluginLayout extends StatefulPluginComponent<PluginLayoutStateProps
         this.updateState(state);
         if (this.root && typeof state.isExpanded === 'boolean' && state.isExpanded !== prevExpanded) this.handleExpand();
 
+        if (this.state.expandToFullscreen) {
+            const body = document.getElementsByTagName('body')[0];
```
arose, line 75: Will this ever be called when run in a none Browser environment?

dsehnal: I don't think so. The same line is another function in the file and it didn't cause problems.

PR #1690, opened 2025-11-02
fix: validation and default params for primitives_from_uri

datichy:
# Description

Added validation of the primitives node which gets downloaded by primitives_from_uri load action.
Added missing default param values to the downloaded primitives node.

Without this fix valid mesh primitives don't always load.
For example, if `show_triangles` and `show_wireframe` aren't present in the downloaded primitives node, the isApplicable check of mesh builder fails and nothing gets built.

## Actions


dsehnal:
Thanks. I am leaving for a week tomorrow and will have a look after.

PR #1704, opened 2025-11-11
MVS: Allow canvas background interpolation

midlik:
# Description

This is to allow canvas background_color interpolation in MVS animations.

@dsehnal , could you please double-check these changes? I'm not sure if this is the right approach, and I remember you saying something about performance.

Also it's seem suspicious to me that there is the `setAnimationSnapshot` method, but it doesn't get called. (you added in [MVS: Animations #1606](https://github.com/molstar/molstar/pull/1606))

Testing file:
[background-interpolation.mvsj](https://github.com/user-attachments/files/23477350/background-interpolation.mvsj.remove_extension.txt)

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,12 +6,13 @@ Note that since we don't clearly distinguish between a public and private interf
 ## [Unreleased]
 
 - Fix ugly camera clipping in snapshot transitions
-- MolViewSpec
-  - Fix `appendSnapshots` when loading MVSX
 - Add viewport button to toggle illumination mode
 - Fix bounding sphere computation for 3D text
 - MolViewSpec
+  - Fix `appendSnapshots` when loading MVSX
   - Fix all-selector color not applying on substructure
+  - Fix primitives in root not being transformed with reference structure
+  - Color themes do not prefer smoothing (improves performance in animations)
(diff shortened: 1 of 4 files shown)


src/mol-plugin-state/animation/built-in/state-snapshots.ts
```diff
@@ -4,23 +4,30 @@
...
-        if (entry.canvas3d?.props?.trackball) {
-            plugin.canvas3d?.setProps({
-                trackball: entry.canvas3d?.props?.trackball
-            });
-        }
-
     }
 
+    if (entry.canvas3d?.props) {
+        const settings = PD.normalizeParams(Canvas3DParams, entry.canvas3d.props, 'children');
```
dsehnal, line 22: Could you benchmark the if condition code that normalizes and sets the props? If it takes say <1us I think it might be fine to include... otherwise we should optimize. (And even 1ms might be too long for essentially just changing background color).

midlik: Based on console.time,
the `if (entry.canvas3d?.props) {...}` part takes 0.15ms on average,
while the whole method takes 1.96ms.

PR #1707, opened 2025-11-14
Add Chinese language support to the text input box.

rxht:
# Description

## Actions

- [ ] Add Chinese language support to the text input box.


arose (approved):
that's great, thanks

PR #1712, opened 2025-11-24
Import tweaks

arose:
# Description

A few changes to reduce dependencies between mol-* components. Helps to create smaller files with bundlers.

## Actions

### DIFF
--- a/src/mol-util/string.ts
+++ b/src/mol-util/string.ts
@@ -5,9 +5,6 @@
  * @author David Sehnal <david.sehnal@gmail.com>
  */
 
-import { StringLike } from '../mol-io/common/string-like';
-
-
 const reLine = /^/mg;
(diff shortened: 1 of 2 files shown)


src/mol-util/string.ts
```diff
@@ -57,7 +54,7 @@ export function stringToWords(str: string) {
     return capitalize(splitCamelCase(splitSnakeCase(splitKebabCase(str))));
 }
 
-export function substringStartsWith(str: StringLike, start: number, end: number, target: string) {
+export function substringStartsWith(str: string, start: number, end: number, target: string) {
```
dsehnal, line 57: Use `import type { StringLike } from ....`. Could maybe set up a linter for these cases too.

arose: unfortunately using type only imports did not work for amd builds with tsc 🤷

PR #1722, opened 2025-12-17
Add Residue Charge color scheme

luk27official:
# Description

This PR adds a new "Residue Charge" color palette for residues. I have added it as a new palette for the "Residue Name" scheme. Only certain residues are colored as most of the residues are not charged. Histidine is colored in light blue as it might not always be charged.

References:
- https://www.rbvi.ucsf.edu/chimerax/data/electrostatics-apr2021/crm1.html
- https://pubmed.ncbi.nlm.nih.gov/26528483/

Feel free to change the codes for the colors if needed.

Including a screenshot for [6tvo](https://www.rcsb.org/structure/6tvo)

6TVO_A_CENTERED PDB

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Note that since we don't clearly distinguish between a public and private interf
 - Fix webgl1 shader syntax
 - Fix program not compiled for sync picking
 - Fix missing `gl.flush` for async picking (needed for Safari)
-- Add Residue Charge color scheme
+- Add Residue Charge color scheme (#1722)
 
 ## [v5.4.2] - 2025-12-07
 - Fix postprocessing issues with SSAO and outlines for large structures (#1387)


src/mol-theme/color/residue-name.ts
```diff
@@ -63,6 +64,56 @@ export const ResidueNameColors = ColorMap({
 });
 export type ResidueNameColors = typeof ResidueNameColors
 
+// Colors for charged residues
+export const ChargedResidueColors = ColorMap({
+    // standard amino acids (charged)
+    'ARG': 0x0000FF,
```
dsehnal: Can you add links to the references to the code as well?

src/mol-theme/color/residue-name.ts
```diff
@@ -71,7 +122,8 @@ export const ResidueNameColorThemeParams = {
     lightness: PD.Numeric(1, { min: -6, max: 6, step: 0.1 }),
     colors: PD.MappedStatic('default', {
         'default': PD.EmptyGroup(),
-        'custom': PD.Group(getColorMapParams(ResidueNameColors))
+        'custom': PD.Group(getColorMapParams(ResidueNameColors)),
+        'residue-charge': PD.Group(getColorMapParams(ChargedResidueColors))
```
dsehnal: This feels quite misplaced to me -- I would prefer this was a new color theme altogether. Basically a copy-paste of this theme, but rename it to `residue-charge` and perhaps allow alternative color schemes.

What do you think?

src/mol-theme/color/residue-name.ts
```diff
@@ -63,6 +64,56 @@ export const ResidueNameColors = ColorMap({
 });
 export type ResidueNameColors = typeof ResidueNameColors
 
+// Colors for charged residues
+export const ChargedResidueColors = ColorMap({
```
prihoda: Looks like the viewer is changing the hue a bit, so I would use slightly different colors to get a red-blue scale.

```
arg, lys - blue: #015ac8
his - light blue: #2996ff
glu, asp - red: #ff0033
```

image

src/mol-theme/color/residue-name.ts
prihoda: I agree, ideally putting it directly below/above Hydrophobicity

src/mol-theme/color/residue-name.ts
luk27official: I have changed the default `lightness` to 0 and kept the original colors, so now it looks like this

1TQN

the user might tweak the lightness to their needs later so I hope this is good enough

src/mol-theme/color/residue-name.ts
luk27official: Yes, I didn't want to duplicate the code too much but it makes sense

> I agree, ideally putting it directly below/above Hydrophobicity

Now the options are listed alphabetically so I will keep it that way

luk27official:
I have addressed the comments :)

- changed colors
- moved to a separate file
- added references

1TQN

src/mol-theme/color/residue-charge.ts
```diff
@@ -0,0 +1,152 @@
...
+const DefaultResidueChargeColor = Color(0xFF00FF);
+const Description = 'Assigns a color to every residue based on its charge state.';
+
+export const ResidueChargeColorThemeParams = {
+    saturation: PD.Numeric(0, { min: -6, max: 6, step: 0.1 }),
+    lightness: PD.Numeric(0, { min: -6, max: 6, step: 0.1 }),
+    colors: PD.MappedStatic('default', {
+        'default': PD.EmptyGroup(),
+        'custom': PD.Group(getColorMapParams(ChargedResidueColors)),
+    })
```
dsehnal: Instead of this, can you do

`method: PD.MappedStatic('by-name', {
  'by-name': <current params including lightness and saturation>
}, { isFlat: true })`

to allow for potential extensibility later?

luk27official: done

dsehnal (approved):

PR #1725, opened 2025-12-19
Viewer improvements

dsehnal:
# Description

- Viewer app
  - Move viewer extensions, options, and presets to a separate file
  - Add `molstar.lib` export providing access to a wide range of functionality previously not available from the compiled bundle
  - Add `Viewer.subscribe` method that keeps track of subscribed plugin events and disposes them together with the parent viewer
  - Add `Viewer.structureInteraction` that makes it easy to highlight/select elements on the loaded structure
  - Add `viewportBackgroundColor` and `viewportFocusBehavior` options
  - Add `mvs.html` example to showcase the new functionality combined with MolViewSpec
  - Add dark and blue color theme support (import `theme/dark.css` or `theme/blue.css` instead of the default `molstar.css`)
- Add dropdown indicator for mapped parameter definitions and adjust "more options" icon
- MolViewSpec extension
  - Add `tryGetPrimitivesFromLoci` that makes it easier to access primitive element data from hover/click interactions

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ Note that since we don't clearly distinguish between a public and private interf
   - Add `viewportBackgroundColor` and `viewportFocusBehavior` options
   - Add `mvs.html` example to showcase the new functionality combined with MolViewSpec
   - Add dark and blue color theme support (import `theme/dark.css` or `theme/blue.css` instead of the default `molstar.css`)
+- Fix `flipSided` for meshes
 
 ## [v5.4.2] - 2025-12-07
 - Fix postprocessing issues with SSAO and outlines for large structures (#1387)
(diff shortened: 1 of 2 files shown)


src/apps/viewer/app.ts
```diff
@@ -567,7 +513,48 @@ export class Viewer {
...
     }
 
+    /**
+     * Triggers structure element selection or highlighting based on the provided
+     * MolScript expression or StructureElement schema.
+     *
+     * If neither `expression` nor `elements` are provided, all selections/highlights
+     * will be cleared based on the specified `action`.
+     */
+    structureInteraction({ expression, elements, action, applyGranularity = false }: {
```
arose:
```suggestion
    structureInteractivity({ expression, elements, action, applyGranularity = false }: {
```

less confusion with non-bonded interactions

arose: it would be great to have some way to apply this to specific structures

dsehnal: Added a callback to filter the structures.

PR #1723, opened 2025-12-17
Add TM-align structure alignment algorithm

ddelalamo-takeda:
Implement TM-align for structure-based protein alignment, providing an alternative to sequence-based superposition methods. TM-align finds optimal structural alignments regardless of sequence similarity using TM-score.

New features:
- Core TM-align algorithm (src/mol-math/linear-algebra/3d/tm-align.ts)
  - TM-score calculation with length-independent normalization
  - Dynamic programming for optimal alignment
  - Gapless threading and fragment-based initialization
  - Multiple refinement passes for accuracy (~97.7% of US-align reference)

- High-level wrapper (src/mol-model/structure/structure/util/tm-align.ts)
  - tmAlign() function accepting StructureElement.Loci inputs
  - Returns transformation matrix, TM-scores, RMSD, and alignment

- BasicWrapper API (src/examples/basic-wrapper/)
  - tmAlign(pdbId1, chain1, pdbId2, chain2, color1?, color2?)
  - sequenceAlign() - sequence-based superposition
  - loadStructures() - load without alignment
  - Example HTML pages demonstrating usage

- UI integration (src/mol-plugin-ui/structure/superposition.tsx)
  - TM-align option in superposition panel

References:
Zhang Y, Skolnick J. Nucl Acids Res 33, 2302-9 (2005)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1726, opened 2025-12-20
Ribocode extension

agdturner:
# Description
Some generally useful geometry for aligning molecules

## Actions


src/extensions/ribocode/utils/qcprot.ts
```diff
@@ -0,0 +1,396 @@
...
+ * PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR
+ * CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+ * EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+ * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+ * PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+ * LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+export class QCProt {
```
dsehnal, line 46: As @arose pointed out, there is already quaternion-based RMSD minimization code in https://github.com/molstar/molstar/blob/master/src/mol-math/linear-algebra/3d/minimize-rmsd.ts, although the implementation here does seem to better handle edge-cases.

I would prefer we improve the existing code instead of including a duplicate.

src/extensions/ribocode/structure.ts
```diff
@@ -0,0 +1,176 @@
...
+
+/**
+ * Gets alignment data for a molecule including the type and location
+ * of all atoms in the structure.
+ * @param plugin The PluginUIContext instance.
+ * @param trajectory The trajectory StateObjectRef.
+ * @returns A promise that resolves to an object containing arrays of
+ * x, y, z coordinates and atom types.
+ */
+export async function getAlignmentData(plugin: PluginUIContext, trajectory: any) {
```
dsehnal: This is a very hacky way of getting the position elements -- generally I would at least get this from the `Model` which has types data.

Is the purpose of the `loadMoleculeFileToViewer` only to get the positions? The way this thing is coded seems quite specific to the application you want to use this in.

src/extensions/ribocode/utils/geometry.ts
```diff
@@ -0,0 +1,327 @@
...
+ * @param newX X coordinates to align
+ * @param newY Y coordinates to align
+ * @param newZ Z coordinates to align
+ * @param type Array of atom types of data to align with
+ * @param x X coordinates to align with
+ * @param y Y coordinates to align with
+ * @param z Z coordinates to align with
+ * @return Aligned coordinates as {alignedX, alignedY, alignedZ}
+ */
+export function alignDataset(
```
dsehnal: @ddelalamo-takeda recently added support for `tm-align` #1723 which follows the existing patterns for superposing structures (notably the input of `Structure.Loci` instead of the `selectedAtomTypes` filtering you have inside the function). Can you follow this pattern here as well?

Due to the usage of `big.js`, I would still prefer this being a separate extension instead of part of the core library.

src/extensions/ribocode/utils/geometry.ts
```diff
@@ -0,0 +1,327 @@
...
+    centroid[2] /= coords.length;
+    return centroid;
+}
+
+/**
+ * Compute the centroid of a set of coordinates.
+ * @param coords Array of BigVec3 coordinates.
+ * @returns Centroid as Vec3
+ */
+export function computeBigCentroid(coords: Vec3[]): Vec3 {
```
dsehnal: Out of curiosity, how big structures to force the use of an arbitrary precision library?

agdturner: Ribosome datasets can have hundreds of thousands of atoms.

dsehnal: I could be wrong, but the precision of the atom positions themselves isn't very accurate and 15 significant digits still seems quite good.

What is generally the error between using the 64bit floats vs the arbitrary precision computation? And if the different centroid is used in the RMSD calculation, what is the difference then?

terq
dsehnal, line 1: Extra file that should not be here?

package.json
```diff
@@ -165,6 +165,7 @@
     "@types/node-fetch": "^2.6.13",
     "@types/swagger-ui-dist": "3.30.6",
     "argparse": "^2.0.1",
+    "big.js": "^7.0.1",
```
dsehnal, line 169: Generally, I prefer not to include 3rd party dependencies if it could be avoided.

BigJS is MIT and just a [single file](https://github.com/MikeMcl/big.js/blob/main/big.mjs), which means we could actually just copy-paste it into `mol-util/big.ts` (and include proper typescript definition). I could help with that if you want.

src/extensions/ribocode/structure.ts
agdturner: Ah, but you know better what you are doing with Mol* and I am trying to learn how to use it @dsehnal!

The molecule is centralised so that the coordinate origin is at the centre of the molecule (the position of each atom is translated by the average of all the atom positions). If alignment data are provided, then the atoms are also rotated. If the user requests alignment data to be returned, then the centralised atom locations are returned.

dsehnal: Let's step back a little. The functionality you want is:

- Load two or more ribosomes
- Compute their alignment and apply it to the structures
- Output the translations and matrices to the user

Am I missing something?

agdturner: No, I think that is the gist of it @dsehnal.

terq
agdturner: Sorry, not sure where that is from. I did not include it intentionally.

package.json
agdturner: I read that type definitions are included in big.js 7.0.1, but I'm not sure how...

dsehnal: Part of https://github.com/DefinitelyTyped/DefinitelyTyped

src/extensions/ribocode/structure.ts
dsehnal: Ok, then you want to structure your library differently. I would suggest

- Load all the structures using built-in methods (you don't need the trajectory.ts file you added)
- Find the superpositions
  - I am having some trouble understanding what `alignmentData.type` means in relation to `selectedAtomTypes` in `alignDataset`, but basically the method should get a selection you want to align and output a transformation matrix
- Apply `TransformStructureConformation` based on the transformations to update the visuals
- To the user, you can either output the transformation matrix or the transformed positions
  - But this would be part of  your particular instance of Mol*, not necessarily the extension.

I would again encourage you to have a look at https://github.com/molstar/molstar/pull/1723 which achives a lot of what you want (= adding an alternative superposition method)

package.json
agdturner: I understand that the Type definitions for Big.js can be added with:
npm install --save-dev @types/big.js

dsehnal: Yes, but I don't want to include a dependency on another 3rd party library if it's somewhat easily avoidable.

PR #1737, opened 2025-12-30
stable random numbers (PCG)

arose:
# Description

add stable random numbers (PCG) and use for

- ME grayscale colors
- dot offset
- SSAO hemisphere vectors

they are always the same so there are no (subtle) changes between sessions

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Note that since we don't clearly distinguish between a public and private interf
 - Add stable random number generator (PCG)
     - ME grayscale colors; dot offset; SSAO hemisphere vectors
     - Use blue noise for SSAO hemisphere vectors
+- Fix SSAO darkening when sampling background/offscreen pixels
 
 ## [v5.5.0] - 2025-12-22
 - Viewer app
--- a/src/mol-gl/shader/ssao.frag.ts
+++ b/src/mol-gl/shader/ssao.frag.ts
@@ -234,9 +234,11 @@ void main(void) {
                     if (uTransparencyFlag == 1) {
                 #endif
                     float sampleDepth = getMappedDepth(offset.xy, selfCoords);
-                    float sampleViewZ = screenSpaceToViewSpace(vec3(offset.xy, sampleDepth), uInvProjection).z;
+                    if (!isBackground(sampleDepth)) {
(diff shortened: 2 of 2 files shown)


src/mol-canvas3d/passes/ssao.ts
```diff
@@ -681,14 +682,15 @@ function getBlurKernel(kernelSize: number): number[] {
...
-    v[0] = Math.random() * 2.0 - 1.0;
-    v[1] = Math.random() * 2.0 - 1.0;
-    v[2] = Math.random();
+    v[0] = pcg.float() * 2.0 - 1.0;
+    v[1] = pcg.float() * 2.0 - 1.0;
+    v[2] = pcg.float();
     Vec3.normalize(v, v);
-    Vec3.scale(v, v, Math.random());
+    Vec3.scale(v, v, pcg.float());
     RandomHemisphereVector.push(v);
```
dsehnal: Based on what @giagitom points out it seems like this sphere sampling isn't very uniform.

It seems like the first 3*256 numbers with this particular seed either (or a combination of):
- don't produce a very good distribution of vectors 
- the way you are generating a random vector on a sphere is biased. I think you would get a better distribution if you generated two angles in spherical coordinates and converted that to a vector instead. (or see say https://math.stackexchange.com/questions/1585975/how-to-generate-random-points-on-a-sphere; PCG produces uniform distribution, someone mentions gaussian instead needed for this). Basically turning uniform distribution into sphere points requires more transformation than just a normalization.

arose: - yeah it is not a good seed
- it is a little biased but that is not much of an issue here (need to use blue noise, see next comment)

arose:
Had a look at the distribution and it is apparent that they are too clumped together, basically white noise (both the old way and the first iteration of this PR). What we really want is blue noise distributed vectors - still using PCG for reproducible results.

image
First 32 vectors

image
All 256 vectors

In general the SSAO is slightly lighter than with the old method, due to the unbiased sphere sampling (less samples in the "corners"). We could actually set the default SSAO bias to 1 now.

Sidenote: LLMs are great to create little throwaway visualization apps ⬆️.

arose:
> @arose, an unrelated question.. I noticed something when using huge ssao radius (> 12). I would expect to not see any occlusion at all but instead I am seeing it in different places where it shouldn't. Any idea why this is happening?

added a fix to skip samples from background/offscreen pixels

PR #1740, opened 2026-01-02
fix ssao darkening

arose:
# Description

- instead of clamping offscreen pixels, skip samples

## Actions

### DIFF
--- a/src/mol-gl/shader/ssao.frag.ts
+++ b/src/mol-gl/shader/ssao.frag.ts
@@ -235,20 +235,19 @@ void main(void) {
                     if (uTransparencyFlag == 1) {
                 #endif
                         float sampleDepth = getMappedDepth(offset.xy, selfCoords);
-                        if (isBackground(sampleDepth)) continue;
-
-                        float sampleViewZ = screenSpaceToViewSpace(vec3(offset.xy, sampleDepth), uInvProjection).z;
-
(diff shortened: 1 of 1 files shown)


giagitom (commented):
Nice! This is much better than before.
@arose, there is an issue due to the refactoring that breaks transparent occlusion. 

I think it is due to the  `isBackground` check together with the `continue` statements that prevents the transparency evaluations in the loops when the opaque sample is a background pixel. 

I am also thinking about a possible solution for this effect that is easily visible on multilayered ssao on viewport boundaries:

image

To avoid this i think we should decrease NSamples by 1 each time the sample is outside the boundaries, in order to normalize the occlusion effect at the end

`occlusion = 1.0 - (uBias * occlusion / float(dNSamples));` 

This would require to declare something like this outside the loops `float NOcc = float(dNSamples)`  and decrease `NOcc` when sample is outside the boundaries. 

Then (at the end of each layer loop in the case of multilayered approach) normalize the occlusion with the number of samples that was effectively evaluated.

arose:
> I think it is due to the isBackground check together with the continue statements that prevents the transparency evaluations in the loops when the opaque sample is a background pixel.

should be fixed now

> I am also thinking about a possible solution for this effect that is easily visible on multilayered ssao on viewport boundaries:

I think what you describe in the image is almost completely from he screen-space shadows

PR #1743, opened 2026-01-04
fix: add alpha channel handling

zachcp:
# Description

In recent work on mol-view-stories, I notice that new versions of the Monaco editor provide a color picker. This color picker, however, uses hexcodes with an alpha channel (8-character: color + alpha ) that is incompatible with molstar's current color handling (6-character).  This PR adds functionality to handle 8-character hex codes. 

In the image below you can see molstar/molviewspec valid code `#cc3399` but the color picker  will append `ff`, e.g. `#cc3399ff` which will lead to an unrecognized color.

image

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file, following t
 Note that since we don't clearly distinguish between a public and private interfaces there will be changes in non-major versions that are potentially breaking. If we make breaking changes to less used interfaces we will highlight it in here.
 
 ## [Unreleased]
+- handle 8-character hex codes to allow in-browser color-pickers
 - Only show "already registered transformer" warnings in non-production builds
 - Fix `label_seq_id` assignment in PDB parser to use 1-based linear indexing (#1730) if:
   - when insertion codes are present


arose (changes requested):
Could you please use 4 space indents. It is very hard to see what actually changed. 

Thanks for adding tests!

src/mol-util/color/utils.ts
```diff
@@ -1,22 +1,31 @@
...
  * @author David Sehnal <david.sehnal@gmail.com>
+ * @author Zachary Charlop-Powers <zach.charlop.powers@gmail.com>
  */
 
 import { Color, ColorListEntry } from './color';
 import { ColorNames } from './names';
 
-const hexColorRegex = /^#([0-9A-F]{3}){1,2}$/i;
-const rgbColorRegex = /^rgb\(\s*(\d{1,3})\s*,\s*(\d{1,3})\s*,\s*(\d{1,3})\s*\)$/i;
+const hexColorRegex = /^#([0-9A-F]{3}){1,2}([0-9A-F]{2})?$/i;
```
dsehnal, line 11: This matches `/^#([0-9A-F]{3}){1,2}([0-9A-F]{2})?$/i.test("#12345") => true`. Is that expected?

The previous version only matches 3 or 6 digits.

src/mol-util/color/utils.ts
```diff
@@ -38,11 +47,50 @@ export function decodeColor(colorString: string | undefined | null): Color | und
     return undefined;
 }
 
+/**
+ * Decode color string and extract alpha channel if present.
+ * Returns both the Color and alpha value (0-1 range).
+ * Alpha defaults to 1.0 if not specified in the color string.
+ */
+export function decodeColorWithAlpha(
```
dsehnal, line 55: We currently don't support alpha as part of colors anywhere. Maybe we can not include this function yet as it has no use case within mol* currently?

src/mol-util/color/utils.ts
```diff
@@ -1,22 +1,31 @@
...
+export function decodeColor(
+    colorString: string | undefined | null,
+): Color | undefined {
     if (colorString === undefined || colorString === null) return undefined;
     let result: Color | undefined;
     if (hexColorRegex.test(colorString)) {
         if (colorString.length === 4) {
             // convert short form to full form (#f0f -> #ff00ff)
-            colorString = `#${colorString[1]}${colorString[1]}${colorString[2]}${colorString[2]}${colorString[3]}${colorString[3]}`;
+            colorString = `#${colorString[1]}${colorString[1]}${colorString[2]}${
```
dsehnal, line 23: What linter are you using?

it seems like it's something with lint-on-safe. Maybe we can include rules that keep it consistent with what's common across the codebase? It seems like you have things like 80 character column present.

src/mol-util/color/utils.ts
zachcp: Agree. Except  if you are passed a full hex string it will fail.  These fns just strip the alpha out.

src/mol-util/color/utils.ts
zachcp: yeah sorry.  I'm hitting some default editor lints that run whenever  I save.

zachcp: ( zed editor.  has a bunch of auto edit features enabled by default )

PR #1746, opened 2026-01-10
20261110 hex fix clean

zachcp:
# Description

Note: Taking over from #1743. The motivation was due to the standard return from the Monaco Code Editor which returns the 8-character hex with alpha included. This fix simply extends the Regex and ignores the alpha channel. 

- Feat: Updated the hex regex which used to only allow 3 or 6 character hex codes. This PR allows us to accept 4 and 8-character regex which include an alpha channel. In all cases we return the 6-character color:
   - 3: `#f0a` -> `#ff00aa`
   - 4:  `#f0ab` -> `#ff00aa`
   - 6:  `#ff00aa` -> `#ff00aa`
   - 8:  `#ff00aacc` -> `#ff00aa`

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file, following t
 Note that since we don't clearly distinguish between a public and private interfaces there will be changes in non-major versions that are potentially breaking. If we make breaking changes to less used interfaces we will highlight it in here.
 
 ## [Unreleased]
+- handle Hex codes that are submitted with alpha channels by ignoring the alpha channel (#1746)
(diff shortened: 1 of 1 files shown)


src/mol-util/_spec/color.spec.ts
```diff
@@ -0,0 +1,60 @@
+
+/**
+ * Copyright (c) 2023 mol* contributors, licensed under MIT, See LICENSE file for more info.
```
dsehnal:
```suggestion
 * Copyright (c) 2026 mol* contributors, licensed under MIT, See LICENSE file for more info.
```

PR #1750, opened 2026-01-16
Fix outline postprocessing artifacts

sbittrich:
# Description
- add a lazy curvature veto to suppress outline false positives on flat/near-planar surfaces
- simpler/cheaper alternatives exist but altered outline visuals more drastically

## Actions


sbittrich:
The existing loop could store the needed depths/viewZs for L, R, U, D. That way, the veto doesn't double fetch textures if it runs. Might be worth it?

arose:
If someone wants to look more closely at performance, feel free to open another PR.

PR #1752, opened 2026-01-16
Non-covalent interactions: Prevent self-occlusion on quaternary amines

sbittrich:
# Description
- [paper describing interactions in 2reg](https://www.jbc.org/article/S0021-9258(20)60468-9/fulltext)
  - TYR-119 should bind via pi-cation interaction
  - ASP-45 should bind via ionic interaction
- Mol* detects the features correctly but rejects this particular interaction due to the line-of-sight check (shows up when reducing the dist factor)
- this fix marks all atoms in quaternary amine group as positive without shifting the feature centroid
- a dedicated array with stuff to ignore during LoS would be cleaner but this seems safe-ish to me

## Actions

| Current behavior | New behavior |
|---|---|
| 2REG-old | 2REG-new |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,8 +28,10 @@ Note that since we don't clearly distinguish between a public and private interf
     - ME grayscale colors; dot offset; SSAO hemisphere vectors
     - Use blue noise for SSAO hemisphere vectors
 - Fix SSAO darkening when sampling background/offscreen pixels
+- Adding structure wireframe visuals on molecular and gaussian surfaces
 - Fix caching of `__srcIndexArray__`
 - Prevent self-occlusion on quaternary amine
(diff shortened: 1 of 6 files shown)


sbittrich:
Also, while we're here: Isn't `eachBondedAtom()` sketchy in this context?
```ts
eachBondedAtom(structure, unit, j, (_, k) => {
    if (typeSymbol(unit, k) === Elements.N) {
        addedElements.add(k);
        builder.pushMember(x[elements[k]], y[elements[k]], z[elements[k]], k);
    }
});
```

Should that be `eachIntraBondedAtom` to ensure `k` and the coords are specific to the original unit/feature?

arose (approved):
looks good, thanks

arose:
> Should that be eachIntraBondedAtom to ensure k and the coords are specific to the original unit/feature?

yes, changed it, thanks

PR #1754, opened 2026-01-18
fix boundingSphere reuse in structure visuals

### LINKED ISSUE #1748: Poor performance when playing large XTC trajectory (~200k atoms) on official Mol* web viewer
### Describe the problem
Hi Mol* developers, thank you for your great work!

I'm using the official Mol* web viewer to visualize a membrane-protein trajectory (XTC) with ~200k atoms. The visualization quality is great, but I experience very poor performance during trajectory playback.

When the trajectory is playing, the FPS is around ~1 fps, and interaction (drag/rotate) becomes noticeably laggy. I also observe intermittent high usage on a single CPU core during playback.

### Expected behavior
Trajectory playback should be reasonably smooth (or at least not stall) and interaction should remain responsive.

### Actual behavior
- Playback FPS: ~1 fps
- While playing, drag/rotate interaction is also laggy
- CPU: intermittent single-core high utilization during playback

### Steps to reproduce
1. Open the official Mol* web viewer
2. Load structure (membrane protein system, ~200k atoms)
3. Load XTC trajectory 
4. Press play
5. Observe ~1 fps playback and laggy interaction

### Data details
- System: membrane protein + membrane + solvent
- Atom count: ~200,000
- Trajectory format: XTC
- Frames: `100`
- Playback settings: default

### Environment
- Viewer: official Mol* web viewer
- OS: Ubuntu 25.04
- Browser: `Edge 143.0.3650.96`
- GPU: `Nvidia RTX 2080 Super`
- Driver: `580.95.05`

### What I tried
- Switched to line representation for performance (no improvement)
- The main issue is during playback; when try to interact when trajectory paused, the viewer is smooth.

### Attachments / profiling
I can provide:
- A **Chrome/Edge Performance trace** recorded during playback
- A **reproducer dataset** (structure + trajectory)

[Google Drive](https://drive.google.com/drive/folders/1OH_DFxcvaJIm7irxWJ0Vep44X5OC3ynz?usp=sharing)

arose:
# Description

fix boundingSphere reuse in structure visuals
- was triggering extra calculation
- add hasBoundingSphere method to geometries

partially fixes #1748

## Actions

### DIFF
--- a/src/mol-geo/geometry/points/points.ts
+++ b/src/mol-geo/geometry/points/points.ts
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2018-2024 mol* contributors, licensed under MIT, See LICENSE file for more info.
+ * Copyright (c) 2018-2026 mol* contributors, licensed under MIT, See LICENSE file for more info.
  *
  * @author Alexander Rose <alexander.rose@weirdbyte.de>
  */
@@ -46,6 +46,7 @@ export interface Points {
     readonly groupMapping: GroupMapping
 
     setBoundingSphere(boundingSphere: Sphere3D): void
+    hasBoundingSphere(): boolean
 }
 
 export namespace Points {
@@ -99,6 +100,9 @@ export namespace Points {
             setBoundingSphere(sphere: Sphere3D) {
                 Sphere3D.copy(boundingSphere, sphere);
                 currentHash = hashCode(points);
+            },
+            hasBoundingSphere() {
+                return currentHash === hashCode(points);
             }
         };
         return points;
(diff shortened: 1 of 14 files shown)


src/mol-geo/geometry/points/points.ts
```diff
@@ -99,6 +100,9 @@ export namespace Points {
             setBoundingSphere(sphere: Sphere3D) {
                 Sphere3D.copy(boundingSphere, sphere);
                 currentHash = hashCode(points);
+            },
+            hasBoundingSphere() {
+                return currentHash === hashCode(points);
```
dsehnal, line 105: Why the hashcode wrapper here?
Seems like equality with extra steps given it's defined as

```ts
function hashCode(points: Points) {
    return hashFnv32a([
        points.pointCount, points.centerBuffer.ref.version, points.groupBuffer.ref.version,
    ]);
}
```

Maybe define an `equals` function here (and in the other places affected by this) that directly checks for the property equality?

arose: we would still want the hashCode stored and this to check if a boundingSphere still reflects the contents of the buffers

PR #1770, opened 2026-02-07
Line strips

arose:
# Description

- Support memory efficient line-strips in Lines geometry,
    - Add `StripLinesBuilder`
- Add `computeFrenetFrames` helper
- Geometry export
    - Fix vertex-based coloring for non-mesh geometries
    - Support line-strips

By (mis)using the GPU vertex cache (or just de-duplication of vertex shader invocations?) we can even get smoothly connected screen-space lines.

molstar-image (30)
line strip

molstar-image (31)
regular line segments

### Sidenote

the geometry exporter is quite complex and hard to follow, should refactor at some point.

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,13 @@ Note that since we don't clearly distinguish between a public and private interf
 - [Breaking] Add PQR file format support (#157)
     - Replace `isPdbqt` with `variant` param in `TrajectoryFromPDB`
 - Add `CustomVolumeProperty` (like for models and structures)
-- Fix missing `usePalette` support in `MeshExporter`
+- Geometry export
+    - Fix missing `usePalette` support
+    - Fix vertex-based coloring for non-mesh geometries
+    - Support line-strips
+- Support memory efficient line-strips in Lines geometry,
+    - Add `StripLinesBuilder`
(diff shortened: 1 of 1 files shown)


src/extensions/geo-export/mesh-exporter.ts
```diff
@@ -395,10 +423,133 @@ export abstract class MeshExporter<D extends RenderObjectExportData> implements
...
+        const aGroup = values.aGroup.ref.value;
+        const stripCount = values.stripCount.ref.value;
+
+        const stripOffsets = values.stripOffsets.ref.value;
+        const aMapping = values.aMapping.ref.value;
+
+        if (this.options.linesAsTriangles) {
+            const instanceCount = values.instanceCount.ref.value;
+            const meshes: Mesh[] = [];
+            const radialSegments = 6;
```
dsehnal, line 438: Should this be a parameter somewhere?

And can it be a function of the current curvature? And the "segment count" just being a "scaling factor"?

arose: At some point we should make it configurable. It is just a potential fallback for exporters that dont support line primitives. 

The Segment length is given by the input geometry which may or may not consider curvature. Don't want to reinterpret it here.

PR #1771, opened 2026-02-09
Basic streamlines

arose:
# Description

Streamlines support
 - Add basic calculation method: fixed step size in grid space
 - Add custom-volume-property: for caching and lifecycle management
 - Add representation with lines and tube-mesh visuals: gap and anchor support

This is based on @corredD's work in #1638.

Streamline seed points use PCG for reproducible pseudo random numbers.

Tube-mesh visual is constructed with frenet-frames from streamline points.

Lines visual uses line-strips by default for memory efficiency which commonly use 60% of what standard lines need.

image
Anchor: show only streamlines whose start or end point is within the anchor center and radius

image
Gap: skip streamline points; can be shifted for alternating animation effects

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Note that since we don't clearly distinguish between a public and private interf
     - Fix missing `usePalette` support
     - Fix vertex-based coloring for non-mesh geometries
     - Support line-strips
+    - Support vertex-based sizing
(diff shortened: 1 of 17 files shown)


corredD (approved):
Looking really good to me. Let me know when you are ready to merge and I will do couple of test.

PR #1774, opened 2026-02-14
Text label improvements

giagitom:
# Description

- Fix label attachment inversion (top/bottom and left/right)
- Improve label background vertical centering
- Add label near-clip fade out
- Handle label depth variant for correct transparent background and near fade
- Draw border under text using fragment depth to prevent overlap on adjacent characters
- Clamp border width to avoid exceeding SDF range
- Increase font atlas quality (2x font size multiplier)
- Use clip-space billboard rendering to avoid perspective distortion

output

## Actions

### DIFF
--- a/src/mol-gl/shader/text.vert.ts
+++ b/src/mol-gl/shader/text.vert.ts
@@ -103,11 +103,12 @@ void main(void){
     cornerOffset.y += offsetY;
 
     if (uHasHeadRotation) {
-        cornerOffset = (uInvHeadRotation * vec4(cornerOffset, 0.0, 0.0)).xy;
+        vec3 rotatedOffset = (uInvHeadRotation * vec4(cornerOffset, 0.0, 0.0)).xyz;
+        clip += uProjection * vec4(rotatedOffset, 0.0);
+    } else {
+        // apply offset in clip space to avoid perspective distortion on the quad
+        clip.xy += vec2(uProjection[0][0], uProjection[1][1]) * cornerOffset;
     }
-
-    // apply offset in clip space to avoid perspective distortion on the quad
-    clip.xy += vec2(uProjection[0][0], uProjection[1][1]) * cornerOffset;
 
     gl_Position = clip;
 


arose:
> Fix label attachment inversion (top/bottom and left/right)

that is actually intentional, it describes which part of label is attached to the target

src/mol-geo/geometry/text/text-builder.ts
```diff
@@ -66,60 +67,61 @@ export namespace TextBuilder {
                 let yShift: number, xShift: number;
                 // vertical
                 if (attachment.startsWith('top')) {
-                    yShift = bHeight;
```
arose, line 69: that is actually intentional, it describes which part of label is attached to the target

src/mol-gl/shader/text.frag.ts
```diff
@@ -26,7 +27,52 @@ void main(){
     #include fade_lod
     #include clip_pixel
 
+    // fade out label as anchor approaches the near clip plane
+    float anchorDepth = vViewPosition.z;
+    float nearFadeStart = uNear + (uFar - uNear) * 0.03;
+    float nearFade = smoothstep(uNear, nearFadeStart, anchorDepth);
+    if (nearFade <= 0.0) discard;
```
arose: I would prefer something like `if (nearFade < 1.0) discard;`

src/mol-gl/shader/text.vert.ts
```diff
@@ -73,39 +74,44 @@ void main(void){
...
+    // project center to clip space
+    vec4 clip = uProjection * mvCenter;
+
+    // compute corner offset in screen-space units
+    vec2 cornerOffset = aMapping * size * scale;
+    cornerOffset.x += offsetX;
+    cornerOffset.y += offsetY;
+
+    if (uHasHeadRotation) {
+        cornerOffset = (uInvHeadRotation * vec4(cornerOffset, 0.0, 0.0)).xy;
```
arose: This does not work, it needs to include a z value. Can you test this with a headset?

arose (changes requested):
can you also update the year in the headers please

src/mol-gl/shader/text.vert.ts
giagitom: @arose I committed a possible fix, could you test it?
I haven't a headset to test it.

arose: thanks, this works nicely

Issue #1715, opened 2025-12-08
How to filter secondary structure selections using pick?


zhonghua-wang:
Hi mol* team.

I follow the example in the [MolQL explorer](https://molql.org/explorer.html) and try to select **helices formed by at least 30 residues**, however either use mol-script like this:
```typescript
const script = Script(`(sel.atom.pick
  (sel.atom.atom-groups
  :residue-test (atom.sec-struct.is helix)
  :group-by atom.key.sec-struct)
  :test (<= 30 (atom.set.count-query sel.atom.res)))`, 'mol-script')
  
  const loci = Script.toLoci(script, structure)
```
nor build Expression using the following code works:
```typescript
  const query = MS.struct.filter.pick({
    "0":MS.struct.generator.atomGroups({
        'entity-test': _proteinEntityTest,
        'residue-test': MS.core.flags.hasAny([
            MS.ammp('secondaryStructureFlags'),
            MS.core.type.bitflags([2])
        ]),
        'group-by': MS.ammp('secondaryStructureKey')
    }),
    test: MS.core.rel.gre([
      MS.struct.atomSet.countQuery(
        [MS.struct.generator.atomGroups({
          'residue-test': true,
          'group-by': MS.ammp('residueKey')
        })]
      ),  30
    ])
  })
```

ps. The selection on the same test structure works as expect in MolQL explorer

PR #1796, opened 2026-03-23
fix macro expansion when used as arguments

zhonghua-wang:
When macros like `sel.atom.res` are used as arguments (e.g., in `(atom.set.count-query sel.atom.res)`), the transpiler throw an empty array `[]` to the macro's translate function, causing it to fail. This change passes an empty object `{}` instead, allowing macros to expand with their default arguments. Also adds `atom.sec-struct.is` macro for secondary structure testing and fixes indentation in symbol definitions.

related issue: #1715 

## Actions


src/mol-model/structure/query/queries/atom-set.ts
```diff
@@ -22,15 +22,22 @@ export function atomCount(ctx: QueryContext) {
...
     return (ctx: QueryContext) => {
-        const sel = query(ctx);
+        const { currentStructure } = ctx;
+        if (!currentStructure) {
+            const sel = query(ctx);
+            return StructureSelection.structureCount(sel);
+        }
+        const tempCtx = new QueryContext(currentStructure, { currentSelection: ctx.currentSelection });
+        tempCtx.currentStructure = currentStructure;
+        const sel = query(tempCtx);
```
dsehnal, line 34: Why is this needed?

src/mol-script/script/mol-script/symbols.ts
dsehnal: Can you delete this since you added it above?

src/mol-script/script/mol-script/symbols.ts
```diff
@@ -331,7 +334,7 @@ function substSymbols(expr: Expression): Expression {
         if (!SymbolMap[expr.name]) return expr;
         const s = SymbolMap[expr.name]!;
         if (s.kind === 'alias') return Expression.Symbol(SymbolMap[expr.name]!.symbol.id);
-        throw s.translate([]);
+        return s.translate({});
```
dsehnal, line 334: Oops, wonder how this happened :D

dsehnal (commented):
Can you please update the headers of the files you modified and add yourself to the package.json author list?

src/mol-model/structure/query/queries/atom-set.ts
zhonghua-wang: The newly added code provides a clean context for the execution of nested queries, which preserves `currentSelection` from parent context, as a result, each `countQuery` execution gets its own isolated context, ensuring correct counting behavior.
e.g.

```(sel.atom.pick
  (sel.atom.atom-groups
  :residue-test (atom.sec-struct.is helix)
  :group-by atom.key.sec-struct)
  :test (<= 30 (atom.set.count-query sel.atom.res)))
```

dsehnal: Right, but would it be enough to just cache the old value and restore it after the inner query call?

zhonghua-wang: As I understand it, with this approach, the nested query will use currentSelection rather than the full Structure  in the `structureCount` function.
```typescript
    export function structureCount(sel: StructureSelection) {
        if (isSingleton(sel)) return sel.structure.elementCount;
        return sel.structures.length;
    }
```

dsehnal: No, the "current selection" is only accessed via `MolScript.internal.generator.current`, which has a different use case.

There is no  `:test (<= 30 (atom.set.count-query sel.atom.res))` in `sel.atom.atom-groups`. You need to use `sel.atom.pick` to filter out the helices you are looking for.

src/mol-script/script/mol-script/symbols.ts
dsehnal: I have realised this is actually working as intended. You are supposed to "call" the function with `(atom.key.sec-struct)` (the extra `()`) and then things should work as expected. 

Symbols are not supposed to be applied automatically.

PR #1805, opened 2026-04-02
Procedural animation

arose:
# Description

Adds procedural animation to wiggle (move atoms) and tumble (rotate chains). It runs on the GPU, is very cheap and has absolutely no regard for physics or chemistry - might still be useful to convey dynamics or highlight uncertainty.

viewer: https://molstar.org/dev/anim/
me: https://molstar.org/dev/anim/me/

easiest way to activate is with the new panels

image

image

## Examples

https://github.com/user-attachments/assets/a7d83ebb-4040-47af-a124-6e45054e8de7

Wiggle atoms based on B-factor

https://github.com/user-attachments/assets/6b4d8fc1-8bbd-4128-83a5-fc761e06a78f

Convey dynamics for a simulation system

https://github.com/user-attachments/assets/77fabb33-cd47-4e63-8acf-e4a37eaad0ae

Mesoscale model dynamics

## Actions

### DIFF
--- a/src/mol-plugin-ui/structure/procedural-animation.tsx
+++ b/src/mol-plugin-ui/structure/procedural-animation.tsx
@@ -1,0 +1,113 @@
+/**
+ * Copyright (c) 2026 mol* contributors, licensed under MIT, See LICENSE file for more info.
+ *
+ * @author Alexander Rose <alexander.rose@weirdbyte.de>
+ */
+
+import { clearStructureWiggle, setStructureWiggleFromUncertainty } from '../../mol-plugin-state/helpers/structure-wiggle';
+import { CollapsableControls, PurePluginUIComponent } from '../base';
+import { Button } from '../controls/common';
+import { AnimationSvg } from '../controls/icons';
+
+export class StructureProceduralAnimationControls extends CollapsableControls {
+    defaultState() {
+        return {
+            isCollapsed: false,
+            header: 'Procedural Animation',
+            brand: { accent: 'gray' as const, svg: AnimationSvg }
+        };
+    }
+
+    renderControls() {
+        return <StructureProceduralAnimation />;
+    }
+}
+
+interface StructureProceduralAnimationState {
+    busy: boolean;
+}
+
+class StructureProceduralAnimation extends PurePluginUIComponent<{}, StructureProceduralAnimationState> {
+    state: StructureProceduralAnimationState = { busy: false };
+
+    private get components() {
+        return this.plugin.managers.structure.hierarchy.selection.structures.flatMap(s => s.components);
+    }
+
+    async applyUncertaintyWiggle() {
+        this.setState({ busy: true });
+        try {
+            const options = this.plugin.managers.structure.component.state.options;
+            await this.plugin.managers.structure.component.setOptions({
+                ...options,
+                animation: {
+                    ...options.animation,
+                    wiggleAmplitude: 0,
+                    tumbleAmplitude: 0
+                }
+            });
+            await setStructureWiggleFromUncertainty(this.plugin, this.components);
+        } finally {
+            this.setState({ busy: false });
+        }
+    }
+
+    async applyDynamics() {
+        this.setState({ busy: true });
+        try {
+            const options = this.plugin.managers.structure.component.state.options;
+            await this.plugin.managers.structure.component.setOptions({
+                ...options,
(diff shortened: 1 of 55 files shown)


giagitom:
Very nice implementation!

I see some separations at group level, particularly on cartoons and sticks (on cartoons only when using Uncertainty mode)

image

src/mol-plugin-ui/structure/procedural-animation.tsx
```diff
@@ -0,0 +1,113 @@
...
+        return <StructureProceduralAnimation />;
+    }
+}
+
+interface StructureProceduralAnimationState {
+    busy: boolean;
+}
+
+class StructureProceduralAnimation extends PurePluginUIComponent<{}, StructureProceduralAnimationState> {
+    state: StructureProceduralAnimationState = { busy: false };
```
dsehnal, line 31: This should likely come from the PluginContext is busy too, no?

src/mol-plugin-ui/structure/procedural-animation.tsx
```diff
@@ -0,0 +1,113 @@
...
+        }
+    }
+
+    async applyDynamics() {
+        this.setState({ busy: true });
+        try {
+            const options = this.plugin.managers.structure.component.state.options;
+            await this.plugin.managers.structure.component.setOptions({
+                ...options,
+                animation: {
```
dsehnal, line 61: Let's be explicit here that this is `proceduralAnimation`?

dsehnal (commented):
What would be a good way to make this available in MolViewSpec? Could be a "custom extension" -- but should this be on "representation" node?

src/mol-plugin-ui/structure/procedural-animation.tsx
arose: it follows the same pattern as in QuickStyles UI

arose:
> I see some separations at group level, particularly on cartoons and sticks (on cartoons only when using Uncertainty mode)

yes, we animate vertex positions without knowledge of atom connectivity. In 'position' mode things generally stay reasonably together because noise is position-based. In 'group' mode the noise is group based which can look better for some representations but will rip different groups apart. For bond cylinder imposter geometry the gaps are due to differently tilts of each bond half. If you disable impostors they will stay connected. When wiggle by uncertainty the wiggle amplitude changes per group (atom) which will inevitable "rip" geometry corresponding to different groups apart.

giagitom (approved):

corredD (commented):
Really excellent and clean procedural noise. This can help create very nice movies that feel more alive than a static model, without having to run a full molecular simulation.

I have only one comment: is there any reason not to support volume representations in this PR?

The isosurface is a mesh, so it could be displaced directly. The slice representation could have some jitter applied to the pixel values. The dots are spheres and could wiggle. The streamlines could wiggle and tumble. Finally, for direct volume rendering, what would it look like if the voxel coordinates were wobbled?

arose:
just added missing animation support for texture-mesh geometry

> I have only one comment: is there any reason not to support volume representations in this PR?
> 

it works but you need to go to the representation parameters. The "Procedural Animation" UI is only for structures, like "Quick Styles" Volume data is quite heterogeneous and it is unclear what wiggling would mean generally.

> The isosurface is a mesh, so it could be displaced directly. The slice representation could have some jitter applied to the pixel values. The dots are spheres and could wiggle. The streamlines could wiggle and tumble. Finally, for direct volume rendering, what would it look like if the voxel coordinates were wobbled?

isosurface now works, dots and streamlines already did. I excluded slice and direct-volume as it is too much of an effect and really conveying something physical, at least not when just animating the proxy geometry.

corredD:
> it works but you need to go to the representation parameters. The "Procedural Animation" UI is only for structures, like "Quick Styles" Volume data is quite heterogeneous and it is unclear what wiggling would mean generally.

Sorry, I hadn’t noticed. This is really excellent! The streamlines look great as they wiggle like spaghetti.

corredD (approved):
This is really nice, I will use it for all my next mesoscale movies !

src/mol-plugin-ui/structure/procedural-animation.tsx
arose: this maps to the low level geoetry animate params which may not be purely procedural in the future, so I would keep the name general

PR #1806, opened 2026-04-03
Kinemage file loading

russell-taylor:
# Description

This brings the NGL-based Kinemage file parser into an Molstar and builds an extension that is linked to the viewer app that enables both file/open and drag-and-drop handlers for the Kinemage file format. There is a README.md in the extensions/kinemage directory describing which features have been implemented and which are still missing.

It shoe-horns KiNG-like interaction buttons into the State Tree GUI objects using JSON objects to enable viewpoint selection, group/subgroup/master controls, and (by toggling visibility on Animate button) the two animation cycles that can be defined for Kinemages.

You'll probably want to do this as a squash merge. There were a lot of unfruitful side paths taken along the way.

## Actions

- [*] Added description of changes to the `[Unreleased]` section of `CHANGELOG.md`
- [*] Updated headers of modified files
- [*] Added my name to `package.json`'s `contributors`


russell-taylor:
Note that to get this to compile under Node 22.22.2 there were a few files that I had to adjust code to be able to handle both string and string[] inputs, which I changed by making them take the first entry in the string array if there was one and replacing with and empty string on failure.

russell-taylor:
[1xsoFH-kinemages.zip](https://github.com/user-attachments/files/26472279/1xsoFH-kinemages.zip)
Here are some Kinemage files to try out. One has ribbons. The other has an Animate button.

arose (changes requested):
Hi, some high level comments

- the kin parser should be part of the extension as it is only useful in its context (see g3d extension for formats in extensions)
- for the UI, please consider something like in the assembly-symmetry extension that would show in the right panel
- any node related changes that are not specific to this should be in a separate PR

russell-taylor:
I've removed all of the changes made to the original files that were not needed to hook in the kinemage code and to get the code to compile after I rebased to the latest master.

Still working on consolidating the kin files to the extensions directory and moving the UI elements.

russell-taylor:
Consolidated new files into the extensions/kinemage directory.

Still working on moving the GUI elements.

russell-taylor:
Okay, I've pulled the GUI controls to the right side. They are indeed much better there, and much cleaner not having to be JSON objects repurposed to do the job!

Let me know if there are other changes needed.

package-lock.json
arose, line 1: please don't commit a package-lock

src/servers/membrane-orientation/web-api.ts
```diff
@@ -49,7 +49,8 @@ async function predictMembraneOrientation(req: express.Request, res: express.Res
     try {
         const ctx = { runtime: SyncRuntimeContext, assetManager };
 
-        const entryId = req.params.id;
+        // Make sure this is a string, taking the first element if it's an array (which can happen if the query parameter is repeated)
```
arose: please move these changes to a separate PR and discuss why/when they are needed

src/extensions/kinemage/reader/ngl-based-parser.ts
```diff
@@ -0,0 +1,905 @@
...
+
+// import { Debug, Log, ParserRegistry } from '../globals'
+// import { Vector3 } from 'three'
+// import Parser from './parser'
+
+/// @todo Fill in commments
+
+import { Kinemage, RibbonObject } from './schema';
+
+function hsvToRgb (h: number, s: number, v: number) {
```
arose: could you please do a PR adding mol-util/spaces/hsv?

src/extensions/kinemage/reader/kinparser.ts
arose, line 1: from my side there is no need to name this ngl-based and generally it would good to clean this up a bit, remove unnecessary code & comments, and conform more to the style used in molstar

src/extensions/kinemage/behavior.ts
```diff
@@ -0,0 +1,530 @@
...
+import { PluginCommands } from '../../mol-plugin/commands';
+import { StateObjectRef } from '../../mol-state';
+import { getPluginBoundingSphere } from '../../mol-plugin-state/manager/focus-camera/focus-object';
+import { KinemageControls } from './ui';
+
+const Tag = KinemageData.Tag;
+
+const Transform = StateTransformer.builderFactory('sb-kinemage');
+
+let g_kinemageData: KinemageData | undefined = undefined;
```
arose: We want to avoid this kind of global state. The canonical way in molstar is to have a series of Transforms that get input from a parent (plus params) and output something for the next transfomer(s) in the state tree. You sidestep this by having a) all transfomers with root as parent b) use your global state as input and c) shove the whole kin data into the params. It would be great if you can refactor this to be more canonical ParseKinemageTransform (using the existing Transforms to get String data) to create Kinemage data and then have Transforms that make representations. This will also give you a more useful state tree, especially when loading multiple kin files.

src/extensions/kinemage/ui.tsx
arose, line 1: just two quick notes
- once a kin file is loaded there seems to be no way to get rid of the UI again - the UI display should depend on weather kin data is in the state tree or not
- the style could use some teaks to fit better into the rest of the UI

arose (changes requested):
Thank you for your patience and effort to iterate on it.

src/extensions/kinemage/reader/ngl-based-parser.ts
russell-taylor: Okay, moved the existing method into a class in that folder. Did not write the inverse-conversion function there but did modify the Kinemage parsing code to deal with Color rather than number[] internally.

src/extensions/kinemage/reader/kinparser.ts
russell-taylor: Renamed and removed commented-out code. I'm not familiar with either your original NGL style code or Molstar style code, so will let you do the style conversion.

russell-taylor:
@arose I've removed the global state and am now using Transforms, and also made the GUI elements disappear when the associated State Tree objects are deleted. I think the only unaddressed change is making the right-hand-side GUI elements match "the Molstar style", but I don't know what that style is so it will probably be faster for you to change this than to tell me how to change it.

src/extensions/kinemage/ui.tsx
russell-taylor: Fixed the UI destruction with State Tree destruction.
I don't know what style you'd like to match and hope that you're willing to tweak it to fit.

arose:
> @arose I've removed the global state and am now using Transforms, and also made the GUI elements disappear when the associated State Tree objects are deleted. I think the only unaddressed change is making the right-hand-side GUI elements match "the Molstar style", but I don't know what that style is so it will probably be faster for you to change this than to tell me how to change it.

thanks, will have look

russell-taylor:
@arose I've re-styled the right-hand-side buttons to try and better match the MolStar style. This project ends in July so I'd like to make sure we have everything you need for a merge sometime this month.

Thank you for maintaining MolStar and for entertaining pull requests!

arose (changes requested):
Hi, I went over it and did some linting/formatting. I could not push it into your branch so its here https://github.com/molstar/molstar/tree/kinemages if you want to pull from it.

There is still a bit of a misunderstanding how we usually use the state tree in molstar. The idea is that data in the tree nodes is immutable and we use transforms to create new nodes that may augment or take pieces of its parent data for itself. As far as I can tell your UI code mutates the kin data and than manually rebuilds the state tree.

A canonical way would be to parse the kin data and have transforms that create the visuals. These transforms would have parameters to control what parts of the kin data should be visible. Then in your UI you can either just expose the parameters with the built-in parameter UI or create a more specific custom UI that ultimately changes the parameters values.

Regarding the UI, I noticed that it is flat but objects in kinemages are usually arranged hierarchically, no?

russell-taylor:
I merged your changes into the branch. Thank you for those!

Regarding the flatness of the UI, Kinemage has group/subgroup relationships that cause indentation (which is implemented) and then a flat set of masters. Below are images from KiNG and Molstar showing the similarity of nesting.

Regarding the re-generation of the State Tree, the controls in Kinemage behave differently (all masters must be on) than the visibility controls in the State Tree (any visibility must be on). Also, there are "pointmaster" entries that are primitive-by-primitive entries that tie to masters. Because each point can be in a subgroup and a group and part of an arbitrary number of masters, doing this selection via groups would require a combinatorial explosion of Molstar visibility groups, one for every setting of every master and group and subgroup. The current approach of re-generation handles all of the cases and seems to update rapidly even for large models, so I hope it is an acceptable approach.

image
image

arose:
I don't mean to use the state tree for visibility, just have parameters in the transform that control the visibility.

russell-taylor:
Okay, I think I've done what you asked.

arose:
Thank you, this looks good to me. Can you please bring it up-to date with master and remove any commented out code that is not needed.

russell-taylor:
Done!

arose (approved):
Thank you. Will merge this. Please consider adding parser tests in a new PR.

arose:
before I merge, please fix the linter issues

russell-taylor:
Lint picked.
Please point me to the correct location to add parser tests, and perhaps an example.

arose:
Please expand src/extensions/kinemage/_spec/kin.spec.ts

russell-taylor:
Done.

russell-taylor:
Looks like this has been approved. I do not have permission to do the merge. Looking forward to the squash merge happening!

PR #1818, opened 2026-04-19
Fix GraphQL POST request handling for Assembly Symmetry

corredD:
# Description

I was trying to display the symmetry axis of 2PLV (Component->Preset->Annotation->Assembly Symmetry), but it raise an error (Error: Download failed with status code 415). I think maybe the PDB change some policies on their query API and are more strict and require the header. This PR seems to solve the problem.
 
## Actions


josemduarte:
Nice catch, thank you for fixing. Indeed the RCSB PDB Data API GraphQL implementation recently changed and it turns out that the previous implementation was too lenient in this case, incorrectly allowing `content-type: text/plain`, see the "Request format" section here: https://graphql.org/learn/serving-over-http/

arose (approved):
thanks

PR #1820, opened 2026-04-24
Add support for the ZML format

drailab:
# Description

Add support for the ZML (Zipped Molecular Lot) file format — a ZIP-based container produced by the Atommap's Python library — as both an input (reader) and output (writer) format in Mol*.

# What's included
## Reader (src/mol-io/reader/zml/)
- V1 archives: version + molsys.json (topology as nested molecules → residues → atoms) + pos.npy (NumPy-format coordinates).
- V3 archives: version + header.json + columnar binary topology buffers (Z.i16, fc.i16, atom_ids.i32, res_ids.i32, res_atom_offsets.i32, mol_res_offsets.i32, bonds.i32, bond_orders.i8) + pos.f64 coordinates.
- Topology is mapped to a Mol* Trajectory via trajectoryFromZml in src/mol-model-formats/structure/zml.ts, using EntityBuilder + ComponentBuilder + BasicSchema.
- Registered as a built-in trajectory format (ZmlProvider) so .zml files appear in both Open Files → Auto (via binary extension sniffing) and Open Files → Specific (dropdown entry "ZML"). The TrajectoryFromZML state transform wires everything together.
## Writer (src/mol-io/writer/zml.ts)
- encodeZml(runtime, name, structure) serializes a Structure to a V1 ZIP archive (version, molsys.json, pos.npy-equivalent raw little-endian float64 bytes).
- Exposed through the existing Export Models UI (src/extensions/model-export/) as a new zml format option.
## Tests (src/mol-io/reader/_spec/zml.spec.ts, 7 cases)
- V1 minimal archive parse.
- V3 minimal archive parse.
- Unsupported version (e.g. 2) rejected.
- Trajectory construction from V3 (frame count, atom count, bonds).
- Element symbol assignment verified against the V3 Z.i16 buffer.
- Wrong-length pos.f64 → error.
- encodeZml → parseZml round-trip integrity (atoms, positions, bonds).

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Note that since we don't clearly distinguish between a public and private interf
 - Handle CCD bonds with Deuterium atoms
 - [Breaking] ComponentBond.Entry.map now returns ComponentBond.Pairs
 - Fix volume slice marking performance regression
+- Add ZML (Zipped Molecular Lot) format support: reader for V1 and V3 archives (topology + coordinates; force field data ignored) and V1 writer
 
 ## [v5.8.0] - 2026-04-03
 - Dependencies: remove `utils.promisify`, `node-fetch` (#1797)
--- a/package.json
+++ b/package.json
@@ -125,7 +125,8 @@
     "Kim Juho <juho_kim@outlook.com>",
     "Victoria Doshchenko <doshchenko.victoria@gmail.com>",
     "Diego del Alamo <diego.delalamo@gmail.com>",
-    "Tianzhen Lin (Tangent) <tangent@usa.net>"
+    "Tianzhen Lin (Tangent) <tangent@usa.net>",
+    "Yujie Wu <yujie@atommap.com>"
   ],
   "license": "MIT",
   "devDependencies": {
--- a/src/extensions/model-export/export.ts
+++ b/src/extensions/model-export/export.ts
@@ -6,12 +6,15 @@
  */
 
 import { utf8ByteCount, utf8Write } from '../../mol-io/common/utf8';
+import { encodeZml } from '../../mol-io/writer/zml';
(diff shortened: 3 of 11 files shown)


arose:
Never heard of the ZML format before. What software writes it? Where can I find a spec and example files?

drailab:
Hi, thank you for asking.

ZML is currently a custom file format designed by [Atommap](www.atommap.com) to represent molecular systems. We introduced it because we needed a format that could capture more than atomic coordinates and molecular topology alone. For example, ZML can also store force field data required for molecular dynamics simulations, which allows us to support workflows with OpenMM.

We use Mol* internally for visualization of molecular structures and simulation trajectories, and this PR helps improve that workflow.

Atommap is still a relatively small post–Series A company, and we are planning to open up the ZML format more formally in the future, together with IO libraries and related tooling. In the meantime, I’m happy to attach a draft spec and an example ZML file for reference.

[zml_format.tar.gz](https://github.com/user-attachments/files/27088368/zml_format.tar.gz)

If this approach looks reasonable, I would greatly appreciate consideration for merging this PR.

arose:
I can give it some time but my current thinking is that this too much of an internal format that is only useful (at present) for you and should be your internal extension. That said we can certainly consider PRs that make creating an internal extension for a file format better.

am-atommap:
Thanks, Alex, for taking the time to review the PR.
I understand your reasoning and completely respect the decision not to accept it at this stage. Would you be open to us revisiting the PR after we make the file format specification and I/O source code publicly available?

arose:
> Thanks, Alex, for taking the time to review the PR. I understand your reasoning and completely respect the decision not to accept it at this stage. Would you be open to us revisiting the PR after we make the file format specification and I/O source code publicly available?

certainly

PR #1821, opened 2026-04-27
Add Volume Mask Creator extension

Poko18:
## Summary
- add a new `src/extensions/volume-mask` extension and standalone example
- support mask creation from volume thresholding or structure proximity, combined with drawn polygon views
- add chain selection highlighting plus volume utilities for save/apply/flip/dust workflows
- smooth threshold preview updates, including deferred slider refresh for large maps

## Demo
- https://poko18.github.io/molstar/

https://github.com/user-attachments/assets/e1a84e73-38bc-4d2e-a6d4-231c666a68dd

https://github.com/user-attachments/assets/0930c5fa-8e53-48dd-ae02-c27b2f58619c


src/extensions/volume-mask/internal/mrc-export.ts
```diff
@@ -0,0 +1,111 @@
...
+ * Minimal MRC/CCP4 binary writer (MODE=2, float32).
+ * Spec: https://www.ccpem.ac.uk/mrc_format/mrc2014.php
+ */
+
+import { Grid } from '../../../mol-model/volume';
+import { Vec3 } from '../../../mol-math/linear-algebra';
+
+const HEADER_BYTES = 1024;
+
+export function writeMrc(grid: Grid, maskData: Uint8Array | Float32Array): ArrayBuffer {
```
dsehnal: Can you add this to mol-io/writer instead please?

src/examples/volume-mask/ui/drawing-canvas.tsx
```diff
@@ -0,0 +1,206 @@
+/**
```
dsehnal, line 1: What do you think about keeping the compute code as an extension and moving the UI (as that is more opinionated) to the examples folder?

dsehnal (commented):
Cool extension!  Possibly few tweaks before we merge this.

Also, could you add your name to the contributors list in package.json?

src/extensions/volume-mask/internal/polygon.ts
arose, line 1: could you use/improve src\mol-math\geometry\polygon.ts?

Poko18: I added the polygon code to `geometry\polygon.ts`, but did not want to do any changes to the  previous API, so there is currently small duplication (about 10 lines) as while algorithm is similar, data layout is different

src/extensions/volume-mask/internal/mrc-export.ts
Poko18: Done

src/examples/volume-mask/ui/drawing-canvas.tsx
Poko18: Good point and I completely agree. Moved UI code to examples

Poko18:
Thanks for looking into this PR! I think I addressed most of the concerns. Let me know if there's anything else

dsehnal (commented):
Sorry for the review delay.

Looks good I think, thanks for the contribution. Before margining this, can you please update CHANGELOG.md as well?

src/mol-math/geometry/polygon.ts
```diff
@@ -22,3 +22,17 @@ export function pointInPolygon(point: Vec2, polygon: NumberArray, count: number)
     }
     return inside;
 }
+
+/** raycast along x-axis and apply even-odd rule; accepts an array of [x, y] tuples */
```
arose, line 26: Let's move this back into the extension. I would not want two helpers in core that do the same thing like this.

Poko18: done

Poko18:
I added the changes.. had to rebase, seems like new version was released in the meantime. Anyway, I think we are all set now!

dsehnal (approved):

PR #1826, opened 2026-05-08
Fix `State.dispose()` leaking live-cell transformer resources

### LINKED ISSUE #1825: PluginContext.dispose() leaks cells: per-transformer dispose() not called for live cells at teardown
## Summary

`PluginContext.dispose()` does not flush the state tree before tearing the plugin down, so transformer-level `dispose()` callbacks are never invoked for cells that are still present at unmount. This silently leaks the GPU/native resources those callbacks were responsible for releasing — most visibly in single-page apps that mount/unmount plugin contexts repeatedly.

The contract on `Transformer.Definition.dispose` says *"automatically called on deleting an object"* (`mol-state/transformer.ts`, comment above the optional `dispose` field), but `State.dispose()` (`mol-state/state.ts:161`) currently only disposes its own event subjects and action manager — it does **not** iterate the still-live cells to call their per-cell dispose. Cells holding GL buffers, mesh data, etc. only get their dispose fired on explicit deletion (e.g. `clear()`), so any consumer that calls `plugin.dispose()` without first awaiting `plugin.clear()` retains the callback chain.

```js
// mol-state/state.ts (current upstream master)
dispose() {
    this.ev.dispose();
    this.actions.dispose();
}
```

## Reproduction

E2E mount/unmount harness driving 20 cycles of `createPluginUI` → load one structure (1AKE, ~417 KB CIF) → unmount, sampled with `HeapProfiler.collectGarbage` + heap snapshot diff between cycle-0 and cycle-20.

| disposal sequence in consumer | Δ JS heap (post-GC, 20 cycles) | Δ Chrome process RSS (20 cycles) |
|---|---|---|
| `plugin.dispose()` | +266 MB | +541 MB |
| `await plugin.clear(); plugin.dispose()` | +217 MB | +287 MB |

Heap snapshot diff (`plugin.dispose()` only) shows the typical fingerprint of skipped per-cell dispose:

| heap object class | count delta (20 cycles) | hint |
|---|---|---|
| `system / Context` | +74,312 | retained closures |
| generic `Object` | +164,796 | state cell `obj` / `cache` / `params` |
| `array` | +134,734 | internal Mol* arrays |
| `destroy` | +19,361 | GL resource cleanup callbacks queued but never fired |
| `ArrayBuffer` | +9,671 | raw GL buffers |
| `Uint32Array` | +7,410 | vertex/index buffers |
| `native` | +30,839 (~167 MB) | DOM/C++ wrappers tied to those callbacks |

The +19k `destroy` callbacks correspond closely to the +74k closures (~4 closures per `destroy` site). Calling `await plugin.clear()` before `plugin.dispose()` flushes them, halving the residual leak — confirming the per-cell dispose path is what's missing on the unconditional `dispose()` route.

## Proposed fix

Have `State.dispose()` walk the tree once and invoke the per-transformer `dispose` for every still-live cell, descendants first (so children release before parents):

```js
// mol-state/state.ts
dispose() {
    // Iterate cells in reverse-topological (post-order) so each transformer's
    // dispose runs after its children's, matching the deletion order in
    // applyChanges().
    const order: StateTransform.Ref[] = []
    const root = this._tree.root.ref
    StateTree.doPreOrder(this._tree, this._tree.transforms.get(root)!, { refs: order }, (n, _, s) => { s.refs.push(n.ref); return true })
    for (let i = order.length - 1; i >= 0; i--) {
        const cell = this.cells.get(order[i])
        if (!cell) continue
        const def = cell.transform.transformer.definition
        if (!def.dispose) continue
        try {
            def.dispose({ b: cell.obj !== StateObject.Null ? cell.obj : undefined, params: cell.transform.params, cache: cell.cache }, this._parent.globalContext)
        } catch (e) {
            console.warn('Error in transformer dispose during State.dispose', e)
        }
    }
    this.ev.dispose();
    this.actions.dispose();
}
```

(The exact iteration helper depends on what's idiomatic in `mol-state` — `doPreOrder` is one option already in the file; `StateTreeBuilder` ordering would also work.)

This makes the disposal sequence consistent: a consumer that calls `plugin.dispose()` after a successful mount no longer leaks the cells' GL buffers and per-transformer caches. Existing consumers that `await plugin.clear()` first remain correct (the cells are gone by then, so the new loop is a no-op).

## Why it matters

In a long-running single-page app where the user navigates away and back from a route that hosts a Mol* viewer, every mount/unmount cycle leaks roughly 25–50 MB of process RSS even with `plugin.dispose()` correctly called. We've measured this in production (Boltz Bio's design platform — Edge/Windows on iGPU): customers reported multi-GB Chrome RSS within ~20 minutes of activity. Adding `await plugin.clear()` before `plugin.dispose()` on our side cut the per-cycle leak by ~46% and prompted this investigation.

## Environment

- Mol* 5.8.0 (also present in `master` as of fetching `mol-state/state.ts` for this report)
- Chrome 147 (headless, ANGLE/SwiftShader for reproducibility, but the leak path is the same on real GPU)
- React 18 host, single `PluginUIContext` per mount

Happy to send a PR with the iteration + a small unit test if this analysis tracks. Want to align on the iteration helper / where to put the dispose callsite (inside `State.dispose` vs. inside `PluginContext.dispose` calling `clear()` automatically) before I cut one.

armyalpaca:
## Summary

`PluginContext.dispose()` does not flush the state tree before tearing the plugin down, so transformer-level `dispose()` callbacks are never invoked for cells that are still present at unmount. Per-cell dispose only fires from `updateNode` / `findDeletes` paths, so any consumer that calls `plugin.dispose()` without first awaiting `plugin.clear()` silently leaks the GPU/native resources those callbacks were responsible for releasing.

The contract on `Transformer.Definition.dispose` says *"automatically called on deleting an object"* (`mol-state/transformer.ts`), but `State.dispose()` (`mol-state/state.ts`) currently only disposes its own event subjects and action manager.

Fixes #1825.

## Reproduction (from the issue)

E2E mount/unmount harness driving 20 cycles of `createPluginUI` → load 1AKE (~417 KB CIF) → unmount, with `HeapProfiler.collectGarbage` + heap snapshot diff between cycle 0 and cycle 20:

| disposal sequence in consumer | Δ JS heap (post-GC, 20 cycles) | Δ Chrome process RSS (20 cycles) |
|---|---|---|
| `plugin.dispose()` | +266 MB | +541 MB |
| `await plugin.clear(); plugin.dispose()` | +217 MB | +287 MB |

Heap diff fingerprint of the unconditional-`dispose()` case (skipped per-cell dispose):

| heap object class | count delta (20 cycles) | hint |
|---|---|---|
| `system / Context` | +74,312 | retained closures |
| `Object` | +164,796 | state cell `obj` / `cache` / `params` |
| `array` | +134,734 | internal Mol* arrays |
| `destroy` | +19,361 | GL resource cleanup callbacks queued but never fired |
| `ArrayBuffer` | +9,671 | raw GL buffers |
| `Uint32Array` | +7,410 | vertex/index buffers |
| `native` | +30,839 (~167 MB) | DOM/C++ wrappers tied to those callbacks |

## Fix

`State.dispose()` walks the cell tree once via the existing `StateTree.doPostOrder` helper and invokes per-transformer `dispose` for every still-live cell, reusing the same per-cell `dispose()` helper that `updateNode` / `findDeletes` already use. Errors thrown by a single transformer's dispose are caught and logged (`console.warn`) so they don't prevent siblings from cleaning up.

Existing consumers that already `await plugin.clear()` before `plugin.dispose()` remain correct — by then `cells` only contains the root, so the new loop is effectively a no-op.

## Tests

Added `src/mol-state/_spec/state.spec.ts` covering:
- Chained transformer chain (root → leafA → leafB) — both `dispose` callbacks fire exactly once.
- Sibling subtrees — both fire.
- Throwing-dispose isolation — a transformer whose `dispose` throws does not prevent siblings from being disposed and does not surface the error.
- Transformer without a `dispose` definition — no-op.

Sanity check: with the fix reverted, 3 of 4 tests fail; with the fix applied, all 4 pass and the existing `1061` jest tests still pass.

## Jest transform tweak

Adds `tsconfigRaw: '{"compilerOptions":{"useDefineForClassFields":false}}'` to the jest esbuild transform. The `TransientTree` (and several other classes) rely on parameter-property-before-class-field initialization order — which `tsc` honors via `target: es2018` but esbuild's default `esnext` target does not. Without this, `State.create()` throws `Cannot read properties of undefined (reading 'transforms')` from inside `TransientTree`, which is why none of the existing specs exercise `State` directly. Happy to split this into a separate prep PR if preferred.

## Why it matters

Production reports of multi-GB Chrome RSS within ~20 minutes of activity in single-page apps that mount/unmount a Mol* viewer per route. Adding `await plugin.clear()` before `plugin.dispose()` on the consumer side cut the per-cycle leak by ~46% and prompted this investigation.

## Environment

- Mol* `master` (also reproduced on 5.8.0 / 5.9.0)
- Chrome 147 (headless, ANGLE/SwiftShader for reproducibility — leak path is the same on real GPU)
- React 18 host, single `PluginUIContext` per mount

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
       "js"
     ],
     "transform": {
-      "\\.ts$": "esbuild-jest-transform"
+      "\\.ts$": ["esbuild-jest-transform", { "tsconfigRaw": "{\"compilerOptions\":{\"useDefineForClassFields\":false}}" }]
     },
     "moduleDirectories": [
       "node_modules",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Note that since we don't clearly distinguish between a public and private interf
 
 ## [Unreleased]
 - Fix empty transforms default in `ShapeFromPly`
+- Fix memory leak in `State.dispose()` not invoking transformer `dispose` callbacks for live cells
 
 ## [v5.9.0] - 2026-05-03
 - Fix edge case when `PluginSpec.animations` is empty
--- a/src/mol-state/_spec/state.spec.ts
+++ b/src/mol-state/_spec/state.spec.ts
@@ -1,0 +1,126 @@
+/**
+ * Copyright (c) 2026 mol* contributors, licensed under MIT, See LICENSE file for more info.
+ */
+
+import { State, StateObject, StateTransformer } from '../../mol-state';
+import { Task } from '../../mol-task';
+
+interface TypeInfo { name: string; typeClass: 'Root' | 'Data' }
+const Create = StateObject.factory<TypeInfo>();
+
+class Root extends Create({ name: 'Root', typeClass: 'Root' }) { }
+class Leaf extends Create<{ value: number }>({ name: 'Leaf', typeClass: 'Data' }) { }
+
+const NS = 'state-dispose-spec';
+let counter = 0;
+
+function leafTransformer(spy: () => void) {
+    return StateTransformer.create<Root, Leaf, { value: number }>(NS, {
+        name: `create-leaf-${counter++}`,
+        from: [Root],
+        to: [Leaf],
+        display: { name: 'Create Leaf' },
+        params: () => ({} as any),
+        apply({ params }) { return new Leaf({ value: params.value }); },
+        dispose() { spy(); }
+    });
+}
+
+function chainedTransformer(spy: () => void) {
+    return StateTransformer.create<Leaf, Leaf, {}>(NS, {
+        name: `chained-leaf-${counter++}`,
+        from: [Leaf],
+        to: [Leaf],
+        display: { name: 'Chained Leaf' },
+        apply({ a }) { return new Leaf({ value: a.data.value + 1 }); },
+        dispose() { spy(); }
+    });
+}
+
+function newState() {
+    return State.create(new Root({}), { runTask: <T>(t: Task<T>) => t.run() });
+}
+
+describe('State.dispose', () => {
+    it('calls transformer.dispose for every live cell', async () => {
+        const leafSpy = jest.fn();
+        const chainSpy = jest.fn();
+        const A = leafTransformer(leafSpy);
+        const B = chainedTransformer(chainSpy);
+
+        const state = newState();
+        const builder = state.build();
+        builder.toRoot<Root>().apply(A as any, { value: 1 }).apply(B as any, {});
+        await state.runTask(state.updateTree(builder));
+
+        // root + 2 transformer outputs.
+        expect(state.cells.size).toBe(3);
+
+        state.dispose();
+
+        expect(leafSpy).toHaveBeenCalledTimes(1);
+        expect(chainSpy).toHaveBeenCalledTimes(1);
+    });
+
+    it('disposes all sibling subtrees', async () => {
+        const spyA = jest.fn();
+        const spyB = jest.fn();
+        const A = leafTransformer(spyA);
+        const B = leafTransformer(spyB);
+
+        const state = newState();
+        const builder = state.build();
+        builder.toRoot<Root>().apply(A as any, { value: 1 });
+        builder.toRoot<Root>().apply(B as any, { value: 2 });
+        await state.runTask(state.updateTree(builder));
+
+        state.dispose();
+
+        expect(spyA).toHaveBeenCalledTimes(1);
+        expect(spyB).toHaveBeenCalledTimes(1);
+    });
+
+    it('does not throw when a transformer dispose throws', async () => {
+        const goodSpy = jest.fn();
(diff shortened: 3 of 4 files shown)


dsehnal (approved):
Nice find, thanks for the fix!

package.json
```diff
@@ -74,7 +74,7 @@
       "js"
     ],
     "transform": {
-      "\\.ts$": "esbuild-jest-transform"
+      "\\.ts$": ["esbuild-jest-transform", { "tsconfigRaw": "{\"compilerOptions\":{\"useDefineForClassFields\":false}}" }]
```
arose, line 77: what is this for and why is it needed?

PR #1827, opened 2026-05-09
Camera helpers

arose:
# Description

Add `Camera.changed` event and rotation/translation setter/getter

## Actions

### DIFF
--- a/src/mol-canvas3d/camera.ts
+++ b/src/mol-canvas3d/camera.ts
@@ -7,7 +7,7 @@
 
 import { Viewport, cameraProject, cameraUnproject } from './camera/util';
 import { CameraTransitionManager } from './camera/transition';
-import { BehaviorSubject } from 'rxjs';
+import { BehaviorSubject, Subject } from 'rxjs';
 import { Scene } from '../mol-gl/scene';
 import { assertUnreachable } from '../mol-util/type-helpers';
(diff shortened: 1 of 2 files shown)


src/mol-canvas3d/camera.ts
```diff
@@ -70,6 +77,8 @@ export class Camera implements ICamera {
 
     readonly transition: CameraTransitionManager = new CameraTransitionManager(this);
     readonly stateChanged = new BehaviorSubject<Partial<Camera.Snapshot>>(this.state);
+    /** Fires whenever update() produces a changed view/projection (covers all mutations, including direct ones from controls). */
+    readonly changed = new Subject<void>();
```
dsehnal, line 81: Pass along  the camera object for easier access?

And how is this different from stateChanged?

dsehnal (approved):

src/mol-canvas3d/camera.ts
arose: stateChanged always fires (even if not actually changed) and doesn't fire during transition

arose: side note: should we cleanup the camera to close the events?

dsehnal: What do you mean by this?

arose: have a dispose method on the camera that calls .complete()

PR #1831, opened 2026-05-16
Add chain splitting functionality

prihoda:
Solves my small but long standing frustration with Mol* compared to PyMol, that we can't easily toggle off individual chains or change their representations.

Disclaimer - this is completely vibe coded :) Would love to hear your feedback, happy to polish this further.

Implements manual and automatic chain splitting features:

Auto split: "Split Chains" option in Download Structure dialog (Options section), demo:

https://github.com/user-attachments/assets/79064bfa-f61c-496d-894c-e5367691c3c5

Manual split: apply "Split into Chains" action on Polymer/Water/Ligand components, demo:

https://github.com/user-attachments/assets/35191746-8579-4f70-aae0-358f5e6fe0ea

Features:
- Splits structure components into individual chains (Chain A, B, C, etc.)
- Preserves and copies all representations to each chain
- Automatically hides original component after split (perhaps it could be deleted instead? but I figured it's better to keep it in case user needs to go back to it)
- Only applicable to StructureComponent nodes (not Assembly)

Technical details:
- Uses existing chains() query generator to group atoms by chain
- Creates StructureElement.Bundle references against root structure
- Reusable splitPolymerChains() helper function
- Chain labels use label_asym_id or auth_asym_id from structure properties

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ Note that since we don't clearly distinguish between a public and private interf
 - Add presets option to `ObjectList` param definition
 - Fix memory leak in `State.dispose()` not invoking transformer `dispose` callbacks for live cells
 - Fix `Volume` and `Isosurface` getBoundingSphere ignoring instances
+- Add chain splitting functionality for polymer structures
+    - Manual split via "Split into Chains" action on structure components
+    - Automatic split via "Split Chains" option in Download Structure dialog
+    - Individual chains (Chain A, B, C, etc.) with preserved representations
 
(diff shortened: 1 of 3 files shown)


arose:
How about doing this as an representation preset (src\mol-plugin-state\builder\structure\representation-preset.ts). it would show up here 
image

prihoda:
Hi @arose thanks for the feedback. Do you mean to add a preset on top of the current logic or to replace some/all of the logic by the preset?

arose:
I'd suggest all the logic in the preset.

PR #1835, opened 2026-05-19
Add Polymer Terminus representation with N/C-term directional cones

prihoda:
Adds a new 'polymer-terminus' structure representation that places octagonal pyramid cones at N-termini (dark blue, tip pointing into chain) and C-termini (dark red, tip pointing away from chain end), including chain break endpoints.

image

## Actions


dsehnal:
Instead of separate representation, would it make more sense to add this as an extra visual for cartoon?

We already have for example the "direction wedge"  there:

image

arose:
I would like to see it as a visual of the cartoon representation, too.

The coloring is not optimal but maybe you can get something acceptable with the "cartoon" color theme with sequence-id for sidechains.

PR #1815, opened 2026-04-13
Add Particles data model

corredD:
# Description

This PR introduces particles as a core data model (along structure, volume, shape)

It adds:
- raw readers for RELION STAR, Dynamo TBL, and CryoET Data Portal NDJSON
- a ParticleList and Loci/Location
- ParticlesRepresentation (only orientation so far)
- Transform and Action to create a Structure of a Model and a ParticleList

Still draft as we consider Particles to support not only a homogeneous ParticleList but heterogeneous collections and trajectories as well. 

## Actions

- [ ] Added description of changes to the `[Unreleased]` section of `CHANGELOG.md`
- [ ] Updated headers of modified files
- [ ] Added my name to `package.json`'s `contributors`
- [ ] (Optional but encouraged) Improved documentation in `docs`

PR #1838, opened 2026-05-23
Dynamic state tree dependencies

arose:
# Description

- collect dependencies from ValueRef/DataRef params
- optional .getDependencies()
- cycle detection
- suspend/resume state tree evaluation

This enables/fixes a few things
- state loading of ValueRefs pointing to refs later in the tree
- updating ValueRefs when ref changes
- transforms can handle their own dependencies (no need for extra action)

The particles support in #1815 is one motivation as this will make working with particle structures/volumes much easier.

## Actions

### DIFF
--- a/src/mol-state/transform.ts
+++ b/src/mol-state/transform.ts
@@ -9,6 +9,7 @@ import { StateTransformer } from './transformer';
 import { UUID } from '../mol-util';
 import { hashMurmur128o } from '../mol-data/util/hash-functions';
 import { ParamDefinition as PD } from '../mol-util/param-definition';
+import { arraySetAdd, arraySetRemove } from '../mol-util/array';
 
 export { Transform as StateTransform };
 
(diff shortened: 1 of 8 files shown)


src/mol-state/transform.ts
```diff
@@ -170,6 +172,81 @@ namespace Transform {
...
+     */
+    export function getEffectiveDependsOn(t: Transform, globalCtx?: unknown): Ref[] {
+        const out: Ref[] = [];
+        const seen = new Set<string>();
+        const add = (ref: string | undefined) => {
+            if (!ref || ref === t.ref || ref === RootRef) return;
+            if (seen.has(ref)) return;
+            seen.add(ref);
+            out.push(ref as Ref);
+        };
```
dsehnal: Can we allocate this only if there are actual dependencies?

src/mol-state/transform.ts
```diff
@@ -170,6 +172,81 @@ namespace Transform {
...
+            } catch {
+                // Keep reconciliation robust if a user hook misbehaves.
+            }
+        }
+
+        if (params) {
+            let schema: PD.Params | undefined = void 0;
+            if (def.params) {
+                try {
+                    schema = def.params(undefined as any, globalCtx) as PD.Params;
```
dsehnal: Not a big fan of this. At least cache it on the transform object?

Something like 

```ts
namespace Transform {
  export function tryGetDefaultParamDefinition(t: Transform) {
    if ((t as any)._defaultSchema) return ...
    try { ... } catch { ... }
  }
}
```

The serialisation should work because of the `Transform.toJSON` call.

src/mol-util/param-definition.ts
```diff
@@ -445,6 +445,45 @@ export namespace ParamDefinition {
...
+        }
+    }
+
+    /**
+     * Collect all non-empty `ref` strings of `value-ref` and `data-ref` parameter
+     * values into a set. Used by `mol-state` to derive transform dependencies from
+     * parameter values.
+     */
+    export function collectRefs(params: Params, values: any, out?: Set<string>): Set<string> {
+        const result = out ?? new Set<string>();
```
dsehnal: We can avoid the `Set` declaration for every params if we wrap in in a value ref

```ts
const _refSetWrapper: { set?: Set<string> } = { set: undefined }
function collectRefs(...): Set<string> | undefined {
  _refSetWrapper.set = out
  // ...
  return _refSetWrapper.set
}

// ...
if (!_refSetWrapper.set) _refSetWrapper.set = new Set();
```

dsehnal (commented):
Nice addition.

Most states should not be taking advantage of this, so would like to avoid the extra allocations.

dsehnal (approved):

PR #1841, opened 2026-05-27
Camera improvements

dsehnal:
# Description

- [x] Approximate "least obstructed direction" when focusing camera
- [x] Ability to "zoom out" camera when focusing loci
- [x] Camera transition easing
- [ ] Refactoring/API improvements 

https://github.com/user-attachments/assets/f72de5a4-15fb-4690-8488-792aa165577e

https://github.com/user-attachments/assets/3fa59b1c-8505-4cf0-a07c-54dcfcce397b

## Actions

### DIFF
--- a/src/mol-math/linear-algebra/3d/optimize-direction.ts
+++ b/src/mol-math/linear-algebra/3d/optimize-direction.ts
@@ -34,11 +34,54 @@ function eachPosition(points: ReadonlyArray<Vec3> | { x: ArrayLike<number>, y: A
 }
 
 /**
- * Returns a direction from the selection centroid toward the camera.
+ * Estimate a visually open camera direction around a selection.
  *
- * Input points should usually be nearby, non-selected atom positions.
+ * Geometric intuition:
  *
- * The returned direction is a unit Vec3 or undefined if no valid direction could be computed.
+ * The selection centroid is treated as the origin. Each nearby obstruction
+ * point is converted into a unit direction on the sphere around the selection:
+ *
+ *     v_i = normalize(p_i - origin)
+ *
+ * We then build the directional second-moment matrix:
+ *
+ *     M = sum_i w_i v_i v_i^T
+ *
+ * For any candidate view direction `u`, the quadratic form
+ *
+ *     u^T M u
+ *
+ * expands to:
+ *
+ *     sum_i w_i (u · v_i)^2
+ *
+ * Since `u · v_i = cos(theta_i)`, this value is large when `u` is aligned
+ * with many obstruction directions and small when `u` is mostly perpendicular
+ * to them. Therefore, the eigenvector of `M` with the smallest eigenvalue is
+ * the axis that is least aligned, in a least-squares sense, with the nearby
+ * obstruction directions.
+ *
+ * This gives an unoriented axis: `u` and `-u` have the same score because the
+ * dot products are squared. To choose the camera-facing side, we compute the
+ * weighted mean obstruction direction:
+ *
+ *     m = sum_i w_i v_i
+ *
+ * and return the sign of the axis that points away from this mean direction.
+ *
+ * In short:
+ *
+ * - project nearby points onto a sphere around the selection;
+ * - find the sparsest angular axis using the smallest eigenvector of their
+ *   second-moment matrix;
+ * - choose the side of that axis opposite the average obstruction direction.
+ *
+ * This is a fast, deterministic heuristic. It minimizes average squared
+ * angular alignment with nearby points; it is not the exact largest-empty-cone
+ * or maximum-clearance solution.
+ *
+ * The returned vector is a unit direction from the selection centroid toward
+ * the camera.
  */
 export function leastObstructedDirection(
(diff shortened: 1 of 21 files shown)


src/mol-math/linear-algebra/3d/optimize-direction.ts
```diff
@@ -0,0 +1,141 @@
...
+        const { x, y, z } = points as { x: ArrayLike<number>, y: ArrayLike<number>, z: ArrayLike<number> };
+        const n = Math.min(x.length, y.length, z.length);
+        for (let i = 0; i < n; i++) {
+            callback(x[i], y[i], z[i]);
+        }
+    }
+}
+
+/**
+ * Returns a direction from the selection centroid toward the camera.
```
arose: Could you please explain how this works geometrically? I have a hard time understanding what happens here.

dsehnal: I started with the Kabsch algorithm for the RMSD superposition () and sort of winged it from there with an LLM :)

Added description to the header.

tsconfig.commonjs.json
```diff
@@ -1,6 +1,7 @@
 {
     "extends": "./tsconfig.json",
     "compilerOptions": {
+        "moduleResolution": "node",
```
arose: maybe

"moduleResolution": "nodenext",
"module": "nodenext",

src/mol-plugin-state/manager/camera.ts
```diff
@@ -115,22 +198,60 @@ export class CameraManager {
...
-            this.plugin.canvas3d?.requestCameraReset({ durationMs, snapshot });
+            return pcaFocus(this.plugin, radius, options as { principalAxes: PrincipalAxes, positionToFlip?: Vec3 });
         } else {
-            const snapshot = canvas3d.camera.getFocus(sphere.center, radius);
-            canvas3d.requestCameraReset({ durationMs, snapshot });
+            return canvas3d.camera.getFocus(sphere.center, radius);
+        }
+    }
+
+    private focusSnapshot(snapshot: Partial<Camera.Snapshot> | undefined, options?: Partial<CameraFocusOptions>) {
```
arose, line 215: use this to transition between snapshots in the UI? and expose the zoom-out option as a snapshot transition option?

src/mol-math/linear-algebra/3d/quat.ts
```diff
@@ -71,6 +71,10 @@ namespace Quat {
         return out;
     }
 
+    export function is(a: any): a is Quat {
```
arose, line 74: could be a Vec4... I am not sure that these are better than context specific checks or casts

src/mol-model/structure/structure/element/loci.ts
```diff
@@ -572,6 +572,42 @@ export namespace Loci {
...
+                set.add(unitIdx);
+            }
+        });
+
+
+        const elements: Element[] = [];
+        for (const [unitId, indexSet] of elementsByUnit.entries()) {
+            const unit = loci.structure.unitMap.get(unitId)!;
+            const indices = Array.from(indexSet) as UnitIndex[];
+            indices.sort((a, b) => a - b);
```
arose: sortArray(indices);

src/mol-math/linear-algebra/3d/quat.ts
dsehnal: could still be useful in cases where you have `A | B | Quat` where you know they don't overlap.

src/mol-plugin-state/manager/camera.ts
dsehnal: Yes, eventually. Would like to play around with the separately before making it the default.

tsconfig.commonjs.json
dsehnal: Yeah, this is actually a downstream thing from the `bundler` resolution you added few weeks ago. Still want to support "old" nodejs versions for a bit. I am not sure what node version are they using at PDBe for the servers.

src/mol-math/linear-algebra/3d/quat.ts
arose: Euler then too?

src/mol-plugin-state/manager/camera.ts
arose: I mean expose it here. Doesn't have to be default

image

tsconfig.commonjs.json
```diff
@@ -1,6 +1,7 @@
 {
     "extends": "./tsconfig.json",
     "compilerOptions": {
+        "moduleResolution": "node16",
         "module": "CommonJS",
```
arose: Option 'module' must be set to 'Node16' when option 'moduleResolution' is set to 'Node16'.

src/mol-plugin-state/manager/camera.ts
dsehnal: Yes, wanna do that too, but in a separate PR later once this gets tested first. @midlik also has some improvements he can do in the future. Don't want to rush it.

PR #1842, opened 2026-06-01
add obj format support

arose:
# Description

Add obj format support. Only geometry. Group for each material.

## Actions


src/mol-io/reader/obj/parser.ts
```diff
@@ -1,22 +1,319 @@
...
+                // "vn" — vertex normal
+                tokenizer.position += 2;
+                handleNormal(state);
+            } else {
+                // "vt", "vp", etc — skip
+                skipLine(tokenizer);
+            }
+        } else {
+            // "g", "o", "s", "usemtl", "mtllib", etc. — skip entire line
+            skipLine(tokenizer);
```
corredD, line 307: This works well but do you plan to support the .mtl file in the future ?

src/mol-io/reader/obj/parser.ts
```diff
@@ -1,22 +1,319 @@
...
+            const idx = state.materialNames.length;
+            state.materialMap.set(name, idx);
+            state.materialNames.push(name);
+        }
+        state.currentMaterialIdx = state.materialMap.get(name)!;
+    }
+    skipLine(tokenizer);
+}
+
+function handleVertex(state: State): void {
```
corredD, line 188: maybe add support for vertex color r g b ? MeshLab can export them.

src/mol-model-formats/shape/obj.ts
```diff
@@ -0,0 +1,176 @@
...
+ * Build a mesh from indexed OBJ data.
+ *
+ * OBJ stores positions and normals as indexed arrays (each face-vertex
+ * references a position index and an optional normal index). Mesh.create
+ * requires a flat vertex array, so we expand the indexed data.
+ *
+ * A vertex key encodes (posIdx, normIdx) so that a shared position with
+ * different normals gets a distinct mesh vertex.
+ */
+async function getMesh(ctx: RuntimeContext, obj: ObjFile, mesh?: Mesh): Promise<Mesh> {
```
corredD: The comment says vertices are keyed by (posIdx, normIdx), but getMesh currently creates three new vertices per triangle with no deduplication. This makes normal-less files render faceted, since computeNormals cannot average across shared vertices, and it also makes the mesh larger. Is that intended ?

corredD (approved):
Nice format addition, do you plan to add texture or vertex color support ? Also the obj format support free form geometry which I didn't know! look at https://paulbourke.net/dataformats/obj/ a nice description of the format from Paul Bourke. But I dont think it is used a lot and probably not worth supporting. But the vertex color, and maybe the texture mapping are worth considering.

src/mol-io/reader/obj/parser.ts
arose: added support for sideloading mtl files, textures out of scope)

src/mol-io/reader/obj/parser.ts
arose: added

src/mol-model-formats/shape/obj.ts
arose: refactored to use an index mesh when aplicable

corredD (approved):
looks good to me.

PR #1848, opened 2026-06-07
Bloom on transparent geometry

giagitom:
# Description

Adds bloom support for transparent and emissive geometry, and fixes bloom rendering on a transparent background.

image

- Enable bloom on transparent/emissive geometry — emissive fragments are no longer discarded under transparency; they're scaled by alpha so semi-transparent surfaces emit, and the emissive pass is split into opaque and transparent (MAX-blended) renders.
- Move bloom into the postprocessing/illumination pass — bloom is now owned by PostprocessingPass and composited inline in the postprocessing and illumination shaders, instead of as a separate trailing pass.
- Enable bloom on transparent background — bloom is correctly applied and exported when the background is transparent.
- Background-aware blend — screen on a transparent background (brightens without darkening), PMA-over on an opaque background (keeps the halo visible on bright backdrops), additive on geometry.
- Tighten isBackground — use depth >= 0.99999994 ((2²⁴−1)/2²⁴) so the background test is precise for packed transparent depth.

## Actions


arose (commented):
I like it, thanks!

There are some visual differences for emissive transparent geometry between standard and illumination mode. The emissive transparent geometry is less bright in illumination mode.

Emissive transparent geometry over a transparent background quickly (with higher emissiveness) losses any alpha so that the background is not visible anymore.

Are those intentional, want to change them?

giagitom:
>There are some visual differences for emissive transparent geometry between standard and illumination mode. The emissive transparent geometry is less bright in illumination mode.

>Emissive transparent geometry over a transparent background quickly (with higher emissiveness) losses any alpha so that the background is not visible anymore.

Hi @arose, I missed that, thanks. It should be fixed now.

arose:
standard vs illumination still looks quite different with bloom on transparent objects. Without emissive, I don't see a difference.

Example opacity 0.3, emissive 0.5


image
Standard


image
Illumination


image
Standard


image
Illumination

giagitom:
@arose last commit should fix also this.

arose:
transparent bloom on white background looks odd

image

white bg

image

black bg

giagitom:
Now it should behave uniformly, and also fixed a few issues with environment enabled.

arose:
thanks!

transparent bg still does not look quite right. just adding a bit of emissive to a transparent surface makes it much more opaque

with opacity 0.5

image

emissive 0

image

emissive 0.3

giagitom:
last commit should fix this

giagitom:
it also reproduces the same results when exporting with tranparent bg

giagitom:
@arose I have made other other improvements with last commits.

arose (approved):
this is great, I don't find any issues anymore, thank you @giagitom

PR #1851, opened 2026-06-16
fix ASA for hydrogens: inherit color from residue

papillot:
# Description
This PR fixes an issue wrt ASA coloring for structures that contain hydrogen atoms. Hydrogen atoms are discarded from the computation of the accessible surface area by increasing heavy atoms radius. But currently they are also ignored in the coloration:
Example with PDB 3L2L. Note that the hydrogen atoms appear as white patches.
3L2L-ASA

The change here is to map hydrogen atoms to their residues so that they inherit the color from it. Computation is unchanged.
3L2L-ASA-fix

## Actions


arose (approved):
Very nice, thank you!

PR #1853, opened 2026-06-19
VTK PolyData .vtp file format support

corredD:
# Description
Add VTK PolyData (\`.vtp\`) file format support

  ### Summary
  
  This PR introduces a reader and interactive shape provider for the **VTK XML PolyData** format (\`.vtp\`), enabling direct visualization of polygon mesh surfaces alongside their per-vertex and per-cell scalar
  attributes.

  VTP files are the standard output of VTK-based surface analysis pipelines in structural biology and bioimage analysis. A prominent example is 
  [**SurfaceMorphometrics**](https://github.com/GrotjahnLab/surface_morphometrics), a cryo-ET membrane analysis tool that exports curvature, thickness, and other morphometric scalars directly as \`.vtp\` 
  meshes. This PR makes those outputs immediately viewable in Mol* without any conversion step.

  > **Unit note:** SurfaceMorphometrics (and many VTK-based tools) export coordinates in **nanometres**, whereas Mol* works in **Ångströms**. Use the **Scale** parameter (see below) and set it to \`10\` to 
  bring an nm-space mesh into register with Å-space molecular structures.

  ---
  
  ### What was added

  **Parser** (\`mol-io/reader/vtp/\`)
  - Reads all four VTK 1.0 / 0.1 encoding variants: raw-appended, base64-appended, inline binary, and zlib block-compressed
  - Supports \`UInt32\` and \`UInt64\` header types; \`Float32\`, \`Float64\`, \`Int32\`, \`Int64\`, \`UInt32\`, \`UInt64\` data array types
  - Fan-triangulates quads and n-gons; records \`triangleCellIndex\` so per-cell scalars map correctly after triangulation
  - Parses per-vertex (\`PointData\`) and per-cell (\`CellData\`) scalar arrays with min/max range metadata

  **Shape provider** (\`mol-model-formats/shape/vtp.ts\`)
  - Exposes all scalar arrays (point and cell) as selectable color channels in the UI
  - **Smooth color** mode interpolates per-cell values to vertices for gradient rendering; **flat** mode assigns the cell value to all of its triangles
  - Colormap selector (any Mol* scale), domain min/max controls, uniform-color fallback
  - **Scale** slider (0.01 – 100×) — use \`10\` when loading nm-space VTP files alongside Å-space structures
  - Interior color suppressed by default (clean look for closed surfaces)

  **Transform** (\`mol-plugin-state/transforms/model.ts\`)
  - \`ShapeFromVtp\` state transform wires the parser into the plugin pipeline
  - Supports an optional \`transforms: Mat4[]\` parameter for multi-instance placement

  ---

  ### Multi-channel color mapping

  A single \`.vtp\` file can carry arbitrarily many scalar fields. For a typical SurfaceMorphometrics output these include mean curvature, Gaussian curvature, membrane thickness, and principal curvatures — all 
  immediately switchable from the color-attribute dropdown without reloading:

  | Attribute | Source | Coloring mode |
  |---|---|---|
  | \`mean_curvature\` | CellData | Smooth (gradient) |
  | \`gaussian_curvature\` | CellData | Smooth (gradient) |
  | \`thickness\` | PointData | Per-vertex |
  | Uniform | — | Flat colour |

  ---

  ### Example files

  Sample \`.vtp\` files (cryo-ET membrane surfaces with morphometric scalars, coordinates in nm) are available here:

  📁 [Google Drive — VTP example meshes](https://drive.google.com/drive/folders/1CC8MBWSS6rsMTU7jIo4nh65CvXdwtFz5?usp=sharing)

Screenshot 2026-06-16 at 9 52 43 PM

  ---

  ### Supported VTP variants

  | Feature | Supported |
  |---|---|
  | Raw-appended binary | ✅ |
  | Base64-appended | ✅ |
  | Inline binary (v1.0) | ✅ |
  | zlib block compression | ✅ |
  | UInt32 / UInt64 headers | ✅ |
  | Triangles, quads, n-gons (fan) | ✅ |
  | PointData scalars | ✅ |
  | CellData scalars | ✅ |
  | Multi-component arrays | skipped (vectors/tensors not yet visualised) |
  | BigEndian byte order | not supported |
  
  ---

  ### Testing

  Four Jest unit tests cover the parser end-to-end:
  - Minimal triangle (positions, connectivity, round-trip)
  - Quad fan-triangulation (1 quad → 2 triangles, correct indices)
  - CellData scalar parsing (range min/max, value accuracy)
  - \`triangleCellIndex\` correctness for quads with per-cell data

## Actions


src/mol-model-formats/shape/vtp.ts
```diff
@@ -0,0 +1,420 @@
...
+    const attrOptions = buildAttrOptions(vtpFile);
+    const defKey = defaultAttrKey(vtpFile);
+    const [defMin, defMax] = defaultDomain(vtpFile, defKey);
+    const defStats = vtpFile ? computeAttrStatsText(vtpFile, defKey) : '';
+
+    return {
+        ...Mesh.Params,
+        doubleSided: { ...Mesh.Params.doubleSided, defaultValue: true },
+        interior: { ...Mesh.Params.interior, defaultValue: { ...Mesh.Params.interior.defaultValue, colorStrength: 0 } },
+        attribute: PD.Select(defKey, attrOptions, { label: 'Color Attribute' }),
```
arose: use the following approach to add statistic info about the chosen attribute

        attribute: PD.Select(defKey, attrOptions, { label: 'Color Attribute', help: (x) => {
            console.log(x);
            return { description: 'add stats here' };
        } }),

src/mol-plugin-state/transforms/representation.ts
```diff
@@ -1191,11 +1191,29 @@ const ShapeRepresentation3D = PluginStateTransform.BuiltIn({
             return new SO.Shape.Representation3D({ repr, sourceData: a.data }, { label: a.data.label });
         });
     },
-    update({ a, b, newParams }) {
+    update({ a, b, newParams, spine }, globalContext: PluginContext) {
+        // Capture ref synchronously while spine.current is still set to this cell
```
arose: let's not do this, I added an comment for an alternative approach to the vtp shape params

src/mol-model-formats/shape/vtp.ts
```diff
@@ -0,0 +1,420 @@
...
+export function shapeFromVtp(source: VtpFile, params?: { transforms?: Mat4[] }) {
+    return Task.create<ShapeProvider<VtpData, Mesh, VtpShapeParams>>('Shape Provider', async () => {
+        const getter = makeShapeGetter();
+        const provider: ShapeProvider<VtpData, Mesh, VtpShapeParams> & { onParamsUpdate?: (props: PD.Values<VtpShapeParams>) => Record<string, unknown> | null } = {
+            label: 'VTP Mesh',
+            data: { source, transforms: params?.transforms },
+            params: createVtpShapeParams(source),
+            getShape: getter.getShape,
+            geometryUtils: Mesh.Utils,
+            onParamsUpdate(props: PD.Values<VtpShapeParams>) {
```
arose: handle via help provider in param

src/mol-model-formats/shape/vtp.ts
```diff
@@ -0,0 +1,420 @@
...
+    const transforms = baseTransforms.map(t => Mat4.mul(Mat4(), t, scaleT));
+    return Shape.create(
+        'vtp-mesh', vtpData.source, mesh,
+        colorFn,
+        () => 1,
+        (gid: number) => {
+            if (gid < 0) return '';
+            return `Triangle ${Math.floor(gid / 3)}`;
+        },
+        transforms
```
arose: provide number of groups, i.e., largest groupId, as another argument after transform, otherwise it will be dynamically derived which can be quite noticeable for larger meshes

src/mol-io/reader/vtp/parser.ts
```diff
@@ -0,0 +1,673 @@
...
+    return -1;
+}
+
+function extractAppendedInfo(data: Uint8Array): AppendedInfo {
+    const tagPos = findBytesPosition(data, '<AppendedData');
+    if (tagPos === -1) throw new Error('AppendedData section not found');
+
+    let pos = tagPos;
+    while (pos < data.length && data[pos] !== 0x3E) pos++;
+    // Read the encoding attribute without regex.
```
arose: I would remove any regex mentions, no need to note a previous approach

src/mol-io/reader/vtp/parser.ts
```diff
@@ -0,0 +1,673 @@
...
+//
+// VTK base64 appended: each DataArray's binary block is split into two base64-encoded chunks:
+//   1. The compressed-block header  (nblocks, blockSize, lastSize, compSize[i])
+//   2. All compressed blocks concatenated
+//
+// The DataArray XML `offset` attribute is a byte offset in the base64 ASCII stream after '_'.
+// Since base64 is ASCII (1 char == 1 byte), the character offset equals the byte offset.
+// We decode base64 directly from the source without going through atob or string copies.
+
+const B64_DECODE = new Int8Array(256).fill(-1);
```
arose: please move the decode helpers to mol-io/reader/common/decode.ts and add spec test

src/mol-io/reader/vtp/parser.ts
```diff
@@ -0,0 +1,673 @@
...
+        if (a < 0 || b < 0) break;
+        if (di < nOut) out[di++] = (a << 2) | (b >> 4);
+        if (c >= 0 && di < nOut) out[di++] = ((b & 0xF) << 4) | (c >> 2);
+        if (d >= 0 && di < nOut) out[di++] = ((c & 0x3) << 6) | d;
+    }
+    return out;
+}
+
+// Decode nOut bytes of base64 from a string with start index (inline path — avoids atob).
+function decodeB64Str(src: string, srcStart: number, nOut: number): Uint8Array {
```
arose: use StringLike as input type, here and in any other place where the str could be large

src/mol-io/reader/vtp/parser.ts
```diff
@@ -0,0 +1,673 @@
...
+    if (hdr.compressor && hdr.compressor !== 'vtkZLibDataCompressor') {
+        throw new Error(`Unsupported VTP compressor: "${hdr.compressor}". Only vtkZLibDataCompressor is supported.`);
+    }
+    if (hdr.headerType !== 'UInt32' && hdr.headerType !== 'UInt64') {
+        throw new Error(`Unsupported VTP header_type: "${hdr.headerType}". Only UInt32 and UInt64 are supported.`);
+    }
+    if (hdr.nVerts + hdr.nLines + hdr.nStrips > 0) {
+        console.warn(`VTP: file contains ${hdr.nVerts} verts, ${hdr.nLines} lines, ${hdr.nStrips} strips — only Polys are rendered.`);
+    }
+    const headerType = hdr.headerType as 'UInt32' | 'UInt64';
```
arose: why cast?

src/mol-io/reader/vtp/parser.ts
```diff
@@ -0,0 +1,673 @@
...
+}
+
+function decodeInt64AsInt32(raw: Uint8Array, count: number): Int32Array {
+    const dv = new DataView(raw.buffer, raw.byteOffset, raw.byteLength);
+    const out = new Int32Array(count);
+    for (let i = 0; i < count; i++) out[i] = dv.getUint32(i * 8, true);
+    return out;
+}
+
+function decodeToFloat64(raw: Uint8Array, type: string, count: number): Float64Array {
```
arose: why always decode to float64 and not a more specific type?

arose (changes requested):
first review pass, looking good overall, thanks

src/mol-model-formats/shape/vtp.ts
corredD: make total sense!

src/mol-io/reader/vtp/parser.ts
corredD: its for downstream coloring, would you recommand to use TypedArray ?

arose: Maybe NumberArray?

src/mol-model-formats/shape/vtp.ts
```diff
@@ -0,0 +1,417 @@
...
+    const attrOptions = buildAttrOptions(vtpFile);
+    const defKey = defaultAttrKey(vtpFile);
+    const [defMin, defMax] = defaultDomain(vtpFile, defKey);
+    const defStats = vtpFile ? computeAttrStatsText(vtpFile, defKey) : '';
+
+    return {
+        ...Mesh.Params,
+        doubleSided: { ...Mesh.Params.doubleSided, defaultValue: true },
+        interior: { ...Mesh.Params.interior, defaultValue: { ...Mesh.Params.interior.defaultValue, colorStrength: 0 } },
+        attribute: PD.Select(defKey, attrOptions, { label: 'Color Attribute', help: () => {
```
arose: how about something like this

```ts
colorTheme: PD.MappedStatic('attribute', {
    attribute: PD.Group({
        name: PD.Select(defKey, attrOptions, { help: () => {
            return { description: getStats ? getStats() : defStats };
        } }),
        colors: PD.Select('viridis' as ColorListName, ColorListOptionsScale),
        domain: PD.MappedStatic('auto', {
            custom: PD.Interval([-1, 1], { step: 0.001 }),
            auto: PD.Group({
                symmetric: PD.Boolean(false, { description: 'If true the automatic range is determined as [-|max|, |max|].' })
            })
        })
    }),
    uniform: PD.Group({
        color: PD.Color(ColorNames.grey, { label: 'Uniform Color' })
    })
}),
```

src/mol-model-formats/shape/vtp.ts
```diff
@@ -0,0 +1,417 @@
...
+        uniformColor: PD.Color(ColorNames.grey, { label: 'Uniform Color' }),
+        scale: PD.Numeric(1, { min: 0.01, max: 100, step: 0.01 }, { label: 'Scale', description: 'Uniform scale factor applied to the mesh.' }),
+    };
+}
+
+export const VtpShapeParams = createVtpShapeParams();
+export type VtpShapeParams = typeof VtpShapeParams;
+
+// --- Mesh building ---
+// Vertices are intentionally NOT shared (3 unique entries per triangle).
```
arose: is there still flat face coloring? can this be simplified?

src/mol-io/reader/vtp/parser.ts
corredD: I tried with TypedArray.

src/mol-model-formats/shape/vtp.ts
corredD: much better like this !

src/mol-model-formats/shape/vtp.ts
corredD: no flat color, I removed the vertex duplication.

src/mol-model-formats/shape/vtp.ts
```diff
@@ -0,0 +1,409 @@
...
+        sum[v2] += val; count[v2]++;
+    }
+    const smoothed = new Float64Array(numberOfPoints);
+    for (let v = 0; v < numberOfPoints; v++) {
+        smoothed[v] = count[v] > 0 ? sum[v] / count[v] : 0;
+    }
+    return smoothed;
+}
+
+// Average multi-component cell vectors to vertices, then return the magnitude of each vertex vector.
```
arose: please use multi-line comments /** */ for comments for functions, classes, members, ... that way they are shown as docs in editors

src/mol-io/reader/common/decode.ts
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * @author Ludovic Autin <autin@scripps.edu>
+ */
+
+import { StringLike } from '../../common/string-like';
+import { TypedArray } from '../../../mol-util/type-helpers';
+
+// --- Base64 ---
+
+export const B64_DECODE = new Int8Array(256).fill(-1);
```
arose, line 11: I meant to put just decodeB64Bytes and decodeB64Str here, the other decode herlper are too specific

src/mol-io/reader/vtp/schema.ts
```diff
@@ -0,0 +1,52 @@
+/**
+ * Copyright (c) 2026 mol* contributors, licensed under MIT, See LICENSE file for more info.
+ *
+ * @author Ludovic Autin <autin@scripps.edu>
+ */
+
+export interface VtpDataArrayDescriptor {
+    readonly name: string;
+    /** VTK type string: "Float32", "Float64", "Int32", "Int64", "UInt32", etc. */
+    readonly type: string;
```
arose: can this be a union type?

src/mol-io/reader/vtp/schema.ts
```diff
@@ -0,0 +1,52 @@
...
+    readonly numberOfComponents: number;
+    /** Encoding format of this DataArray element. */
+    readonly format: 'ascii' | 'binary' | 'appended';
+    /** Byte/char offset into the appended data section (after the `_` marker). -1 for inline/ascii. */
+    readonly offset: number;
+    readonly rangeMin: number;
```
arose: allow undefined

src/mol-io/reader/vtp/schema.ts
```diff
@@ -0,0 +1,52 @@
...
+    readonly rangeMax: number;
```
arose: allow undefined

src/mol-io/reader/vtp/schema.ts
```diff
@@ -0,0 +1,52 @@
...
+    /** True if RangeMin/RangeMax were explicitly present in the XML (absent in many writers). */
+    readonly hasRange: boolean;
```
arose: remove in favor of undefined of min/max

src/mol-io/reader/vtp/schema.ts
```diff
@@ -0,0 +1,52 @@
+/**
+ * Copyright (c) 2026 mol* contributors, licensed under MIT, See LICENSE file for more info.
+ *
+ * @author Ludovic Autin <autin@scripps.edu>
+ */
+
+export interface VtpDataArrayDescriptor {
```
arose: apart from the rangeMin/Max and type (anything else?) this should be private to the parser. don't want to keep refs that prevent GC

src/mol-io/reader/vtp/schema.ts
```diff
@@ -0,0 +1,52 @@
...
+import { TypedArray } from '../../../mol-util/type-helpers';
+
+export interface VtpScalarArray {
+    readonly desc: VtpDataArrayDescriptor;
+    /**
+     * Flat decoded values in the native VTK type. For scalar arrays (numberOfComponents=1) length = nElems.
+     * For multi-component arrays length = nElems * numberOfComponents.
+     * Int64/UInt64 are represented as Float64Array.
+     */
+    readonly values: TypedArray;
```
arose: lets wrap this in a Column (can be backed by the same array for now, but the Column interface would allow us to defer full parsing if we need to without breaking)

arose (changes requested):
added some more comments and also, please add a changelog entry

src/mol-io/reader/vtp/schema.ts
corredD: yes

src/mol-io/reader/vtp/schema.ts
corredD: moved to Column, and replace TypeArray by ArrayLike<number>

src/mol-io/reader/vtp/schema.ts
corredD: only numberOfComponents, everything else was put private for the parser.

corredD:
Thanks @arose your comments are so usefull!

arose (approved):
thanks, looking good

PR #1854, opened 2026-06-19
Fix size-only updates in updateRepresentationsTheme

Himanshuraj677:
Fix size-only updates in `updateRepresentationsTheme()`.

The size theme update path incorrectly depended on `params.color`, causing size-only updates to be skipped for some representations. This change uses `params.size` instead.

Verified locally that calling `updateRepresentationsTheme(..., { size: 'physical' })` correctly updates all representations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file, following t
 Note that since we don't clearly distinguish between a public and private interfaces there will be changes in non-major versions that are potentially breaking. If we make breaking changes to less used interfaces we will highlight it in here.
 
 ## [Unreleased]
+- Fix size-only representation theme updates in `updateRepresentationsTheme`.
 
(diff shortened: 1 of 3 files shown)


arose (approved):
Nice find and thank you for the fix!

Please consider doing

- Added description of changes to the [Unreleased] section of CHANGELOG.md
- Updated headers of modified files
- Added my name to package.json's contributors

Himanshuraj677:
> Nice find and thank you for the fix!
> 
> Please consider doing
> 
> * Added description of changes to the [Unreleased] section of CHANGELOG.md
> * Updated headers of modified files
> * Added my name to package.json's contributors

Thanks for the review and approval!

I've added an entry to the Unreleased section of CHANGELOG.md, updated the file header, and added myself to the contributors list in package.json.

Please let me know if anything else needs to be adjusted.

arose (approved):

Issue #1361, opened 2024-11-25
`isConjugated` is `false` for Arginine terminal Ns in PDBs


pechersky:
https://github.com/molstar/molstar/blob/c53b651472c63adaf1115753375bdea6e5b56045/src/mol-model-props/computed/chemistry/valence-model.ts#L40-L67

This affects H-bond calculation, because the ideal donor angle is then based off of tetrahedral (109) instead of the correct trigonal (120) geometry. The `isConjugated` calculation gives `false` because in PDBs, there are no bond orders, so there are no bonds with `bond.order > 1`.

arose:
We assign bond order for known cases, see src\mol-model\structure\model\properties\atomic\bonds.ts

There is a comment to interpret connect records (if given in a non-standard way) in PDB files to assign bond orders, see src\mol-model-formats\structure\pdb\conect.ts

Maybe we can also do a distance check in `isConjugated`? Something base on section 9 in https://www.daylight.com/meetings/mug01/Sayle/m4xbondage.html

> unsatisfied sp2-hybridized atom by checking each neighboring atom for a terminal oxygen, and when found testing the bond length against the table of bond lengths

Feel free to do a PR!

pechersky:
Not sure what the best way to overcome this would be. Somehow in the 3D viewer, the Arg does show a double bond to one of the terminal Ns in the arginine -- so somewhere bond orders are inferred. Is it possible to run the `isConjugated` after that?

Additionally, in our case, we would satisfy the `donHAngle` check -- given that the Hs exist, why should we check the `X -- N` angle at all? Maybe we only check `donAngle` within ideal angle if Hs don't exist?

Short term, we're going to hardcode `NH1` and `NH2` to return `isConjugated` true just so that they get trigonal geom.

papillot:
> Additionally, in our case, we would satisfy the `donHAngle` check -- given that the Hs exist, why should we check the `X -- N` angle at all? Maybe we only check `donAngle` within ideal angle if Hs don't exist?
>

There are cases where the H atom position is not compatible with a putative H-bond. It is necessary to combine both data (i.e. "Are there donors and acceptors at suitable positions?", "When Hydrogens are present, do they contradict the H-bond perception?" ) when they are present to discard false positive.

pechersky:
In my case, I have "Hs are present, they should be the only determinants of angles". I read your statement as:
```
don and acceptors look okay
oh wait, Hs aren't placed in a way that is acceptable
```
[...]

Issue #626, opened 2022-11-18
Difference in ligand interactions between pdb and cif


jpattle:
Hello,

I have observed that there is a difference between the interaction features assigned to atoms in a ligand when loading a pdb verses loading a cif file.

As an example, [1udt](https://www.ebi.ac.uk/pdbe/entry/pdb/1udt) loads differently into Mosltar if you download the '[PDB file](https://www.ebi.ac.uk/pdbe/entry-files/download/pdb1udt.ent)' verses downloading the '[updated mmCIF file](https://www.ebi.ac.uk/pdbe/entry/pdb/1udt#:~:text=Archive%20mmCIF%20file-,Updated%20mmCIF%20file,-PDB%20file%0APDB)' from the ebi site for that structure.

The image below shows the carbonyl oxygen for the pdb file of 1udt as having both the Hydrogen Acceptor and Hydrogen Donor interaction features.

![Screenshot 2022-11-18 at 13 41 28](https://user-images.githubusercontent.com/84507154/202718854-2f32f7c6-1965-4bc8-978e-a0edd626a27d.png)

The image below shows the carbonyl oxygen for the cif file of 1udt as having only the Hydrogen Acceptor interaction feature (which is correct).

![Screenshot 2022-11-18 at 13 42 35](https://user-images.githubusercontent.com/84507154/202719015-32ad2ac5-1ba6-4939-98dc-2a3cc2d179a8.png)

There are also differences for the aromatic nitrogen in the ligand, as the pdb file shows them as having a positive charge, where as the cif does not.

We see similar issues when using proprietary pdb files.

Is this an issue in the parsing of pdb files?

Thanks,

Jason

arose:
I suspect that is due to different bond orders of the marked oxygen-carbon bond. You need to provide bond orders. This can be done in mmcif file using the `_chem_comp_bond` category. There is nothing officially for bond orders in the pdb format. Sometimes a  convention (Interpret CONECT records where the 'to' atom is given multiple times as double/triple bonds, e.g. CONECT 1529 1528 1528 is a double bond.) is used but that is not implemented in molstar, however, a PR to add that would be OK. Finally, you may implement something in molstar to derive bond orders, see #36.

arose:
@jpattle does the above comment explain what you observe?

jpattle:
[...]

Issue #36, opened 2020-05-10
Bond order


Valdes-Tresanco-MS:
Hello
The bonds order is not displayed correctly in non-polymers representations. It only shows correctly if it appears in the mmcif file (at least in pdb format it doesn't show correctly)

arose:
That is expected, we only show bond orders if it is included in the file. We do not attempt to calculate/derive the bond order from geometry/element/connectivity. If you know a standard way to do so we can consider implementing it.

Valdes-Tresanco-MS:
I think this method would be good. OpenBabel uses an approach based on this description. In addition to the exquisite explanation, it has examples of the possible problems that you can find in the PDB.
It also contains part of the feature you want to implement in #1

arose:
> I think this method would be good.

what method are you referring?

Valdes-Tresanco-MS:
Oh i'm so sorry i forgot the link
Here it is:
https://www.daylight.com/meetings/mug01/Sayle/m4xbondage.html

arose:
> https://www.daylight.com/meetings/mug01/Sayle/m4xbondage.html

Thank you. The relevant part seems to be in section 9 of that document. This can in principle be done. After inter and intra bonds are calculated (https://github.com/molstar/molstar/tree/master/src/mol-model/structure/structure/unit/bonds) there would need to be a second pass (probably only for non-polymer molecules to stay performant) that assigns the bond order (if it was not given otherwise). There is code for functional groups (https://github.com/molstar/molstar/tree/master/src/mol-model-props/computed/chemistry) and aromatic rings (https://github.com/molstar/molstar/blob/master/src/mol-model/structure/structure/unit/rings.ts) available. That said, it will require quite some work integrating it all and making sure it is sufficiently performant. Any contributions are very welcome.

Valdes-Tresanco-MS:
Hi @arose. 
The truth is that the project seems spectacular to me and I would like to be able to contribute, but unfortunately I have no experience in TypeScript programming. I will be pending and if I can contribute I will do so without hesitation. Right now I am learning a bit about TypeScript (only by this) so that in the future I can actively contribute and implement some of my own features. I'm sorry, at the moment I can't be of more help :(

PR #1855, opened 2026-06-21
Bonds orders perception

papillot:
# Description
This PR is a WIP. POC for the Mol* implementation of Roger Sayle's algorithm for bond perception https://www.daylight.com/meetings/mug01/Sayle/m4xbondage.html which is mentioned in different bug reports #36 #626 #1361

image

The PR is not ready yet: the code has been partly vibe coded: it requires simplifications and thorough review. Also, the validation has only started and the implementation is incomplete (inter-unit bonds are not considered as of now, behavior with ).

While working on these tasks, I would like advices about the integration in Mol* codebase:
- Parameter for bond order perception: for now it's optional, enabled by default but only kicking when there is no known bond order connectivity (not conn_bond records, or not a canonical-templated residue for which there is a table)
- tests: unit tests have been added (they will be reviewed and cleaned), but more importantly validation test. These rely on a manifest file (committed) which lists ligands codes and PDB codes to download legacy PDB and extract the ligand coordinates only as well as the CCD cif as the source of truth (not committed). As of now, all this code is buried within the src/mol-model/**/bonds source tree. Shall it move under src/tests?
- integration with the other chemistry aware code: this code assigns geometries to atom, computes aromaticity, matches some functional groups with charges, etc.. It should simplify heuristics and remove some assumptions that are made in the ValenceModel. For now, those assignments are lost after completion.

## Actions


arose:
this is a tricky one

validation tests should be separate from the spec tests

unless we move chemistry perception into structure creation (which would be a very big change I am not sure about) I feel this should better be calculate property like the other chemistry perception stuff and you would need to explicitly opt-in to use it instead of the given bond-orders in unit.bonds. For sure, more cumbersome but much less obtrusive.

papillot:
The feature is "almost" ready. Implementation is complete, the vibe-coded code has been reviewed and mostly rewritten. It has been benchmarked against Labute 2005 dataset and it fares well (if we ignore hemes which I don't think are part of the dataset - it only cites the PDB codes, not the chem comp codes - there are 11 failures out of the 177 structures, which is the same failure rate as Labute 2005 and slightly worse than the 9 reported by Zhang 2011), most failures being related to poor structures. 

Things that remain:
- handling alternate locations: at the moment, these would break the indices.
- inter-unit bonds are mostly ignored for now. This is work for future
- some cases to better handle
- unit tests have not been reviewed yet. They were created by the agent initially, and some have become pointless. I guess the altloc guard will be an interesting addition.

Questions regarding the architecture:
- benchmarking code has been moved to the CLI section (with an npm script handle added to package.json), but I wonder if it should be there at all or part of another project: I don't see this code as being of general use?
- computation is encapsulated in a custom property. This had required some changes throughout the code base to expose the bond order values (for example to the representation related code). Not sure if there is a better design?

papillot:
PR should be ready now. Since last comments, handling of alternates has been added and some improvements added to the geometric checks (slight improvement measured in the benchmarks).

src/mol-model-formats/structure/pdb/atom-site.ts
```diff
@@ -273,7 +275,7 @@ export function getAtomSite(sites: AtomSiteTemplate, labelAsymIdHelper: LabelAsy
 
         labelAsymIds[i] = labelAsymIdHelper.get(i);
 
-        if (atomIdCounts.has(atomId)) {
+        if (atomIdCounts.has(atomId) && altloc === '') {
```
dsehnal, line 278: Nice fix

src/mol-model-props/computed/chemistry/bond-orders/chemcomp-cache.ts
```diff
@@ -0,0 +1,94 @@
...
+    return pattern;
+}
+
+export function applyCachedChemCompPattern(unit: Unit.Atomic, bonds: IntraUnitBonds, start: UnitIndex, end: UnitIndex, cachePrefix = ''): boolean {
+    const model = unit.model;
+    const cache = getModelCache(model);
+    const { label_comp_id, label_atom_id } = model.atomicHierarchy.atoms;
+    const compId = label_comp_id.value(unit.elements[start]);
+    const compHash = cachePrefix + getCompHash(unit, start, end, compId);
+        const pattern = cache.get(compHash);
```
dsehnal: formatting

src/mol-model/structure/model/types.ts
```diff
@@ -635,7 +636,8 @@ export namespace BondType {
         HydrogenBond = 0x4,
         Disulfide = 0x8,
         Aromatic = 0x10,
-        Computed = 0x20
+        Computed = 0x20,
+        AromaticHuckel = 0x40 // derived from bond-order perception
```
dsehnal: These values are not serialised anywhere. I think it makes more sense to put the Huckel version next to Aromatic.

src/mol-model/structure/structure/unit/bonds/common.ts
```diff
@@ -89,7 +89,7 @@ export const MetalsSet = (function () {
     return set;
 })();
 
-function pair(a: number, b: number) {
+export function pair(a: number, b: number) {
```
dsehnal: This is a duplicate of `sortedCantorPairing` in `hash-functions.ts`, can use that everywhere instead.

src/mol-model/structure/structure/unit/rings/compute.ts
```diff
@@ -146,6 +147,7 @@ function processResidue(state: State, start: number, end: number) {
             resetDepth(state);
             mark = findRings(state, i, mark);
         }
+        recoverLargeCycles(state);
```
dsehnal: I am not a big fan of this as the function potentially does a lot of memory allocations.

What is the reason for adding this here? Could this call be conditional?

src/mol-model/structure/structure/unit/rings/compute.ts
```diff
@@ -295,6 +297,243 @@ function findRings(state: State, from: number, mark: number) {
...
+function recoverLargeCycles(state: State) {
+    const count = state.endVertex - state.startVertex;
+    // an undetected cycle is necessarily larger than 2 * MaxDepth atoms, so a component must have at
+    // least that many atoms; skip smaller residues cheaply before doing any graph analysis.
+    if (count <= 2 * Constants.MaxDepth) return;
+
+    const adj = buildResidueAdjacency(state, count);
+
+    // connected components of the covalent graph (iterative flood fill)
+    const comp = new Int32Array(count).fill(-1);
```
dsehnal: Can this be allocated lazily as past of `State`? You only need to keep an array with the largest encountered count.

Similar for the other allocations.

src/mol-model/structure/structure/unit/rings/compute.ts
```diff
@@ -295,6 +297,243 @@ function findRings(state: State, from: number, mark: number) {
...
+
+/**
+ * Horton minimum cycle basis on one connected component. Seeds a GF(2) edge-incidence basis with the
+ * rings the BFS already found, then adds the shortest Horton candidate cycles that increase the rank
+ * until it reaches `mu` (the cyclomatic number). The candidates added beyond the seeds are the missing
+ * minimal cycles (e.g. the porphyrin 16-macrocycle); they are appended to `state.currentRings`.
+ */
+function hortonRecover(state: State, adj: number[][], verts: number[], foundRings: SortedArray<StructureElement.UnitIndex>[], mu: number) {
+    // reduce to the 2-core: iteratively drop degree <= 1 vertices (side chains carry no cycles)
+    const deg = new Map<number, number>();
```
dsehnal: This could also be a map on state and you just .clear it here. This goes for all the state. can be created lazily with a `ensureHortonState(state, size)` function.

src/mol-model/structure/structure/unit/rings/compute.ts
```diff
@@ -295,6 +297,243 @@ function findRings(state: State, from: number, mark: number) {
     return mark + 1;
 }
 
+/** Residue-local covalent adjacency (both endpoints in `[startVertex, endVertex)`); indices are `atom - startVertex`. */
+function buildResidueAdjacency(state: State, count: number): number[][] {
+    const { bonds, startVertex, endVertex } = state;
+    const { b: neighbor, edgeProps: { flags }, offset } = bonds;
+    const adj: number[][] = new Array(count);
```
dsehnal: no need to always allocate this for each adjacency. can keep a single array and index it with with `i * count + j`

dsehnal (commented):
That's a lot of work, thanks for going through this and contributing it.

What effect does this have on performance as this seems like there is a lot more going on.

src/mol-model/structure/structure/unit/rings/compute.ts
papillot: Woops... This was an exploration, made at some stage. I thought I had killed it. There is an internal conditional execution based on Euler relation. But I'll remove it anyways.

I added that when I was trying to solve perception in hemes. In that case there are 4 aromatic rings of size 5, but also a macrocyclic aromatic ring of size 16. Going for larger ring sizes allowed to perceive the aromaticity, but the final kekulization was not looking good (due to symmetry: it's more a question of convention and a template based standardization may be the best to address this in a better future). All in all, it adds code complexity without fixing anything.

src/mol-model/structure/structure/unit/rings/compute.ts
papillot: Sorry for this. Code removed

src/mol-model/structure/structure/unit/rings/compute.ts
papillot: Sorry for this. Code removed

papillot:
> That's a lot of work, thanks for going through this and contributing it.
> 
Thanks! It's interesting on some scenarios where users have PDB files or mmCIF that lack connectivity (e.g. from computational methods). The results are better than third party libraries like obabel. This brings us one step closer to some Pymol users expectations!

In the future, this could be used to fix connectivity in structures that have inter-unit bonds (e.g. with some covalent ligands). It should also improve other features such as the interactions (better nitrogen hybridization perception)

> What effect does this have on performance as this seems like there is a lot more going on.

I haven't benchmarked it against baseline, something I should do. In principle, the code does not run at all with an mmCIF file that contains connectivity or an SDF file (connectivity always known).
For a PDB file, canonical residues (e.g. ALA) are skipped (Mol* has templates based on atom names for those) unless the `force` option is used. For the other residues, only the first occurrence is checked, then a pattern is extracted and applied to every other copy.

The test suite of 190 residues from PDB files runs under 1 second on my computer, and this accounts also for reading the files, building the structure and comparing with the baseline JSON. 

But yeah, we need the numbers and in particular see if there are performance bottlenecks.

src/mol-model/structure/structure/unit/rings/compute.ts
dsehnal: That ring detection algorithm was written specifically to detect primitive rings, so if we kept this, would like to have this separate anyway.

papillot:
Re. benchmarking,
I've tried with structure 6FUX, which contains 3 ligands like units. The `perceiveBondOrders` code runs in 11ms, 40% of it being primitive rings detection which I think would have ran anyways to display the aromatic bond cylinders. Regarding optimizations from the Performance analysis, it's noticeable that eachBondedAtom/eachIntraBondedAtom account for nearly 10% of the time. As you've already noticed in https://github.com/molstar/molstar/pull/797/changes/BASE..f3a5369690537844f87aa8b96845004e2993f4d4#r1186795226 it's closures in loops, which probably impact perf. I'll try the top level fn refactor suggested.

I've noticed in the flame graph an issue with the way the BondProvider is attached: it forces the computations of all the bond orders (and the bond graph) on the entire structure, instead of a lazy initialization which is the case right now. This adds an extra 10ms, mostly for nothing.

arose (commented):
I need more time to review this, so just a few comments/questions.

- My original suggestion was to have the calculated bond orders be really separate from IntraUnitBonds so you would have a completely new property. I understand that this is very awkward and intrusive when we actually want to use it. So we need to rethink the integration. Either have it as an integral part of structure/unit (if performant) or have a state tree transform from Structure to Structure.
- Are you actually mutating the data in the unit? Or a fresh copy? With the state tree all data needs to be immutable.
- The validation could go into a validation folder similar to the perf folder which is not part of the npm package.

src/mol-model-formats/structure/pdb/atom-site.ts
arose, line 278: please (also) do this as a separate PR

src/mol-model/structure/model/types.ts
```diff
@@ -635,7 +636,8 @@ export namespace BondType {
         HydrogenBond = 0x4,
         Disulfide = 0x8,
         Aromatic = 0x10,
-        Computed = 0x20
+        AromaticHuckel = 0x20, // derived from bond-order perception
```
arose, line 639: do we really need a new aromatic type? we don't have new types for the bond-orders

arose (changes requested):

src/mol-model-formats/structure/pdb/atom-site.ts
arose: in general it would be great to have a PR with the unrelated changes, I saw a few tweaks here and there and moving them would reduce the noise

papillot: Will do.

I'll try to chase down those unrelated changes. There was one request from David above to reuse CantorPairing fn, otherwise it should be only new exports related to this functionality.

papillot:
> I need more time to review this, so just a few comments/questions.
> 
> * My original suggestion was to have the calculated bond orders be really separate from IntraUnitBonds so you would have a completely new property. I understand that this is very awkward and intrusive when we actually want to use it. So we need to rethink the integration. Either have it as an integral part of structure/unit (if performant) or have a state tree transform from Structure to Structure.

My apologies for misunderstanding your previous suggestion. In the current PR, the computation has been moved to a provider registering a custom property. The original `bond` property of each unit *is not* mutated, and still exists as is. Each place in the code where the bond order is used, calls an accessor fn (`getUnitBondProp`) that checks if a Bond order provider is available and go through it to get the bond order and/or the bond flag. That part is indeed awkward and intrusive, because it means that every place in the code where bond orders are required needs to "be aware" of this quirk.
It seems that having a transform which returns a new Structure object, would make the rest of the code completely agnostic to the bond order code.

Regarding performance, in the last iteration, when displaying the default preset with the 6FUX structure mentioned above (3 unique ligands), the call to `perceiveBondOrders` accounts for 6ms, which includes 2.5ms of ring finding (I think this executes anyways when the ball+stick representation is used, for aromatic rings - if I am right the perception itself executes ). This does not execute in the case of mmCIF files with chem_comp_bond provided, or SDF files, etc...

image

If you have suggestions about other performance checks, I am willing to investigate.

> * Are you actually mutating the data in the unit? Or a fresh copy? With the state tree all data needs to be immutable.

See above, the unit.bonds property is not mutated.

> * The validation could go into a validation folder similar to the perf folder which is not part of the npm package.

Will do!

Also, I understand that this is a lot of code to review. If there is a way I could make this easier, I am open to suggestions. I will already make the suggested splits (alt locs PR).

src/mol-model/structure/model/types.ts
papillot: I think so, but this may be a bit pedantic. RDKit has 4 models of aromaticity, including one based on Huckel's rule. 

Currently, in Mol*, the aromatic type is either inherited from the mmCIF (I don't know the definition there, but there are instances where a single bond between 2 aromatic rings is classified as aromatic), or it is computed in the rings routine, based on elements and planarity.

With the bonds perception, the requirements on planarity are a bit less severe, but conditions are enforced on the 4n+2 rule. Ideally, when the flag AromaticHuckel is set, we should be able to trust the kekulization and make better assumptions on the Nitrogens protonation (this is not completely true now, because there is no mechanism to propagate the charge perception on atoms - maybe another argument in favor of a transform based architecture).

src/mol-model-props/computed/chemistry/bond-orders.ts
```diff
@@ -0,0 +1,155 @@
...
+import { State } from './bond-orders/common';
+import { applyFunctionalGroups } from './bond-orders/functional-groups-connectivity';
+import { perceiveAromaticRings } from './bond-orders/huckel';
+import { applyCachedChemCompPattern, cacheChemCompPattern } from './bond-orders/chemcomp-cache';
+import { assignHybridization } from './bond-orders/hybridization';
+import { assignBondOrders } from './bond-orders/kekule';
+
+/**
+ * Perceive and assign bond orders in place on `bonds` for residues of `unit` whose
+ * orders are not otherwise known. Mutates the `order`/`flags` edge properties.
```
arose, line 68: the 'mutates' is what tripped me, but you are passing a copy of bonds, so its ok

arose:
> > I need more time to review this, so just a few comments/questions.
> > 
> > * My original suggestion was to have the calculated bond orders be really separate from IntraUnitBonds so you would have a completely new property. I understand that this is very awkward and intrusive when we actually want to use it. So we need to rethink the integration. Either have it as an integral part of structure/unit (if performant) or have a state tree transform from Structure to Structure.
> 
> My apologies for misunderstanding your previous suggestion. In the current PR, the computation has been moved to a provider registering a custom property. 

No worries, you understood me right. Having thought about this for a bit more I think a Structure -> Structure transform (probably as decorator) is what we should use. Bond-order are not really a property but an integral part and too much depends on them.

I would also suggest moving it to an extension. This is quite a bit beyond visualization and we need to see how far we want to go. At some point it might be better to integrate rdkit.

Some care is needed to run it before, e.g., builder.insertStructureProperties, so they have access to the calculated bond-orders. Maybe we need to establish a new pattern to run structure prepare/fix/optimize calculations. Things like add hydrogen atoms come to mind. Anyway, for now you can write a custom TrajectoryHierarchyPresetProvider.


> If you have suggestions about other performance checks, I am willing to investigate.
> 

Many MD sim data pairs don't have bond-orders so the calculation gets triggered. I tried a PSF/DCD pair with 200k atoms and calculation takes ~500ms per frame. Might be due to water molecules? Also would be good to somehow reuse among frames.

> > * Are you actually mutating the data in the unit? Or a fresh copy? With the state tree all data needs to be immutable.
> 
> See above, the unit.bonds property is not mutated.
> 

Thanks, it was only the description of an internal method that made me think so.

src/mol-model/structure/model/types.ts
arose: Could it be an implementation detail of the bond-order calculation? Does the rest of molstar need to know?

papillot: I guess that if we lean on the Structure->Structure transform implementation, we can guarantee that the aromaticity property read by the rest of Mol*, comes from a single, well defined, origin. Will do.

papillot:
> At some point it might be better to integrate rdkit.

Just for clarification, I had tried to run RDKit bond order perception in another project and it only works when all hydrogen atoms are provided and total formal charge is known (similar requirements as antechamber from ambertools). https://greglandrum.github.io/rdkit-blog/posts/2022-12-18-introducing-rdDetermineBonds.html

OpenBabel does provide a function that works from PDB files, but the results are often rather disappointing.

papillot:
@arose before starting with a new integration could you validate the following?

The functionality could be implemented as a Structure to Structure transform, that intercepts calls to the `ComponentBond` provider. If there is a miss to a specific comp id, the bond perception routine is run and results are cached in the provider for subsequent access.

This would live in an extension. It seems that this won't necessitate any code change in the Mol* core code.

arose:
> @arose before starting with a new integration could you validate the following?
> 
> The functionality could be implemented as a Structure to Structure transform, that intercepts calls to the `ComponentBond` provider. If there is a miss to a specific comp id, the bond perception routine is run and results are cached in the provider for subsequent access.
> 
> This would live in an extension. It seems that this won't necessitate any code change in the Mol* core code.

I was thinking to use Unit.create to create new units and provide bonds in the props.

dsehnal:
> > @arose before starting with a new integration could you validate the following?
> > The functionality could be implemented as a Structure to Structure transform, that intercepts calls to the `ComponentBond` provider. If there is a miss to a specific comp id, the bond perception routine is run and results are cached in the provider for subsequent access.
> > This would live in an extension. It seems that this won't necessitate any code change in the Mol* core code.
> 
> I was thinking to use Unit.create to create new units and provide bonds in the props.

Do we even need to intercept anything?

```ts
const WithBondOrders = Transform.create({
  apply: (structure: Structure, params: { .. }) => {
    const bonds = computeDifferentBonds(structure, params);
    return applyBonds(structure, bonds);
  }
}
```

papillot:
> I was thinking to use Unit.create to create new units and provide bonds in the props.

@arose I think that calls to `Unit.getChild` will get past the bonds from these new units and re-derive the bond orders from the parent model directly?

> Do we even need to intercept anything?
> 
> ```ts
> const WithBondOrders = Transform.create({
>   apply: (structure: Structure, params: { .. }) => {
>     const bonds = computeDifferentBonds(structure, params);
>     return applyBonds(structure, bonds);
>   }
> }
> ```

@dsehnal In this scenario, we would eagerly create the entire bond graph, vs lazily? I guess it's a trade of that may be acceptable for an extension: consumers are likely interested in bond orders and less likely to render mesoscale models.

arose:
> > I was thinking to use Unit.create to create new units and provide bonds in the props.
> 
> @arose I think that calls to `Unit.getChild` will get past the bonds from these new units and re-derive the bond orders from the parent model directly?

Right, this is more complicated. Maybe do a subclass of Unit.Atomic with your own .bond method. Would need to replace the Atomic() calls in the Atomic class with this.constructor().

dsehnal:
> > I was thinking to use Unit.create to create new units and provide bonds in the props.
> 
> @arose I think that calls to `Unit.getChild` will get past the bonds from these new units and re-derive the bond orders from the parent model directly?
> 
> > Do we even need to intercept anything?
> > ```ts
> > const WithBondOrders = Transform.create({
> >   apply: (structure: Structure, params: { .. }) => {
> >     const bonds = computeDifferentBonds(structure, params);
> >     return applyBonds(structure, bonds);
> >   }
> > }
> > ```
> 
> @dsehnal In this scenario, we would eagerly create the entire bond graph, vs lazily? I guess it's a trade of that may be acceptable for an extension: consumers are likely interested in bond orders and less likely to render mesoscale models.

@papillot 
Right: we could make "bond provider" customizable at the structure level and it would be lazily resolved. Wdyt?

papillot:
> > > Do we even need to intercept anything?
> > > ```ts
> > > const WithBondOrders = Transform.create({
> > >   apply: (structure: Structure, params: { .. }) => {
> > >     const bonds = computeDifferentBonds(structure, params);
> > >     return applyBonds(structure, bonds);
> > >   }
> > > }
> > > ```
> > 
> > 
> > @dsehnal In this scenario, we would eagerly create the entire bond graph, vs lazily? I guess it's a trade of that may be acceptable for an extension: consumers are likely interested in bond orders and less likely to render mesoscale models.
> 
> @papillot Right: we could make "bond provider" customizable at the structure level and it would be lazily resolved. Wdyt?

@dsehnal there are currently 4 different procedures to construct the bond graph depending on the source (SDF/MOL2 provide everything, mmCIF provide a mix of chem_comp + struct_conn bonds, PDB may provide some struct_conn without bond order, fallback is distance). Having an agnostic way for requesting the bond graph, would make customizations "easier".

I am actually wondering whether the simplest and future proof way could be to brute force it and build the bond graph eagerly in that transform (i.e. drop the laziness). This will populate an IndexPair provider, same as a model loaded from MOL2 or SDF. As Alex envisioned, the future steps will likely be more editing capabilities like adding Hydrogens. In that scenario, having the full graph makes sense.

dsehnal:
It's ok to brute-force it inside a "modifier" transform as you would have to include it manually anyway.

Another option is to add a parameter to `CommonStructureParams`, something like `bondProviderName` + add `PluginContext.structure.bondProviderRegistry`:

```ts
class PluginContext {
   readonly structure = {
     bondProviderRegistry: ...
   }
}
```

arose:
> It's ok to brute-force it inside a "modifier" transform as you would have to include it manually anyway.
> 
> Another option is to add a parameter to `CommonStructureParams`, something like `bondProviderName` + add `PluginContext.structure.bondProviderRegistry`:
> 
> ```ts
> class PluginContext {
>    readonly structure = {
>      bondProviderRegistry: ...
>    }
> }
> ```

bondProviderRegistry sounds interesting. To easily integrate it with the current bond compute code it would need to be on the model. units only know about the model not the structure. 

I think the registry need to be a global since structure/model don't know anything about PluginContext or we inject the registry as a model property during model creation in the state. 

Probably want to expose the current bond compute so the other methods can use it to boot strap their calculation.

PR #1862, opened 2026-06-27
Feat/spherical harmonic surface

corredD:
# Description
 ## Add Spherical Harmonic Surface representation
  
  Adds a `spherical-harmonic-surface` structure representation: a smooth shape **envelope**
  approximated by a real spherical-harmonic expansion of the radial function r(θ,φ) about a
  centroid. `sphericalHarmonicL` controls the level of detail; `maxLobes` splits genuinely
  non-star-shaped inputs into blended star-shaped lobes.

  Three visual flavours, differing only in which atoms each envelope is fit to:
  - **units** (default) — one envelope per chain, instanced across the assembly natively;
  - **structure** — one envelope over the whole input merged (a coarse single-object outline
  of an entire complex);
  - **assembly** — one envelope fit to the asymmetric unit and replicated across the assembly
  operators.
  
image

image

  ### Why fit atoms, not a computed molecular surface

  The fit is to **atom positions** (each pushed out by its size + `radiusOffset`), not to a
  marching-cubes molecular surface. At the low degrees used here (L ≈ 6–15) the truncated
  single-center expansion is **low-pass** — well below the ~1.4 Å probe-scale detail that
  separates the solvent-excluded surface from a vdW+probe atom envelope — so fitting the
  molecular surface yields the *same* low-order coefficients at substantially higher cost
  (Ritchie & Kemp 1999, *J Comput Chem* 20:383; Duncan & Olson 1993, *Biopolymers* 33:219). A
  surface/volume-based SH or 3D-Zernike expansion only earns its cost for
  shape-complementarity docking or property mapping onto the real surface — neither of which
  this coarse-envelope representation targets.

  ### Caveat

  A single-center radial expansion is **star-convex only**: it can represent surfaces
  single-valued in r from the center, so pockets and concavities collapse to the outermost
  crossing. It's best understood as a smooth shape envelope, not a faithful molecular
  surface. (Documented in the representation description and param help.)

  ### Notes
  
  - New, additive only — no changes to shared/core files.
  - Self-contained SH math in `mol-math/geometry/spherical-harmonics.ts` (reuses Mol*'s 
  generic `svd`/`Matrix`), with a `_spec` test.

## Actions


dsehnal:
Looks nice. Although I think I would prefer this being an extension.

- Spherical harmonics can stay in mol-geometry
- Move the representation to `extensions/` and add a plugin behaviour that registers the representation.

arose:
This has a very nice clean look.

Could this work without any MC? It would be nice to have a surface variant that can go from a point cloud (atoms) to a mesh without a grid and MC.

Ribosomes look odd, fit is not good, even with many lobes.

corredD:
> Could this work without any MC? It would be nice to have a surface variant that can go from a point cloud (atoms) to a mesh without a grid and MC.
I kept a MC for a watertight option when we have multiple blob for one chain.
> Ribosomes look odd, fit is not good, even with many lobes.
I though a Kmeans approach would look good on theses long and big chains. I tried splitting along the sequence too which give interesting shape, but the kmeans looks the best for me. @arose what do you think ? 

PDB 9QF4 watertight on : 
Screenshot 2026-06-28 at 9 28 31 AM
Screenshot 2026-06-28 at 9 28 15 AM

corredD:
close in favor of https://github.com/molstar/molstar/pull/1886

Issue #1872, opened 2026-07-08
LAMMPS trajectory doesn't work as expected


Himanshuraj677:
Hi,

I found a possible issue while loading a LAMMPS trajectory in Mol*.

When I load my `.data` + `.lammpstrj` files in Mol*, the trajectory doesn't look correct. In the last frame, the atoms are still mixed, whereas the same trajectory in **OVITO** shows the atoms clearly segregated.

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/f25ed880-8d7c-4787-b830-bdaf71452e7d" />

<img width="1019" height="759" alt="Image" src="https://github.com/user-attachments/assets/ab1e9ab6-2044-4df7-9152-9eb915be4a3a" />

Interestingly, if I export the **last frame** as a separate `.data` file and load it directly into Mol*, it looks exactly like OVITO. So the issue only happens when viewing the frame through the trajectory.

<img width="1907" height="867" alt="Image" src="https://github.com/user-attachments/assets/5445fcd2-20cb-462a-98e0-7c3c8795c499" />

I also noticed that selecting **Color Theme → Atom Id** only shows **two colors**, which looks more like **Atom Type** coloring (my system has only two atom types).

I've attached the files to reproduce the issue:

[https://limewire.com/d/wkmCp#CoGc3cX3UZ](https://limewire.com/d/wkmCp#CoGc3cX3UZ)

Is this is a bug in the LAMMPS trajectory reader or if I'm missing something?

Thanks!

corredD:
Thanks for sharing the data, I was able to reproduce it. Looks like the parser was assuming the trajectories used the same atom order than the data file. 

<img width="690" height="619" alt="Image" src="https://github.com/user-attachments/assets/cd0931d9-32f3-4daa-ac47-b0837b3b1034" />

And yes the Atom type is recorded in the Atom Id color theme, if you want to color by atom order/index use the Element index color theme.

<img width="649" height="603" alt="Image" src="https://github.com/user-attachments/assets/c5cfcd09-6104-431e-b7d8-9140221e1eb2" />

Fix proposed in PR #1873.

Himanshuraj677:
I found the same issue in the data file parser. If the atom IDs are not in order, Mol* shows a random/incorrect structure. Could you please check the attached `ethanol(1).data` file?

https://limewire.com/d/wkmCp#CoGc3cX3UZ

[...]

PR #1873, opened 2026-07-09
Fix lammps trajectory

corredD:
# Description
LAMMPS dumps can list atoms in a different order each frame, so coordinates were landing on the wrong atoms (frames out of sync with each other and with the topology). Each frame is now scattered into canonical atom-id order (index = id − 1); falls back to file order if ids aren't a contiguous 1..N set. Applied to both the per-frame coordinates and the .lammpstrj topology in src/mol-model-formats/structure/lammps-trajectory.ts.

See issue #1872.

## Actions

### DIFF
--- a/src/mol-model-formats/structure/lammps-data.ts
+++ b/src/mol-model-formats/structure/lammps-data.ts
@@ -28,7 +28,6 @@ async function getModels(mol: LammpsDataFile, ctx: RuntimeContext, unitsStyle: U
     const count = atoms.count;
     const scale = lammpsUnitStyles[unitsStyle].scale;
     const type_symbols = new Array<string>(count);
-    const id = new Int32Array(count);
     const cx = new Float32Array(count);
     const cy = new Float32Array(count);
     const cz = new Float32Array(count);
@@ -45,11 +44,10 @@ async function getModels(mol: LammpsDataFile, ctx: RuntimeContext, unitsStyle: U
         cx[j] = atoms.x.value(j) * scale;
         cy[j] = atoms.y.value(j) * scale;
         cz[j] = atoms.z.value(j) * scale;
-        id[j] = atomId;
         if (atomId > maxId) maxId = atomId;
     }
     const rowOfId = new Int32Array(maxId + 1).fill(-1);
-    for (let j = 0; j < count; j++) rowOfId[id[j]] = j;
+    for (let j = 0; j < count; j++) rowOfId[atoms.atomId.value(j)] = j;
 
     const MOL = Column.ofConst('MOL', count, Column.Schema.str);
     const asym_id = Column.ofLambda({
@@ -69,7 +67,7 @@ async function getModels(mol: LammpsDataFile, ctx: RuntimeContext, unitsStyle: U
         Cartn_x: Column.ofFloatArray(cx),
         Cartn_y: Column.ofFloatArray(cy),
         Cartn_z: Column.ofFloatArray(cz),
-        id: Column.ofIntArray(id),
+        id: atoms.atomId,
 
         label_asym_id: asym_id,
         label_atom_id: type_symbol,
--- a/src/mol-model-formats/structure/lammps-data.ts
+++ b/src/mol-model-formats/structure/lammps-data.ts
@@ -34,15 +34,19 @@ async function getModels(mol: LammpsDataFile, ctx: RuntimeContext, unitsStyle: U
     const cz = new Float32Array(count);
     const model_num = new Int32Array(count);
 
-    let offset = 0;
+    // A LAMMPS `.data` file may list atoms in any order (rows are not necessarily sorted by atom id),
+    // while the Bonds section references atoms by id. Map each id to its row so bonds connect the
+    // correct atoms (handles any unique-id scheme, not just contiguous 1..count).
+    const rowOfId = new Map<number, number>();
     for (let j = 0; j < count; j++) {
-        type_symbols[offset] = atoms.atomType.value(j).toString();
-        cx[offset] = atoms.x.value(j) * scale;
-        cy[offset] = atoms.y.value(j) * scale;
-        cz[offset] = atoms.z.value(j) * scale;
-        id[offset] = atoms.atomId.value(j) - 1;
-        model_num[offset] = 0;
-        offset++;
+        const atomId = atoms.atomId.value(j);
+        type_symbols[j] = atoms.atomType.value(j).toString();
+        cx[j] = atoms.x.value(j) * scale;
+        cy[j] = atoms.y.value(j) * scale;
+        cz[j] = atoms.z.value(j) * scale;
+        id[j] = atomId - 1;
+        model_num[j] = 0;
+        rowOfId.set(atomId, j);
     }
 
     const MOL = Column.ofConst('MOL', count, Column.Schema.str);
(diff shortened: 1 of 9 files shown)


src/mol-model-formats/structure/lammps-trajectory.ts
```diff
@@ -21,6 +21,30 @@ import { MoleculeType } from '../../mol-model/structure/model/types';
...
+    const order = new Int32Array(count);
+    const seen = new Uint8Array(count);
+    for (let j = 0; j < count; j++) {
+        const dst = atomId.value(j) - 1;
+        if (dst < 0 || dst >= count || seen[dst]) return undefined;
+        seen[dst] = 1;
+        order[j] = dst;
+    }
+    return order;
+}
```
dsehnal: - Only allocate the order/seen if they are actually needed
- Add `dst === j` to check for identity, for identity, and also return `undefined` (no need to allocate anything) if all indices are identity

src/mol-model-formats/structure/lammps-trajectory.ts
```diff
@@ -21,6 +21,42 @@ import { MoleculeType } from '../../mol-model/structure/model/types';
...
+    for (let j = 0; j < count; j++) {
+        const dst = atomId.value(j) - 1;
+        if (dst < 0 || dst >= count) return undefined;
+        if (!order) {
+            // still in the leading identity run - nothing to permute yet
+            if (dst === j) continue;
+            // first out-of-order row: allocate and backfill the identity prefix [0, j)
+            order = new Int32Array(count);
+            seen = new Uint8Array(count);
+            for (let k = 0; k < j; k++) { order[k] = k; seen[k] = 1; }
```
dsehnal, line 58: You can add one more check here: if `dst < j` then it was definitely seen and you can return undefined.

dsehnal (approved):

src/mol-model-formats/structure/lammps-data.ts
```diff
@@ -34,15 +34,19 @@ async function getModels(mol: LammpsDataFile, ctx: RuntimeContext, unitsStyle: U
...
-        cz[offset] = atoms.z.value(j) * scale;
-        id[offset] = atoms.atomId.value(j) - 1;
-        model_num[offset] = 0;
-        offset++;
+        const atomId = atoms.atomId.value(j);
+        type_symbols[j] = atoms.atomType.value(j).toString();
+        cx[j] = atoms.x.value(j) * scale;
+        cy[j] = atoms.y.value(j) * scale;
+        cz[j] = atoms.z.value(j) * scale;
+        id[j] = atomId - 1;
```
arose: not new in this PR, but why not just use atomId?

src/mol-model-formats/structure/lammps-data.ts
```diff
@@ -34,15 +34,19 @@ async function getModels(mol: LammpsDataFile, ctx: RuntimeContext, unitsStyle: U
...
+        model_num[j] = 0;
```
arose: also not new but this should be a const column

src/mol-model-formats/structure/lammps-data.ts
```diff
@@ -34,15 +34,19 @@ async function getModels(mol: LammpsDataFile, ctx: RuntimeContext, unitsStyle: U
     const cz = new Float32Array(count);
     const model_num = new Int32Array(count);
 
-    let offset = 0;
+    // A LAMMPS `.data` file may list atoms in any order (rows are not necessarily sorted by atom id),
+    // while the Bonds section references atoms by id. Map each id to its row so bonds connect the
+    // correct atoms (handles any unique-id scheme, not just contiguous 1..count).
+    const rowOfId = new Map<number, number>();
```
arose: do we really need a map here or just an Int32 array with length of max(atomId) - 1? what is atom id min/max usually, compared to count?

src/mol-model-formats/structure/lammps-trajectory.ts
```diff
@@ -21,6 +21,44 @@ import { MoleculeType } from '../../mol-model/structure/model/types';
 import { ModelFormat } from '../format';
 import { ModelSymmetry } from './property/symmetry';
 
+/**
+ * A LAMMPS dump may list atoms in a different order in every frame (atoms are
+ * reordered as they migrate between MPI domains), and that order generally does
```
arose: MPI?

src/mol-model-formats/structure/lammps-trajectory.ts
```diff
@@ -90,21 +132,29 @@ async function getModels(mol: LammpsTrajectoryFile, ctx: RuntimeContext, unitsSt
...
-        model_num[offset] = 0;
-        offset++;
+        const dst = order ? order[j] : j;
+        type_symbols[dst] = atoms.atomType.value(j).toString();
+        cx[dst] = atoms.x.value(j) * sx + ox;
+        cy[dst] = atoms.y.value(j) * sy + oy;
+        cz[dst] = atoms.z.value(j) * sz + oz;
+        id[dst] = atoms.atomId.value(j);
+        molecule_ids[dst] = atoms.moleculeId.value(j);
+        model_num[dst] = 0;
```
arose: should be const column

src/mol-model-formats/structure/lammps-trajectory.ts
```diff
@@ -90,21 +132,29 @@ async function getModels(mol: LammpsTrajectoryFile, ctx: RuntimeContext, unitsSt
...
-        id[offset] = atoms.atomId.value(j);
-        model_num[offset] = 0;
-        offset++;
+        const dst = order ? order[j] : j;
+        type_symbols[dst] = atoms.atomType.value(j).toString();
+        cx[dst] = atoms.x.value(j) * sx + ox;
+        cy[dst] = atoms.y.value(j) * sy + oy;
+        cz[dst] = atoms.z.value(j) * sz + oz;
+        id[dst] = atoms.atomId.value(j);
+        molecule_ids[dst] = atoms.moleculeId.value(j);
```
arose: directly create asymId array? molecule_ids is only used for asymids, right?

arose (changes requested):

arose (changes requested):
some comments, spec for the reordering would be nice, thanks

src/mol-model-formats/structure/lammps-data.ts
corredD: That's a good question, I don't have enough samples files. We can assume max ID is close to count, but we can support if the Id are sparse (e.g. 1,3,5...), this will work with holes (-1, e.g. 1,-1,3,-1,5).

src/mol-model-formats/structure/lammps-trajectory.ts
corredD: I believe the unordered id comes from the multiprocessing Lammps run (MPI), where each processor writes its atoms as they arrive.

corredD:
> some comments, spec for the reordering would be nice, thanks

Thanks @arose, the changes you requested and pointed out actually improved the performance ! this is great. Should the provider visual use a custom preset for the lammps data so it always use the spacefill with structure element sphere by default ?

arose:
> > some comments, spec for the reordering would be nice, thanks
> 
> Thanks @arose, the changes you requested and pointed out actually improved the performance ! this is great. Should the provider visual use a custom preset for the lammps data so it always use the spacefill with structure element sphere by default ?

there is an approach to handle that but the threshold was too high, addressed in https://github.com/molstar/molstar/pull/1873/commits/282aacf4957ceb8cfc1e29c0f288236bed793113

src/mol-model-formats/structure/lammps-data.ts
```diff
@@ -32,18 +32,24 @@ async function getModels(mol: LammpsDataFile, ctx: RuntimeContext, unitsStyle: U
...
+    let maxId = 0;
     for (let j = 0; j < count; j++) {
-        type_symbols[offset] = atoms.atomType.value(j).toString();
-        cx[offset] = atoms.x.value(j) * scale;
-        cy[offset] = atoms.y.value(j) * scale;
-        cz[offset] = atoms.z.value(j) * scale;
-        id[offset] = atoms.atomId.value(j) - 1;
-        model_num[offset] = 0;
-        offset++;
+        const atomId = atoms.atomId.value(j);
```
arose, line 42: atoms.atomId can now be used as is, not need for an extra array

arose (approved):
thanks @corredD

PR #1887, opened 2026-07-22
Camera transitions

midlik:
# Description

This PR adds customization of camera transition trajectory shape. 

Changing camera transition trajectory shape and easing (added in https://github.com/molstar/molstar/pull/1841) is available via:
  - Parameters of `PluginCommands`
  - Parameters in `PluginState.Snapshot`, i.e. they are saved in MOLJ, and can be changed when creating a snapshot in the UI (Plugin State > Save Options)
  - `transitionTrajectory` and `transitionEasing` parameters to `FocusLoci` behavior
  - `cameraResetTrajectory` and `cameraResetEasing` parameters to `Canvas3DParams`
  - In MolViewSpec via the new `transition` node with params `duration_ms`, `trajectory`, `easing`.

Supported trajectory shapes:
- `linear`: interpolates along a straight line with constant absolute speed - the original
- `linear-relative`: like `linear` but moves relatively slower when zoomed-in more (constant speed relative to radius), this creates more linear impression when zooming in/out
- `leap`: zooms out during the transition to capture both the initial and the final visible sphere. The transition is partially linearized if the initial and final spheres overlap, and becomes fully linear if the overlap is big enough (this is to avoid disturbing leaping between near targets)
- `leap-relative`: like `leap` but moves relatively slower when zoomed-in more (constant speed relative to radius)

Other changes:
- MolViewSpec snapshot metadata:
  - `linger_duration_ms` renamed to `duration_ms` (old name still works but is deprecated)
  - `transition_duration_ms` becomes deprecated, preferred way is to use `transition` node.

Testing files (MVSJ):

[transitions-mvsj.zip](https://github.com/user-attachments/files/30473025/transitions-mvsj.zip)

Associated mol-view-spec PR: https://github.com/molstar/mol-view-spec/pull/110

---

Notes:
- We might consider changing the default transition for focus and camera reset (currently 250ms linear with linear easing) to 500ms leap with sin-in-out easing.
- `linear-relative` and `leap-relative` trajectories feel more natural then `linear` and `leap` when used without easing. However, in combination with easing it doesn't make much difference and actually `leap` looks better than `leap-relative` (`leap-relative` looks to hasty)

## Actions


src/extensions/mvs/tree/mvs/mvs-tree.ts
```diff
@@ -399,6 +375,44 @@ export const MVSTreeSchema = TreeSchema({
...
+                /** Camera transition trajectory shape. "linear": interpolates along a straight line with constant absolute speed; "linear-relative": like "linear" but moves relatively slower when zoomed-in more; "leap": zooms out during the transition to capture both the initial and the final camera target (becomes linear if the targets are near); "leap-relative": like "leap" but moves relatively slower when zoomed-in more. Defaults to "linear". */
+                shape: OptionalField(CameraTransitionShapeT, 'linear', 'Camera transition trajectory shape. "linear": interpolates along a straight line with constant absolute speed; "linear-relative": like "linear" but moves relatively slower when zoomed-in more; "leap": zooms out during the transition to capture both the initial and the final camera target (becomes linear if the targets are near); "leap-relative": like "leap" but moves relatively slower when zoomed-in more. Defaults to "linear".'),
```
dsehnal: rename to trajectory or interpolation?

src/extensions/mvs/tree/mvs/param-types.ts
```diff
@@ -138,6 +138,15 @@ export const Matrix = list(float); // TODO impl custom types Matrix3x3 and Matri
 export type LabelAttachments = 'bottom-left' | 'bottom-center' | 'bottom-right' | 'middle-left' | 'middle-center' | 'middle-right' | 'top-left' | 'top-center' | 'top-right';
 export const LabelAttachments = literal<LabelAttachments>('bottom-left', 'bottom-center', 'bottom-right', 'middle-left', 'middle-center', 'middle-right', 'top-left', 'top-center', 'top-right');
 
+/** `shape` parameter values for `transition` node in MVS tree, describes camera transition trajectory shape */
+export type CameraTransitionShapeT = 'linear' | 'linear-relative' | 'leap' | 'leap-relative';
+export const CameraTransitionShapeT = literal<CameraTransitionShapeT>('linear', 'linear-relative', 'leap', 'leap-relative');
+
+/** `easing` parameter values for `transition` node in MVS tree, describes easing function */
+export type EasingT = 'linear' | 'bounce-in' | 'bounce-out' | 'bounce-in-out' | 'circle-in' | 'circle-out' | 'circle-in-out' | 'cubic-in' | 'cubic-out' | 'cubic-in-out' | 'exp-in' | 'exp-out' | 'exp-in-out' | 'quad-in' | 'quad-out' | 'quad-in-out' | 'sin-in' | 'sin-out' | 'sin-in-out';
```
dsehnal, line 146: There is already an easing type here: https://github.com/molstar/molstar/blob/master/src/extensions/mvs/tree/animation/animation-tree.ts#L12

Can you please merge these 2? Prolly use yours in place of mine since that one is private.

src/mol-canvas3d/camera/transition-functions.ts
```diff
@@ -0,0 +1,326 @@
...
+
+
+/** Simple linear transition with constant absolute speed. */
+export function transition_linear(out: Camera.Snapshot, t: number, source: Camera.Snapshot, target: Camera.Snapshot): void {
+    return transition_linear_internal(out, t, t, source, target);
+}
+
+/** Linear transition with constant speed relative to visible sphere radius (move slower where zoomed-in more, dtarget / dt = (r + CONST_REL_SPEED_OFFSET) * const).
+ * Rotational component of the transition has constant speed. */
+export function transition_linear_relative(out: Camera.Snapshot, t: number, source: Camera.Snapshot, target: Camera.Snapshot): void {
```
dsehnal: We use `snakeCase` everywhere else in the codebase, can you please update this?

dsehnal (approved):
Look great, thanks for adding this.

src/extensions/mvs/tree/mvs/param-types.ts
midlik: Generally, i define all MVS types explicitely, without importing anything from Molstar. To avoid new values creeping into MVS when added in Molstar, and not being in sync with the Python package.
And also to keep the type definition file for LLMs clean.

src/extensions/mvs/tree/mvs/mvs-tree.ts
midlik: trajectory would make sense, i guess. i wanted to avoid confusion with "trajectory" as a sequence of conformations. but then, "shape" could also get confused with other things.
i can rename it.

src/mol-canvas3d/camera/transition-functions.ts
midlik: Fixed

src/extensions/mvs/tree/mvs/param-types.ts
dsehnal: This type is already in MVS for the animation functionality

src/extensions/mvs/tree/mvs/mvs-tree.ts
dsehnal: How about path?

midlik: oops, i already renamed it

src/extensions/mvs/tree/mvs/param-types.ts
midlik: oh, i see, i'll look for it

midlik: Merged.

dsehnal (approved):

PR #1890, opened 2026-07-23
Add external-volume color option and color-mapping modes to VTP

corredD:
When comparing the color channels with ParaView, some results appeared different. ParaView seemed to identify more effective minimum and maximum values for the color-map domain.

This update adds:

Minimum and maximum value labels
New automatic range modes, including Best Fit and Z-score
An option to color the model using values sampled from an external volume

## Actions


dsehnal (approved):
Looks reasonable

arose:
At best, external volume/structure would be supported for all shape formats. This will need substantial changes to how Shape 3D Representation is done. Need to think about this.

corredD:
what about splitting, and keeping only the color mapping update  ?

corredD:
closing in favor of a future change for all shape formats.

PR #1898, opened 2026-07-28
Fix nucleic cartoon trace picking returning empty loci

giagitom:
# Description

On reduced trace structures (atoms limited to {CA, P}) hovering the nucleic cartoon polymer-trace resolved no loci, so it could not be highlighted. The model trace atom for nucleic residues is O3' (traceElementIndex), which is absent from a P-only unit, and getResidueLoci required that exact element to be present. It now resolves the residue from the representative element and gathers whatever of its atoms are in the unit. Protein was unaffected because its trace atom (CA) is kept by such selections.

## Actions


arose (approved):
Thank you, good find. I tweak it a bit to have an early return. Please check.

giagitom:
Thanks @arose, fixed a bug with last commit. It looks good to me.

PR #1906, opened 2026-08-09
Cell & symmetry improvements

arose:
# Description

- Improve `Cell` handling
    - Fix volume calculation for non-orthogonal cells
    - Use as base for `Unitcell` shape
    - Add `order` property (for AU volume estimation)
- Add per-format cache for `FormatProperty`
    - Use for `ModelSymmetry` and `ComponentBond`
- `ModelSymmetry` improvements
    - Fix property not beeing dynamic
    - Defer Symmetry calculation in ModelSymmetry.fromData

## Actions

### DIFF
--- a/src/mol-math/geometry/spacegroup/construction.ts
+++ b/src/mol-math/geometry/spacegroup/construction.ts
@@ -42,6 +42,11 @@ namespace SpacegroupCell {
     /** Create a 'P 1' with cellsize [1, 1, 1] */
     export const Zero: SpacegroupCell = create('P 1', Vec3.create(1, 1, 1), Vec3.create(Math.PI / 2, Math.PI / 2, Math.PI / 2));
 
+    /** True if the cell belongs to a known spacegroup */
+    export function is(cell: Cell): cell is SpacegroupCell {
+        return typeof (cell as SpacegroupCell).name === 'string';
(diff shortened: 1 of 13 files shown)


src/mol-math/geometry/spacegroup/construction.ts
```diff
@@ -42,6 +42,11 @@ namespace SpacegroupCell {
     /** Create a 'P 1' with cellsize [1, 1, 1] */
     export const Zero: SpacegroupCell = create('P 1', Vec3.create(1, 1, 1), Vec3.create(Math.PI / 2, Math.PI / 2, Math.PI / 2));
 
+    /** True if the cell belongs to a known spacegroup */
+    export function is(cell: Cell): cell is SpacegroupCell {
+        return typeof (cell as SpacegroupCell).name === 'string';
```
dsehnal, line 47: Should this actually do a lookup in the table?

dsehnal (approved):

src/mol-math/geometry/spacegroup/construction.ts
arose: not that strict, it is by construction, just need to match the SpacegroupCell interface, changed the comment

PR #1907, opened 2026-08-11
Particles mmcif optimization

arose:
# Description

- optimize mmcif particle parsing/creation
- support fibers in cellpack mmcif

## Actions

### DIFF
--- a/src/mol-model-formats/particles/mmcif.ts
+++ b/src/mol-model-formats/particles/mmcif.ts
@@ -16,6 +16,7 @@ import { ModelFormat } from '../format';
 import { binaryCifHasColumn, getBinaryCifHeader } from '../../mol-io/common/binary-cif';
 import { StringLike } from '../../mol-io/common/string-like';
 import { FileNameInfo } from '../../mol-util/file-info';
+import { RuntimeContext } from '../../mol-task';
 
 export type MmcifVariant = 'auto' | 'cellpack' | 'petworld' | 'standard';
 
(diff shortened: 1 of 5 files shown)


src/mol-model-formats/particles/mmcif.ts
```diff
@@ -108,6 +110,57 @@ export function createParticleListFromMmcifAssembly(cifFile: CifFile, options: M
...
+interface AtomSiteBounds {
+    readonly center: Vec3
+    readonly radius: number
+}
+
+/**
+ * Applying the operator of a particle to the centroid of its reference atoms gives a meaningful
+ * position even when the operator has zero translation (pure rotation).
+ */
+function getAtomSiteBounds<K>(db: mmCIF_Database, key: (row: number) => K): Map<K, AtomSiteBounds> {
```
corredD: Two full synchronous passes over `_atom_site` here, and `createParticleListFromMmcifAssembly` takes no `RuntimeContext`, so
  the `Task` in `transforms/particles.ts` can't poll the abort token. Only ~900 ms on a 2.93M-atom CellPack file so not a
  problem today, but it scales linearly and it's the last pass in the file without progress/cancel now that operators are down
  to 39 ms. Worth threading `ctx` through, or leave it?

src/mol-model-formats/particles/operators.ts
```diff
@@ -0,0 +1,199 @@
...
+    }
+    if (ops.index) return (ops.index as ReadonlyMap<string, number>).get(id) ?? -1;
+    const i = +id - 1;
+    return Number.isInteger(i) && i >= 0 && i < ops.count ? i : -1;
+}
+
+export function getParticleOperator(ops: ParticleOperators, operatorIndex: number, out: Mat4): Mat4 {
+    const { data } = ops;
+    const o = operatorIndex * 16;
+    for (let k = 0; k < 16; ++k) out[k] = data[o + k];
```
corredD, line 129: why not using return Mat4.fromArray(out, ops.data, operatorIndex * 16);

corredD (commented):
Great improvement, thanks for the cellpack fibers. We could do something similar for the YASARA/PetWorld files, where there's
  no naming convention to go on: look at the distance between consecutive instances of a model — if it's constant it's a fiber
  (the two microtubules in postsynstubby_2.cif come out at a dead-constant 52-54 A repeat), and then split into segments wherever a step is
  much larger than the median.

src/mol-model-formats/particles/mmcif.ts
arose: yes, very good point

src/mol-model-formats/particles/operators.ts
arose: observed it to be faster to inline here

arose:
> Great improvement, thanks for the cellpack fibers. We could do something similar for the YASARA/PetWorld files, where there's no naming convention to go on: look at the distance between consecutive instances of a model — if it's constant it's a fiber (the two microtubules in postsynstubby_2.cif come out at a dead-constant 52-54 A repeat), and then split into segments wherever a step is much larger than the median.

i would like to defer such guessing for a later iteration

corredD (approved):
lgtm

PR #1908, opened 2026-08-11
 Improve illumination rendering

arose:
# Description
- Add depth-aware denoising
- Optimize screen-space ray marching

## Actions


giagitom (commented):
Hi @arose,

i have found that there is a lighter rim on the interface of objects with higher depth gap that feels wrong to me.  Also comparing the same scene with master it is generally lighter.

Image

I have also found an other issue related to multisample that seems increased by this PR. 
Using multisample enabled with higher sampling produces more artifacts than on master.   

Image

arose:
Thanks @giagitom - I tried too many things here. I mainly wanted to fix something step-size related. New PR coming.

PR #1911, opened 2026-08-12
Particles target mapping

arose:
# Description

Decorate particles with structures, volumes, shapes.

## Actions


src/mol-model-formats/particles/mmcif.ts
```diff
@@ -103,12 +110,26 @@ export function getAsymIdsFromMmcif(cifFile: CifFile, assemblyId: string): strin
...
-    return await buildCellpackStandardParticleList(cifFile, block, options, variant, ctx);
+
+    // The reference structures of the targets are built via the regular trajectory/model/structure path.
+    const trajectory = await trajectoryFromMmCIF(block, cifFile).runInContext(ctx);
+
+    const built = variant === 'petworld'
+        ? await buildPetworldParticleList(ctx, cifFile, block, options, await getModelTargets(ctx, trajectory))
+        : await buildCellpackStandardParticleList(ctx, cifFile, block, options, variant, await getChainTargets(ctx, trajectory));
+
+    if (options.resolveTargets === false || built.targetMapping.size === 0) return built.list;
```
corredD, line 125: resolveTargets: doesn't skip anything — mmcif.ts:119 builds every Model and every per-chain/per-model Structure before the flag is read here, then discards them

src/mol-repr/particles/representation/target/common.ts
```diff
@@ -0,0 +1,508 @@
...
+
+        // Dots are sized uniformly and scaled by `sizeFactor`; the representation's per-particle
+        // size theme would make every dot as large as the particle it belongs to.
+        const theme = isVolumeDot(target, props)
+            ? { color: representationTheme.color, size: getDotSizeTheme() }
+            : representationTheme;
+
+        const data = targetData(target);
+        const targetChanged = data !== currentTargetData;
+        const particlesChanged = particleIndices !== currentParticleIndices;
```
corredD, line 410: particleIndices comes from getParticleTargetGroups(...).sets[g] — an Interval, which molstar packs into a plain number, so two
  frames with the same target layout compare ===. targetChanged is false too (targets are shared across frames), so
  updateMatrix/updateColor/updateSize all stay false and the render object keeps frame 0's matrices.

src/mol-model/particles/particle-list.ts
```diff
@@ -184,26 +210,121 @@ export function getParticleTransforms(data: ParticleList): Float32Array {
...
+            if (t < p) sorted = false;
+            if (t !== p) runCount += 1;
+            if (t < minId) minId = t;
+            if (t > maxId) maxId = t;
         }
+    }
 
-        data.customProperties.add(ParticleTransformsAsMat4Descriptor);
-        data._propertyData[ParticleTransformsAsMat4Descriptor.name] = transforms;
+    const targetIds = new Int32Array(runCount);
```
corredD, line 285: runCount counts runs of equal adjacent ids. On the sorted fast path runs == distinct ids, so it's fine. On the
  counting-sort fallback it overcounts — the same id appears in several runs — so targetIds gets trailing phantom 0s and sets
  trailing empty sets, breaking the documented "the distinct target ids, ascending" contract. Those phantoms then feed
  getParticleTargetCoverage in formats/particles.ts and flip hasUntargeted, adding an always-empty second spacefill node.

src/mol-repr/particles/representation/target/common.ts
```diff
@@ -0,0 +1,508 @@
...
+    for (const k of Object.keys(params)) {
+        out[k] = { ...params[k], hideIf: (p: { type: string }) => p.type !== type };
+    }
+    return out as P;
+}
+
+/** Params shared by all target kinds; the per-kind groups below are merged on top of these. */
+export const ParticleTargetCommonParams = {
+    ...BaseGeometry.Params,
+    instanceGranularity: PD.Boolean(true),
```
corredD, line 62: what happen if someone set the instanceGranualrity to off ?

src/mol-plugin-state/formats/particles.ts
```diff
@@ -141,37 +209,22 @@ export const MmcifParticlesProvider = DataFormatProvider({
...
-
-        const particleList = StateObjectRef.resolveAndCheck(plugin.state.data, data.list)?.obj?.data;
-        if (particleList?.fibers && particleList.fibers.count > 0) {
-            builder.to(data.list)
-                .apply(StateTransforms.Particles.ParticlesRepresentation3D, { type: { name: 'fibers', params: {} } });
-        }
-
-        return builder.commit();
-    },
+    visuals: (plugin: PluginContext, data: ParticleFormatData) => complexVisuals(plugin, data),
```
corredD, line 212: was the scale: 0.5 important here ? should you pass it ?

src/mol-repr/particles/representation/target/representation.ts
```diff
@@ -0,0 +1,211 @@
...
+                let visual = targetVisuals.get(targetId);
+                if (!visual) {
+                    visual = createTargetVisual(targetId, getNextMaterialId(), webgl);
+                    targetVisuals.set(targetId, visual);
+                }
+
+                await visual.createOrUpdate({ webgl, runtime }, _theme, targetProps(target), particles, indices, target);
+
+                if (visual.renderObject) {
+                    renderObjects.push(visual.renderObject);
```
corredD, line 95: should you pass/apply the _state here ?

corredD (commented):
mostly coding review. I will do testing next with different file format scenario. But why use the URL/File loader and not reference from already loaded object in the scene ?

corredD:
I did some live testing, its looking great so far. Just two comments : 
- the target blob surface representation has a slider for Resolution under custom quality, seems to do nothing.
- under target representation type, shouldn't Structure, Shape, Volume be visible depending on the target type ?

src/mol-model-formats/particles/mmcif.ts
arose: good points, will optimize

src/mol-model/particles/particle-list.ts
arose: yes, will fix

src/mol-plugin-state/formats/particles.ts
arose: not really important just looked a bit better for some examples but now we decorate with structures by default

src/mol-repr/particles/representation/target/representation.ts
arose: yes, will apply state

src/mol-repr/particles/representation/target/common.ts
arose: yeah, nothing, should be hidden, will change

src/mol-repr/particles/representation/target/common.ts
arose: yes, need to take indices into account, will do

arose:
> I did some live testing, its looking great so far. Just two comments :
> 
>     * the target blob surface representation has a slider for Resolution under custom quality, seems to do nothing.
> 

yes, the custom quality params need to be hidden unless quality is set to custom = fixed

>     * under target representation type, shouldn't Structure, Shape, Volume be visible depending on the target type ?

good point, now hiding non-applicable param groups

arose:
thansk @corredD 

also exposed a few more blob-surface params for target repr

PR #1912, opened 2026-08-12
MVS split coloring

midlik:
# Description

**Only to be merged after https://github.com/molstar/molstar/pull/1915 !**

This adds "split color" support to MVS (e.g. 'red/white', '#ff0000/#ffffff'),
which can be applied to carbohydrate symbols and nucleic acid cartoon (other representations just use first component of the split color). This allows implementing SNFG carbohydrate coloring in pure MVS.

I also added a named color dict 'CarbohydrateSymbol' so SNFG coloring can be achieved by:

```
carb.colorFromSource({
    schema: 'all_atomic',
    category_name: 'atom_site',
    field_name: 'label_comp_id',
    palette: {
        kind: 'categorical',
        colors: 'CarbohydrateSymbol',
    },
});
```

Testing files:
[PR1912-mvs-split-coloring.zip](https://github.com/user-attachments/files/30987306/PR1912-mvs-split-coloring.zip)

Corresponding PR in molviewspec:
https://github.com/molstar/mol-view-spec/pull/114

## Actions


arose:
Early comment: please move the carbohydrate and non MVS changes to another PR

midlik:
> Early comment: please move the carbohydrate and non MVS changes to another PR

Thanks Alex, I created a separate PR for the non-MVS changes:
https://github.com/molstar/molstar/pull/1915

src/extensions/mvs/helpers/animation.ts
```diff
@@ -451,10 +469,16 @@ function interpolateColors(start: ColorT | Record<number, ColorT>, end: ColorT |
...
-        return Color.toHexStyle(c);
+        return PaletteFunction.getAsHexStyle(cacheEntry.paletteFn, t);
     }
 
+    if (typeof start === 'string' && typeof end === 'string') {
+        const startColor = cacheEntry.startSplitColor ??= SplitColor.decodeStrict(start);
+        const endColor = cacheEntry.endSplitColor ??= SplitColor.decodeStrict(end);
+        return interpolateSplitColor(startColor, endColor, t);
+    }
+    // the rest of this function does not support split coloring (i.e. ignores color after slash)
```
midlik, line 480: Sorry @dsehnal , i couldn't wrap my head around the rest of this function, i'm not sure what it's trying to achieve, so i left it as is for now (just made sure it won't crash on split color inputs)

dsehnal: @midlik it's to support this: https://github.com/molstar/molstar/blob/master/src/examples/mvs-stories/stories/animation.ts#L214

midlik: OK, i think i fixed the rest of the function, and it works for this kind of use-case. But i didn't check this particular story.

src/extensions/mvs/helpers/animation.ts
```diff
@@ -422,74 +438,78 @@ function interpolateRotation(start: Mat3, end: Mat3 | undefined, t: number, nois
...
+    if (startColors?.kind === 'dict') {
+        if (baseColors === undefined || baseColors === null) throw new Error('Missing color value in the target node');
+        if (typeof baseColors !== 'object') throw new Error('Cannot interpolate from scalar color to color mapping');
 
-    if (typeof startColor === 'object') {
-        if (typeof baseColors !== 'object') {
-            throw new Error('Cannot interpolate from scalar color to color mapping');
-        }
+        const ret: Record<number | string, ColorT> = Array.isArray(baseColors) ? baseColors.slice() as Record<number, ColorT> : { ...baseColors };
+        Object.assign(ret, startColors.dict);
```
dsehnal: Should this get the serialized colors? Something like

```ts
for (const [key, [primary, secondary]] of Object.entries(startColors.dict)) {
    ret[key] = SplitColor.toHexStyle(primary, secondary);
}
```

src/extensions/mvs/helpers/colors.ts
```diff
@@ -69,77 +75,100 @@ const SecondaryStructureColors = {
...
+    }
+    return out;
+}
+
+export const MvsNamedColorDicts: Record<ColorDictNameT, Record<string, ColorT>> = {
+    ElementSymbol: ElementSymbolColorDict,
+    ResidueName: ResidueNameColorDict,
+    ResidueProperties: ResiduePropertyColorDict,
     SecondaryStructure: SecondaryStructureColors,
+    CarbohydrateSymbol: makeCarbohydrateSymbolColors(),
```
dsehnal, line 97: `SaccharideCompIdMap` can change during plugin config (e.g. to use glycam names; `PluginConfig.Structure.SaccharideCompIdMapType`),

so this should be something like:

```
   get CarbohydrateSymbol() {
        return getCachedCarbohydrateSymbolColors();
    },
```

dsehnal:
Also a tine mol-view-spec PR needed with addition of `CarbohydrateSymbol`

src/extensions/mvs/helpers/animation.ts
midlik: You're right, it should.
For my defence, the original was: 
```
const ret = { ...baseColors as any, ...startColor as any };
```

I will fix it

midlik: FIXED

src/extensions/mvs/helpers/colors.ts
midlik: I don't think the content of the named color dict should depend on Molstar config. It's a MVS constant basically, which should have the same value in all MVS implementations.

midlik: Actually the proper way would be to hardcode all MVS named color dicts and named color lists, as they shouldn't really be linked to Molstar's internal constants.
But it would be a lot of repetition (~50 lists with ~10 colors each; CarbohydrateSymbol dict has 1229 entries).

Should I do it this way?

midlik:
> Also a tine mol-view-spec PR needed with addition of `CarbohydrateSymbol`

You're right:
https://github.com/molstar/mol-view-spec/pull/114

src/extensions/mvs/helpers/animation.ts
midlik: I just checked that story and it works as it should.

src/extensions/mvs/helpers/colors.ts
dsehnal: yeah, this is difficult: how about `CarbohydrateSymbol` and `CarbohydrateSymbolGlycam` palette names?

dsehnal: with the default one corresponding to whats in the PDB essentially

src/extensions/mvs/helpers/animation.ts
dsehnal: Fair, seems like AI is smarter than both of us :)

src/extensions/mvs/helpers/colors.ts
midlik: isn't it too niche?
(i mean the glycam codes)

dsehnal: Some people use them. But ok, we can add this later if people complain.

dsehnal (approved):

PR #1914, opened 2026-08-12
Illumination fixes and improvements

giagitom:
# Description

- Fix illumination `auto` thickness mode never being applied
- Fix illumination ray marching stepping over occluders when the acceptance window is narrower than the current step
- Evaluate illumination `auto` thickness at the surface being tested instead of latching it from the shaded pixel

## Actions


arose:
that looks nice, thanks, will dig into it

arose (approved):
very nice, looks good to me!

PR #1915, opened 2026-08-13
Carbohydrate rendering and coloring changes

midlik:
# Description

A few changes in carbohydrate symbol rendering and coloring (this is preparation for MVS split coloring https://github.com/molstar/molstar/pull/1912):

- All carbohydrate symbols are rendered with 2 groups (primary and secondary) and can be potentially colored in two colors
  all-shapes

- `CarbohydrateSymbolColorTheme` decides which shape will be colored by one or two colors

- Changed side length ratio of FlatBox shape from 2:2:1 to 2:1:1 (in accordance with https://pmc.ncbi.nlm.nih.gov/articles/PMC5018049/#cww076C8), please check if this is correct
  Flatbox-side-ratio-change-1MFD

- Fixed DividedCone shape (had inverted base), e.g. 49T in 4y9v
  DividedCone-change-4Y9V

- Fixed shape of known monosaccharides which are in "Unknown" category here (https://en.wikipedia.org/wiki/Symbol_Nomenclature_For_Glycans#/media/File:SNFG.tif)
  - Original plan: Move these monosaccharides to a new category "Other" (FlatHexagon) to distinguish from truly unknown monosaccharides (shape depending on ring count)
  - After discussion: We decided to keep them in "Unknown" category and render all "Unknown" as FlatHexagon (including truly unknown)
  - e.g. KDO in 3k8d (yellow FlatHexagon in SNFG)
  - e.g. X6Y in 9k1g (no SNFG symbol assigned -> render white FlatHexagon)
  FlatHexagon-fix-3K8D

- Fixed highlighting of two-color carbohydrate symbols (used to highlight only first group), e.g. GCS in 5hwa
  Highlight-fix-5HWA

## Actions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,10 @@ Note that since we don't clearly distinguish between a public and private interf
 - Fix camera reset handling for (temporary) empty scenes (#1903)
 - Remove `firstStepSize` tracing parameter, derive automatically
 - Add `.parseRaw` to `DataFormatProvider` for out of state tree parsing
+- Carbohydrate symbols
+  - All carbohydrate symbols are rendered with 2 groups (primary and secondary) and can be potentially colored in two colors
+  - CarbohydrateSymbolColorTheme decides which shape will be colored by one or two colors
+  - Changed side length ratio of FlatBox shape from 2:2:1 to 2:1:1
 - Camera improvements
   - Support multiple camera transition shapes
   - Add `transitionTrajectory` and `transitionEasing` parameters to `PluginState.Snapshot` (MOLJ) and Plugin State > Save Options
--- a/src/mol-geo/geometry/mesh/builder/sphere.ts
+++ b/src/mol-geo/geometry/mesh/builder/sphere.ts
@@ -1,7 +1,8 @@
 /**
- * Copyright (c) 2018 mol* contributors, licensed under MIT, See LICENSE file for more info.
+ * Copyright (c) 2018-2026 mol* contributors, licensed under MIT, See LICENSE file for more info.
  *
  * @author Alexander Rose <alexander.rose@weirdbyte.de>
+ * @author Adam Midlik <midlik@gmail.com>
  */
 
 import { Vec3, Mat4 } from '../../../../mol-math/linear-algebra';
@@ -9,22 +10,23 @@ import { MeshBuilder } from '../mesh-builder';
 import { Primitive } from '../../../primitive/primitive';
 import { Sphere } from '../../../primitive/sphere';
 
-const sphereMap = new Map<number, Primitive>();
+const sphereCache = new Map<string, Primitive>();
 const tmpSphereMat = Mat4.identity();
 
 function setSphereMat(m: Mat4, center: Vec3, radius: number) {
     return Mat4.scaleUniformly(m, Mat4.fromTranslation(m, center), radius);
 }
 
-export function getSphere(detail: number) {
-    let sphere = sphereMap.get(detail);
+export function getSphere(detail: number, options?: { subset: 'ring' | 'caps' | undefined }) {
+    const cacheKey = `${detail}/${options?.subset ?? ''}`;
+    let sphere = sphereCache.get(cacheKey);
     if (sphere === undefined) {
-        sphere = Sphere(detail);
-        sphereMap.set(detail, sphere);
+        sphere = Sphere(detail, options);
+        sphereCache.set(cacheKey, sphere);
     }
     return sphere;
 }
 
-export function addSphere(state: MeshBuilder.State, center: Vec3, radius: number, detail: number) {
-    MeshBuilder.addPrimitive(state, setSphereMat(tmpSphereMat, center, radius), getSphere(detail));
+export function addSphere(state: MeshBuilder.State, center: Vec3, radius: number, detail: number, options?: { subset: 'ring' | 'caps' | undefined }) {
+    MeshBuilder.addPrimitive(state, setSphereMat(tmpSphereMat, center, radius), getSphere(detail, options));
 }
--- a/src/mol-geo/geometry/mesh/builder/sphere.ts
+++ b/src/mol-geo/geometry/mesh/builder/sphere.ts
@@ -5,28 +5,42 @@
  * @author Adam Midlik <midlik@gmail.com>
  */
 
-import { Vec3, Mat4 } from '../../../../mol-math/linear-algebra';
-import { MeshBuilder } from '../mesh-builder';
+import { Mat4, Vec3 } from '../../../../mol-math/linear-algebra';
 import { Primitive } from '../../../primitive/primitive';
 import { Sphere } from '../../../primitive/sphere';
+import { MeshBuilder } from '../mesh-builder';
 
-const sphereCache = new Map<string, Primitive>();
+const sphereCache = new Map<number, Primitive>();
+const sphereSubsetCache = { ring: new Map<number, Primitive>(), caps: new Map<number, Primitive>() };
 const tmpSphereMat = Mat4.identity();
 
 function setSphereMat(m: Mat4, center: Vec3, radius: number) {
     return Mat4.scaleUniformly(m, Mat4.fromTranslation(m, center), radius);
 }
 
-export function getSphere(detail: number, options?: { subset: 'ring' | 'caps' | undefined }) {
-    const cacheKey = `${detail}/${options?.subset ?? ''}`;
-    let sphere = sphereCache.get(cacheKey);
+export function getSphere(detail: number) {
+    let sphere = sphereCache.get(detail);
     if (sphere === undefined) {
-        sphere = Sphere(detail, options);
-        sphereCache.set(cacheKey, sphere);
+        sphere = Sphere(detail);
+        sphereCache.set(detail, sphere);
     }
     return sphere;
 }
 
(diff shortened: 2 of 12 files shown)


arose:
I wrote Oliver Grant an email for input, he is a carbohydrate expert and did the first 3D SNFG version.

dsehnal (commented):
Code looks straightforward and correct.

Can you please include screenshots of before/after for the changes you made for easier review?

arose:
I am generally fine with it. Just want to get Oliver's conformation regarding the FlatBox ratio and handling of Unknown Carbohydrates.

My take so far. To follow SNFG Nomenclature (https://www.ncbi.nlm.nih.gov/glycans/snfg.html) we should use Unknown for the FlatHexagon, i.e., don't change it to a Prism depending on ring member count; add a new category "Unassigned" and use ring -based Prism for it (trying to follow note 11 from the nomenclature page).

src/mol-model/structure/structure/carbohydrates/constants.ts
```diff
@@ -35,7 +36,12 @@ export const SaccharideColors = ColorMap({
 
 export enum SaccharideType {
     Hexose, HexNAc, Hexosamine, Hexuronate, Deoxyhexose, DeoxyhexNAc, DiDeoxyhexose,
-    Pentose, Deoxynonulosonate, DiDeoxynonulosonate, Unknown, Assigned
+    Pentose, Deoxynonulosonate, DiDeoxynonulosonate,
+    /** Flat Hexagon shape, but not unknown monosaccharide */
+    Other,
```
arose, line 41: lets remove `Other` and always use the standard `Unknown`

src/mol-model/structure/structure/carbohydrates/constants.ts
```diff
@@ -68,9 +75,10 @@ const SaccharideTypeShapeMap = {
     [SaccharideType.Pentose]: SaccharideShape.FilledStar,
     [SaccharideType.Deoxynonulosonate]: SaccharideShape.FilledDiamond,
     [SaccharideType.DiDeoxynonulosonate]: SaccharideShape.FlatDiamond,
+    [SaccharideType.Other]: SaccharideShape.FlatHexagon,
     [SaccharideType.Unknown]: SaccharideShape.FlatHexagon,
     [SaccharideType.Assigned]: SaccharideShape.Pentagon,
-};
+} satisfies Record<SaccharideType, SaccharideShape>;
 
 export function getSaccharideShape(type: SaccharideType, ringMemberCount: number): SaccharideShape {
```
arose, line 83: lets remove this function and drop specializing on ring count, it is non standard

src/mol-repr/structure/visual/carbohydrate-symbol-mesh.ts
```diff
@@ -75,85 +89,95 @@ function createCarbohydrateSymbolMesh(ctx: VisualContext, structure: Structure,
         builderState.currentGroup = i * 2;
 
         switch (shapeType) {
-            case SaccharideShape.FilledSphere:
-                addSphere(builderState, center, radius, detail);
-                break;
-            case SaccharideShape.FilledCube:
-                Mat4.scaleUniformly(t, t, side);
-                MeshBuilder.addPrimitive(builderState, t, box);
+            case SaccharideShape.FilledSphere: // e.g. 3d11
```
arose, line 92: thanks for adding example IDs, can you please add them to `docs\docs\misc\interesting-pdb-entries.md`, too.

CHANGELOG.md
```diff
@@ -39,6 +39,10 @@ Note that since we don't clearly distinguish between a public and private interf
 - Fix camera reset handling for (temporary) empty scenes (#1903)
 - Remove `firstStepSize` tracing parameter, derive automatically
 - Add `.parseRaw` to `DataFormatProvider` for out of state tree parsing
+- Carbohydrate symbols
+  - All carbohydrate symbols are rendered with 2 groups (primary and secondary) and can be potentially colored in two colors
+  - CarbohydrateSymbolColorTheme decides which shape will be colored by one or two colors
+  - Changed side length ratio of FlatBox shape from 2:2:1 to 2:1:1
```
arose, line 45: 2:1:1 is good, thanks

arose (changes requested):

src/mol-model/structure/structure/carbohydrates/constants.ts
midlik: This is to distinguish between:
- known monosaccharides with assigned color and with Flat Hexagon shape according to the docs you linked (i.e. Bac | LDmanHep | Kdo | Dha | DDmanHep | MurNAc | MurNGc | Mur) -> `Other`
- truly unknown saccharides, which we render as white prism based on ring count -> `Unknown`

midlik: Oh, now i'm reading your next comment below.
So you are suggesting merging back the `Other` and `Unknown` categories, but always rendering them as flat hexagon?
@arose please confirm if I'm understanding you correctly

midlik: OMG. I read the standard again, and i'm trying to wrap my head around it, especially notes 10 and 11. This is how I read it:

```
if monosaccharide present in the table:
    return the shape and color from the table
elif we can classify monosaccharide into a generic type (Hexose, Deoxyhexose, Pentose etc.):
    return shape of the generic type, white
    // This is not the same as prism based on ring count!
    // e.g. Hexose -> circle, Deoxyhexose -> triangle, Pentose -> star
elif we can at least identify what the monosaccharide is:
    return pentagon, white
else:
    // We cannot fully identify the monosaccharide
    return flat hexagon, white
```

I'm not 100% sure about the difference between the last two cases (white pentagon vs white flat hexagon), but I think white flat hexagon will actually never apply in our case, because we have the structure so we have fully identified the monosaccharide.

midlik: Also, I imagine that the second case (shape of the generic type, white) will be tricky to implement as it needs more than just the ring count.

So we could just do this:

```
if monosaccharide present in the table:
    return the shape and color from the table
else:
    // Life is too short
    return pentagon, white
```

src/mol-repr/structure/visual/carbohydrate-symbol-mesh.ts
midlik: Done.

src/mol-model/structure/structure/carbohydrates/constants.ts
arose: I would go with, and that is how I understood Oliver:

```
if monosaccharide present in the table:
    return the shape and color from the table
else:
    // Life is too short -> Unknown
    return flat hexagon, white
```

midlik: OK, i merged `Other` and `Unknown` back to just `Unknown` and it's always rendered as flat hexagon.

src/mol-model/structure/structure/carbohydrates/constants.ts
midlik: Done.

src/mol-geo/geometry/mesh/builder/sphere.ts
```diff
@@ -1,30 +1,32 @@
...
+ * @author Adam Midlik <midlik@gmail.com>
  */
 
 import { Vec3, Mat4 } from '../../../../mol-math/linear-algebra';
 import { MeshBuilder } from '../mesh-builder';
 import { Primitive } from '../../../primitive/primitive';
 import { Sphere } from '../../../primitive/sphere';
 
-const sphereMap = new Map<number, Primitive>();
+const sphereCache = new Map<string, Primitive>();
```
arose: let's keep this numeric, string is very expensive

src/mol-geo/geometry/mesh/builder/sphere.ts
```diff
@@ -1,30 +1,32 @@
...
 const tmpSphereMat = Mat4.identity();
 
 function setSphereMat(m: Mat4, center: Vec3, radius: number) {
     return Mat4.scaleUniformly(m, Mat4.fromTranslation(m, center), radius);
 }
 
-export function getSphere(detail: number) {
-    let sphere = sphereMap.get(detail);
+export function getSphere(detail: number, options?: { subset: 'ring' | 'caps' | undefined }) {
+    const cacheKey = `${detail}/${options?.subset ?? ''}`;
```
arose: maybe cantorPair? or just bitshift since detail will never be large

arose: optional arg is also not good for perf. probably best to have a special getSphereSubset func

arose (changes requested):
thanks @midlik - carb changes look good, just one perf thing when building sphere mesh

src/mol-geo/geometry/mesh/builder/sphere.ts
midlik: I reverted it to `Map<number, Primitive>`, and I store sphere subsets on negative keys (I assume `detail >= 0`).

src/mol-geo/geometry/mesh/builder/sphere.ts
midlik: I removed optional arg and split the function into `getSphere` and `getSphereSubset`.

(Note: `Sphere` still has an optional arg, but it only gets called once per `detail` value, so I assume it's OK.)

dsehnal (approved):
Looks good and non-break, if there are any issues left, let's create a followup fix PR.

PR #1916, opened 2026-08-14
MVS: vtp/ply/obj format support via a new `shape` node

corredD:
# Description

Adds a `shape` node so a MolViewSpec file can render a mesh from a 3D geometry resource, with new `vtp`, `ply` and `obj` parse formats. Previously molstar could read all three at plugin level but none was reachable from MVS, so a story had to convert meshes to inline `primitives` vertex arrays offline.

The node applies the same `ShapeFrom{Vtp,Ply,Obj}` + `ShapeRepresentation3D` transformers as opening the file through the UI and only supplies the initial parameter values, exactly as `volume` does. An MVS-loaded mesh is therefore the same state object as a UI-loaded one and keeps the full interactive coloring controls; format dispatch is on the parent `Format` object, as `volume` dispatches on Ccp4/Dx/Cif.

The node itself has no parameters. Color comes from child `color` nodes, and `opacity`, `clip`, `transform`, `instance` and `focus` accept a `shape` parent. Format-specific options travel as prefixed custom properties rather than spec parameters: `vtp_attribute` / `vtp_attribute_source` / `vtp_palette` / `vtp_domain`, `ply_coloring`, `obj_coloring` / `obj_material_colors`, plus the existing `molstar_mesh_params`.

`colorTheme`/`coloring` is always set explicitly and never left to the provider's file-derived default, so the same MVSJ renders the same way regardless of which data arrays the resource happens to contain.

See [MVS#111](https://github.com/molstar/mol-view-spec/pull/111)

## Actions


src/extensions/mvs/tree/molstar/conversion.ts
```diff
@@ -156,6 +160,10 @@ const StructureFormatExtensions: Record<ParseFormatT, (FileExtension | '*')[]> =
     map: ['.map', '.ccp4', '.mrc', '.mrcs'],
     dx: ['.dx'],
     dxbin: ['.dxbin'],
+    // geometry
```
dsehnal:
```suggestion
    // shape
```

and elsewhere too

src/extensions/mvs/load-helpers.ts
```diff
@@ -117,6 +118,144 @@ function applyTransformAndInstances(target: UpdateTarget, node: MolstarSubtree,
...
+     * names to pre-fill them. Materials not listed render grey (the provider's own fallback). */
+    obj_material_colors?: Record<string, string>,
+    /** Mesh-level rendering params, as for `primitives`. */
+    molstar_mesh_params?: Record<string, any>,
+}
+
+/** Color for a `shape` node, taken from its `color` children (last one wins), as `volume` does. */
+function shapeColor(node: MolstarSubtree<'shape'>): Color {
+    const children = getChildren(node).filter(c => c.kind === 'color') as MolstarNode<'color'>[];
+    const color = children.length > 0 ? children[children.length - 1].params.color : DefaultColor;
```
dsehnal: Move the private shape-related code a separate file please (say helpers/load-shape.ts), the file is getting too big.

dsehnal (commented):
This looks quite straightforward, well done Ludo and Claude :D

One more thing to explore before we merge this. You mentioned that in the future the VTK will result in multiple shape types. Can we explore that we will be able to support this with this approach? Ideally adding the support for the composite shape provider and a VTK transform that uses it. Can be a separate draft PR. CC @arose here to weight in on the direction.

corredD:
Thanks @dsehnal 
Both comments addressed. The shape node has no parameters so it should not be breaking. The blocker is the parser, but same problem in other formats that can carry different primitive types.

@arose, does a composite `ShapeProvider` belong in `mol-model/shape`, or would you rather keep providers single-geometry and compose higher up?

Happy to look into it in a separate PR, and extend the VTP to support more primitives to help explore this.

dsehnal:
> Thanks @dsehnal Both comments addressed. The shape node has no parameters so it should not be breaking. The blocker is the parser, but same problem in other formats that can carry different primitive types.
> 
> @arose, does a composite `ShapeProvider` belong in `mol-model/shape`, or would you rather keep providers single-geometry and compose higher up?
> 
> Happy to look into it in a separate PR, and extend the VTP to support more primitives to help explore this.

A possible way to do this might be

```ts
export interface ShapeProvider<D, G extends Geometry, P extends Geometry.Params<G>> {
    label: string
    data?: D
    params?: P
    getShape?: ShapeGetter<D, G, P>
    geometryUtils?: GeometryUtils<G>
    
   shapes?: {
        data?: D
        params?: P
        getShape?: ShapeGetter<D, G, P>
        geometryUtils?: GeometryUtils<G>
    }
}
```

or add new

```ts
export interface ShapeGroupProvider {
   label: string
   shapes: ShapeProvider<any>[]
}
```

-------

I think ``ShapeGroupProvider`` is a clearer option.

arose:
> Thanks @dsehnal Both comments addressed. The shape node has no parameters so it should not be breaking. The blocker is the parser, but same problem in other formats that can carry different primitive types.
> 
> @arose, does a composite `ShapeProvider` belong in `mol-model/shape`, or would you rather keep providers single-geometry and compose higher up?
> 
> Happy to look into it in a separate PR, and extend the VTP to support more primitives to help explore this.

Depends a bit on the use case. We do have, e.g., multiple blocks in cif files or cube files with both a structure and a volume and that works fine.

I can see a use case for "composite" shape objects. There is also your other pending PR to add "external" volume coloring to vtp which should be generalized to Shapes. I think this warrants some time to design Shape for more use cases.

dsehnal (approved):

PR #1917, opened 2026-08-20
Surface representation fixes: traceOnly, floodfill, GPU/CPU handoff, blob wireframe

giagitom:
Eight fixes to the surface representations, found while tracking down a `traceOnly` bug in the molecular surface wireframe. Each commit stands alone.

**Bugs**

- `traceOnly` was ignored by both molecular surface **wireframe** visuals: it was missing from `setUpdateState`, so toggling it never invalidated the geometry and the surface kept using every atom. The mesh visuals and both gaussian wireframes already had the check.
- `floodfill` was exposed on the gaussian surface **wireframe**, and rebuilt the geometry when changed, but was never applied — the file did not even import `Tensor`.
- The CPU surface/volume visuals' `mustRecreate` checked fewer conditions than the dispatcher that selects between the GPU and CPU visual. Whenever a missing condition held, `mustRecreate` returned true, the visual was destroyed, and the dispatcher handed back the *same* CPU visual — a full re-march on every update, with the render object recreated and marker state lost. Reachable through `includeParent`, `floodfill`, or simply a context without `OES_texture_float_linear` (not core in WebGL2), where the default gaussian path was permanently in this state. `volume/segment.ts` was worst: every segment mesh re-marched together.
- The blob surface density never yielded — `Task.create` accepted a `RuntimeContext` and discarded it, so it blocked the main thread with no progress and could not be cancelled.

**Cleanups**

- `floodfill` removed from the gaussian density volume, where it was exposed but meaningless: that visual renders a `DirectVolume` from a GPU texture, so there is no scalar field to fill and no iso threshold to fill against.
- The GPU/CPU choice is now one `useGpu` predicate per file, used by the dispatcher and both `mustRecreate` bodies, replacing twelve hand-maintained copies. This is why the drift above was possible twice.

**Feature**

- `blob-surface-wireframe` / `structure-blob-surface-wireframe`, so the blob surface can be drawn as a wireframe like the other two. The blob density already produced everything `Lines` needs; the shared 13-field geometry predicate moved to `util/blob-surface.ts` rather than being duplicated.

Verified with `tsc --noEmit` and `eslint` clean at every commit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,12 @@ All notable changes to this project will be documented in this file, following t
 Note that since we don't clearly distinguish between a public and private interfaces there will be changes in non-major versions that are potentially breaking. If we make breaking changes to less used interfaces we will highlight it in here.
 
 ## [Unreleased]
-- Fix the blob surface density blocking the main thread with no progress and no cancellation (the `RuntimeContext` was accepted by the task and then discarded)
-- Add `blob-surface-wireframe` and `structure-blob-surface-wireframe` visuals, so the blob surface can be drawn as a wireframe like the gaussian and molecular surfaces
-- Remove the `floodfill` param from the gaussian density volume representation, where it was exposed but had no meaning (DirectVolume has no scalar field to fill and no iso threshold to fill against)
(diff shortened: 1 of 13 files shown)


arose (approved):
made some tweaks, otherwise good - thank you

giagitom:
thanks @arose, is this already mergiable?

arose:
It is. Just wanted to let you look at my changes before.

giagitom:
Sure thanks. It looks good to me.

PR #1918, opened 2026-08-23
Viewer app: keep track of viewer instances in global

arose:
# Description

Useful to have the viewer instances as a global sometimes

## Actions


src/apps/viewer/app.ts
```diff
@@ -32,12 +32,15 @@ export { consoleStats, isDebugMode, isProductionMode, isTimingMode, setDebugMode
 // TODO: consider removing these in v6.0
 export type { LoadStructureOptions, LoadTrajectoryParams, VolumeIsovalueInfo } from '../../extensions/plugin/loaders';
 
+export const viewerInstances: Viewer[] = [];
```
dsehnal: I would move this to a static field on the viewer class instead.

Eg Viewer.instances

arose: yeah, that's better

PR #1920, opened 2026-08-23
Fix illumination indirect light and make glow direction aware

giagitom:
Three fixes to the illumination pass.

**Indirect light was not in the same space as the direct term.** A ray escaping on the first bounce carries `tShaded`, which is `clamp(outgoingLight, 0.01, 0.99) * uExposure`. Later bounces carried raw albedo times an absolute light energy, unclamped and unexposed. The default ambient `0.4` plus light `0.6` sums to exactly `1.0`, so the two matched by coincidence; above that the direct term saturates while the indirect one keeps climbing, and occluded pixels end up brighter than lit ones.

**`glow` had no direction term.** It gave every bounced ray the full light, brightening occluded pixels while exposed ones kept their own shading — ambient occlusion with the sign flipped, worst under a back light. The light is now sampled with the escaping ray direction instead of the surface normal, so glow softens the lit/unlit boundary and vanishes where no light reaches.

**`NaN` in the direct light shadow normalization** when `ambientIntensity` and light `intensity` are both zero, which the `0.0` minimum on both allows.

Escape colour before throughput, measured at ambient 0, light 3, exposure 2, albedo 0.8:

| `dot(N, L)` | before | after | glow off |
| --- | --- | --- | --- |
| 1 | 1.98 | 1.98 | 1.98 |
| 0 | 1.98 | 1.04 | 0.00 |
| -1 | 1.98 | 0.00 | 0.00 |

Defaults are unchanged: at ambient 0.4, light 0.6, exposure 1 the output is identical.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


CHANGELOG.md
```diff
@@ -51,6 +51,9 @@ Note that since we don't clearly distinguish between a public and private interf
 - MolViewSpec
   - Added `transition` node with params `duration_ms`, `trajectory`, `easing`
   - Snapshot metadata: `linger_duration_ms` renamed to `duration_ms`, deprecated `transition_duration_ms`
+- Fix `NaN` in illumination direct light shadows when ambient and light intensity are both zero
```
arose: what is the actual visual effect of this?

CHANGELOG.md
```diff
@@ -51,6 +51,9 @@ Note that since we don't clearly distinguish between a public and private interf
...
+- Fix illumination indirect light ignoring `exposure` and the shading clamp, inverting shading at high light intensity
+- Weight illumination `glow` by the direction bounced rays escape in, so it softens the lit/unlit boundary instead of brightening occluded areas
```
arose: I am not sure it is worthwhile to keep this modified glow, the effect is so little. I am ok with removing the artificial effect altogether.

CHANGELOG.md
```diff
@@ -51,6 +51,9 @@ Note that since we don't clearly distinguish between a public and private interf
 - MolViewSpec
   - Added `transition` node with params `duration_ms`, `trajectory`, `easing`
   - Snapshot metadata: `linger_duration_ms` renamed to `duration_ms`, deprecated `transition_duration_ms`
+- Fix `NaN` in illumination direct light shadows when ambient and light intensity are both zero
+- Fix illumination indirect light ignoring `exposure` and the shading clamp, inverting shading at high light intensity
```
arose: good fix but why call it inverting?

arose (changes requested):

CHANGELOG.md
giagitom: To reproduce the issue this is necessary:

- illumination.enabled: true
- illumination.shadowEnable: true
- renderer.ambientIntensity: 0
- renderer.light[0].color: a hex with a 00 byte, e.g. #ff8800

**master**
image

**fixed (same as non-illumination)**
image

CHANGELOG.md
giagitom: Agreed, removed the glow in next commit

CHANGELOG.md
giagitom: Better wording in last commit.

CHANGELOG.md
arose: ahh, ok, just one channel 0, thanks

arose (approved):
thanks!

arose:
PSA: to get some glow it is best to add a tiny amount of emissiveness as needed

image

Issue #1145, opened 2024-05-29
unsafe-eval required when setting up CSP header


cmonnom:
Is there any chance to fix the code being blocked by the CSP in future releases?
https://web.dev/articles/csp?utm_source=devtools#unsafe-eval

We'd like to use Molstar but it goes against our SOC2 policies.

It would require removing instances of `eval(), new Function(), setTimeout([string], …), and setInterval([string], ...)`

🥺 🙏

dsehnal:
Yes, it would be possible. There are only few cases of `new Function`:

- [setImmediate polyfill](https://github.com/molstar/molstar/blob/master/src/mol-task/util/scheduler.ts#L34) -- can be just removed as we don't use the string callback here
- [string interpolate function](https://github.com/molstar/molstar/blob/5af6a3e967937514a76783946046c272ffed6b2b/src/mol-util/string.ts#L69) 
  - [Binding.format](https://github.com/molstar/molstar/blob/5af6a3e967937514a76783946046c272ffed6b2b/src/mol-util/binding.ts#L51): this could be replaced by a simple `.replace('${triggers}', ...)`
  - [Swagger UI](https://github.com/molstar/molstar/blob/master/src/servers/common/swagger-ui/index.ts#L28): this could be modified to use basic string "replace all" too.
- [model](https://github.com/molstar/molstar/blob/master/src/servers/model/config.ts#L254) and [volume](https://github.com/molstar/molstar/blob/master/src/servers/volume/server/web-api.ts#L54) server id -> path mapping. Could be implemented without both `Function` and `interpolate`

Perhaps you can make a PR with it?

arose:
@cmonnom care to do a PR?

arose:
> @cmonnom care to do a PR?

still happy to accept a PR

TaylorHo:
I've opened [this PR](https://github.com/molstar/molstar/pull/1922) to address the `new Function` usage. Can someone please review?

PR #1922, opened 2026-08-28
Restricts the use of eval and new function

TaylorHo:
# Description

Addresses [#1145](https://github.com/molstar/molstar/issues/1145) - Mol* currently requires `unsafe-eval` in Content Security Policy headers, which prevents adoption under SOC 2 policies.

This PR removes all first-party `new Function` usage (the only eval-equivalent patterns found in source). No `eval()`, string-based `setTimeout`/`setInterval`, or other dynamic code execution remains.

Key changes:
- **`interpolate()`** - replaced `new Function`-based template evaluation with safe `${key}` string substitution (used by binding help text and Swagger UI HTML generation).
- **`setImmediate` polyfill** - removed string-callback branch that compiled callbacks via `new Function`.
- **ESLint** - added `no-new-func` and `no-eval` as errors to prevent regressions.
- **Server path/id mapping** - replaced dynamically generated `new Function` switch statements in model and volume server config with `Map` lookup plus **`interpolateIdPath()`**, which uses a **whitelist of allowed string methods** on `id`: `toLowerCase()`, `toUpperCase()`, `substr()`, `substring()`, and `slice()`. Unsupported `${…}` expressions are rejected at startup.

After this change, the viewer bundle should work without `'unsafe-eval'` in CSP.

## Reviewer note: stricter server path/id templates

Model and volume server `--sourceMap` / `--idMap` path templates no longer evaluate arbitrary JavaScript inside `${…}`. That behavior came from `new Function`, which was removed.

Templates now go through `interpolateIdPath()` with a **whitelist** of allowed expressions on `id`:

- `id`
- `id.toLowerCase()`, `id.toUpperCase()`
- `id.substr(n, m)` / `id.substring(n, m)` (two numeric args; length semantics)
- `id.substring(n)` (one numeric arg)
- `id.slice(n)` / `id.slice(n, m)` (numeric args only)

**What may break:** configs that relied on ad-hoc JS in `${}` will fail at **server startup** (via `validateIdPathTemplate`) or at runtime with `Unsupported id path expression`. Examples that worked before but are rejected now:

- Other string methods (`replace`, `split`, `trim`, etc.)
- Chained calls (`id.toLowerCase().substr(1, 2)`)
- Non-literal arguments (`id.substr(id.length - 2, 2)`, variables, arithmetic)
- Any general JS expression inside `${}`

**What should still work:** documented patterns like `./data/${id.substr(1, 2)}/${id}.bcif` and simple `${id}` substitution.

If the reviewers think this may be a problem, we can make it less strict or even reconsider removing the `new Function` in server path/id mapping.

## Actions

- [x] (Optional but encouraged) Improved documentation in `docs`.


dsehnal (approved):
LGTM, thanks

PR #1923, opened 2026-09-01
MolQL selectors for MolViewSpec

dsehnal:
# Description

Adds experimental/proof-of-concept support for serialized MolQL expressions in MolViewSpec selectors.

Corresponding Python version https://github.com/molstar/mol-view-spec/pull/112

MolQL expressions use an explicit `molql` wrapper to distinguish them from existing named and schema-based selectors:

```json
{
  "molql": "<MolQL JSON expression>"
}
```

The wrapper shape and serialized expression tree are validated by MVS. The top-level value must be a MolScript application, and the expression is compiled before execution; validation results are cached by expression identity.

This PR adds support for wrapped MolQL expressions in:

- Component selectors.
- Selective color layers, including `color`, `color_from_uri`, and `color_from_source` selector parameters.
- Primitive positions, with optional `structure_ref` support.
- Component-derived labels and tooltips through the existing selector pipeline.

Color-layer selectors are evaluated against the root structure and then intersected with the represented component. This allows context-dependent queries, such as selecting polymer residues around a ligand, to work when the representation itself contains only the polymer subset.

Primitive positions also support direct schema-based addresses with `structure_ref`, alongside the existing explicit `expressions` list form. Existing named selectors, component-expression rows, and explicit row lists retain their current behavior.

A new four-step MVS story demonstrates:

1. Component selections created with `MolScriptBuilder` and the PyMOL transpiler.
2. MolQL selectors used for multilayer coloring on a polymer-only representation.
3. Reusing a MolQL component for representations, labels, and tooltips.
4. A primitive distance measurement between imatinib N13 and Thr315 OG1, with both endpoints selected using wrapped MolQL expressions.

The story loads the first biological assembly of PDB entry 1IEP and does not prescribe fixed camera positions.

## Testing

```bash
npm run dev -- -e mvs-stories
```

Open [the MolQL MVS story](http://localhost:1338/build/examples/mvs-stories/?story=molql).

## Actions

- [x] Contributor is already listed in `package.json`

### DIFF
--- a/src/extensions/mvs/tree/mvs/param-types.ts
+++ b/src/extensions/mvs/tree/mvs/param-types.ts
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2023-2025 mol* contributors, licensed under MIT, See LICENSE file for more info.
+ * Copyright (c) 2023-2026 mol* contributors, licensed under MIT, See LICENSE file for more info.
  *
  * @author Adam Midlik <midlik@gmail.com>
  * @author David Sehnal <david.sehnal@gmail.com>
--- a/src/extensions/mvs/tree/mvs/param-types.ts
+++ b/src/extensions/mvs/tree/mvs/param-types.ts
@@ -134,15 +134,15 @@ export interface ComponentExpressionT extends ValueFor<typeof _ComponentExpressi
 export const ComponentExpressionT = new iots.Type<ComponentExpressionT>(
     'ComponentExpression',
     isComponentExpression,
-    (value, context) => isMolQLExpression(value)
-        ? iots.failure(value, context, 'MolQL expression wrappers must contain a valid expression')
+    (value, context) => hasMolQLProperty(value) || hasStructureRefProperty(value)
+        ? iots.failure(value, context, 'MolQL wrappers and primitive structure_ref positions are not component expressions')
         : _ComponentExpressionT.validate(value, context),
     value => value,
 );
 
 /** A MolQL expression serialized as JSON. The expression is validated by the MolScript compiler. */
 export interface MolQLExpressionT {
-    expression: unknown
+    molql: unknown
 }
 export const MolQLExpressionT = new iots.Type<MolQLExpressionT>(
     'MolQLExpression',
@@ -186,13 +186,11 @@ const _PrimitiveComponentExpressionT = partial({
     expressions: list(ComponentExpressionT),
 });
 /** Primitives-related types */
-export interface PrimitiveComponentExpressionT extends ValueFor<typeof _PrimitiveComponentExpressionT> { }
+export interface PrimitiveComponentExpressionT extends ComponentExpressionT, ValueFor<typeof _PrimitiveComponentExpressionT> { }
 export const PrimitiveComponentExpressionT = new iots.Type<PrimitiveComponentExpressionT>(
     'PrimitiveComponentExpression',
-    isPrimitiveComponentExpressions,
-    (value, context) => isMolQLExpression(value)
-        ? iots.failure(value, context, 'MolQL expression wrappers must contain a valid expression')
-        : _PrimitiveComponentExpressionT.validate(value, context),
+    isPrimitiveComponentExpression,
+    (value, context) => validatePrimitiveComponentExpression(value, context),
     value => value,
 );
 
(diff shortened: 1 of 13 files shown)


src/extensions/mvs/tree/mvs/param-types.ts
```diff
@@ -129,7 +131,25 @@ const _ComponentExpressionT = partial({
...
+);
+
+/** A MolQL expression serialized as JSON. The expression is validated by the MolScript compiler. */
+export interface MolQLExpressionT {
+    expression: unknown
+}
+export const MolQLExpressionT = new iots.Type<MolQLExpressionT>(
+    'MolQLExpression',
+    isMolQLExpression,
+    (value, context) => validateMolQLExpression(value, context),
```
midlik, line 150: I tried `cartoon.color({ color: 'orange', selector: { expression: 'blabla' } });`
It passes the validation step but fails after with uncaught error.
Is this expected? Perhaps it would be good to use stricter validation.

src/extensions/mvs/tree/mvs/param-types.ts
```diff
@@ -237,11 +275,59 @@ export function isVector3(x: any): x is Vector3 {
...
+    if (!isMolQLExpression(value)) {
+        return iots.failure(value, context, 'Expected an object with an expression property');
+    }
+    if (primitive && !isPrimitiveMolQLExpression(value)) {
+        return iots.failure(value, context, 'Primitive MolQL structure_ref must be a string when provided');
+    }
+    const validation = validateMolQLSyntax(value.expression);
+    return validation === true
+        ? iots.success(value)
+        : iots.failure(value, context, validation);
```
midlik: Could you please also replace any current usage of `{ _tag: 'Right', right: ... }`, `{ _tag: 'Left', left: ...}` by `iots.success`, `iots.failure`; to keep things consistent?

In param-types.ts and field-schema.ts.

midlik (approved):
Seems good and this will be very useful.

midlik:
Please consider changing the name of the field to `molql`.
So instead of

```js
    cartoon.color({ color: 'orange', selector: { expression: ... } });
```

we should have 

```js
    cartoon.color({ color: 'orange', selector: { molql: ... } });
```

I believe this will greatly improve code readability and avoid confusion between this and `expressions` which we already have in selectors for  primitives.

Screenshot 2026-09-02 at 10 40 18

midlik:
Also, there is a slight asymmetry between ComponentExpressionT selectors and MolQL when it comes to primitives with structure references.

```js
const prim = builder.primitives({ opacity: 0.25, color: 'yellow' });

prim.sphere({ center: { structure_ref: 'struct1', label_asym_id: 'A' } }); // no TS error, fails runtime (Error: Structure with ref '<default>' not found.)
prim.sphere({ center: { structure_ref: 'struct1', expressions: [{ label_asym_id: 'A' }] } }); // OK

prim.sphere({ center: { structure_ref: 'struct1', expression: molql } }); // OK
prim.sphere({ center: { structure_ref: 'struct1', expressions: [{ expression: molql }] } }); // TS error
```

No a big deal, just to make sure you're aware of this.

midlik:
This shows no TS error and passes MVS validation:
```js
    cartoon.color({ color: 'orange', selector: { expression: molql, label_seq_id: 100 } });
```

but ignores the `label_seq_id` part of the selector. 
Could we employ stricter validation to avoid bugs caused by developers mixing MolQL and ComponentExpressionT selectors?
I imagine this would be pretty tricky to achieve at TS level, but at least at MVS validation level this should throw an error.

---

**Alternative solution:**

Stringify MolQL expressions instead of including it as JSON object. We could also add a prefix to the stringified expression to avoid confusion with static selectors and get better typing checks and intellisense.

```ts
export type MolQLExpressionT = `molql:${string}`;
export const MolQLExpressionT = new iots.Type<MolQLExpressionT>(...);

const SelectorT = union(ComponentSelectorT, ComponentExpressionT, list(ComponentExpressionT), MolQLExpressionT);

cartoon.color({ color: 'orange', selector: 'ligand' }); // OK
cartoon.color({ color: 'orange', selector: 'ligadn' }); // TS error
cartoon.color({ color: 'orange', selector: 'molql:blablabla' }); // OK
```

dsehnal:
@midlik molql expressions don't have strong typing, so it won't should it for the current version. would implement this in a separate pass.

for other issues i agree.

PR #1924, opened 2026-09-04
Fix CCP4/MRC volumes with unset cell angles and uncomputed rms

corredD:
# Description

Some volume data from IMOD or membrain-seg has `cellb` (α, β, γ) = 0, 0, 0 — it happens with `clip flipx` from IMOD. Zero angles make `Cell.create` compute `z2 = (cos α − cos β·cos γ) / sin γ` = `0/0` = `NaN`, so the fractional transform is NaN, the volume's bounding sphere is NaN, and the instance grid eventually throws `RangeError: Invalid typed array length: -Infinity` in the browser.

Changes:

1. `getCcp4Angles` in `ccp4.ts` guards each angle to (0, 180)°, falling back to 90° with a warning. Placed in the CCP4 reader rather than in `Cell.create`, because `Cell.empty()` is `create(Vec3(), Vec3())` — zero angles are a legitimate sentinel there, also used by `Cell.fromBasis`, so a warn-and-fallback in `create` would fire on a normal path. It mirrors `getCcp4Size`, which already treats zero cell lengths as unset, and keeps the blast radius inside CCP4/MRC so no structure format is affected.
2. A second bug the first fix exposed: these headers also carry `rms = -1`, which by MRC convention means *not computed*. `ccp4.ts` only recomputed sigma when it was exactly 0, so `sigma = -1` was taken literally — and since `isoValue = relativeValue · sigma + mean`, the default 2σ level landed at −1.998, which is not in the volume value range in the case of membrain. The isosurface would have degenerated to the whole box. Now a non-positive rms is recomputed.

Verified against the example file below: with the raw header angles it throws `RangeError: Invalid typed array length: -Infinity`; with these two fixes the same volume gives a finite `fromFractional`/`toFractional`, sigma is recomputed to 0.0761 (was −1), and the default 2σ iso level lands at 0.155, inside the volume's [0, 2] range.

Example file [here](https://drive.google.com/file/d/17qUXIGK8TPoRLGvu6R3wEDLWjdFmfi38/view?usp=sharing)

## Actions


src/mol-math/geometry/lookup3d/grid.ts
```diff
@@ -240,6 +240,12 @@ function build(data: PositionData, boundary: Boundary, cellSizeOrCount?: Vec3 |
         size = [1, 1, 1];
     }
 
+    // guard against degenerate grids, e.g. when the boundary is empty or has non-finite values
+    if (!size.every((s: number) => Number.isFinite(s) && s >= 1)) {
+        size = [1, 1, 1];
+        delta = [1, 1, 1];
+    }
```
dsehnal: Can it happen than only 1 dimension is generate? in that case, only the degenerate dimensions should get the default value?

src/mol-math/geometry/lookup3d/grid.ts
```diff
@@ -240,6 +240,12 @@ function build(data: PositionData, boundary: Boundary, cellSizeOrCount?: Vec3 |
         size = [1, 1, 1];
     }
 
+    // guard against degenerate grids, e.g. when the boundary is empty or has non-finite values
```
arose: not sure we really want this, it is just masking a serious problem upstream. my vote is for not guarding.

arose (changes requested):

src/mol-math/geometry/lookup3d/grid.ts
corredD: Agreed — dropped in 04cd91db7, so `grid.ts` is untouched by this PR now.

For the record, it really was one bad input all the way down: `cellb = 0` makes `Cell.create` compute `z2 = (cos α − cos β·cos γ) / sin γ` = `0/0` = `NaN`, so `fromFractional` is NaN (and `Mat4.invert` logs `non-invertible matrix.`) → the volume's bounding sphere is NaN → `calcBottomGrid` buckets nothing, because `grid[NaN] += 1` never equals 1, and returns `cellCount === 0` → `calcTopGrid` then loops zero times over an untouched `Box3D.setEmpty()` box, so `S = -MAX_VALUE - MAX_VALUE = -Infinity`, `size = ceil(-Infinity / batchSize)`, and `new Uint32Array(-Infinity)` throws.

Checked against the example file with only the two CCP4 commits applied: with the raw header angles it throws `RangeError: Invalid typed array length: -Infinity`; with `getCcp4Angles` the same volume gives a finite `fromFractional`/`toFractional` and `cellCount = 1`. `updateTransformData` already guards `instanceCount > 0`, so there is no other route to a zero-cell top grid today.

src/mol-math/geometry/lookup3d/grid.ts
corredD: Yes, it can — in the `cellSize` branch `S` is divided component-wise and `Vec3.isZero` only catches an all-zero `cellSize`, so a single bad axis can go non-finite on its own. Your point stands; the blanket `[1,1,1]` was too coarse.

That said, I've dropped the guard entirely per @arose. The `RangeError` here came from a NaN cell transform, which the CCP4 angle fix removes at the source, so `grid.ts` is no longer touched by this PR. Happy to revisit as a separate change if a real repro turns up.

arose (approved):

PR #1928, opened 2026-09-07
Add Volume Segmentor for interactive multi-body volume segmentation

Poko18:
## Summary

Adds the Volume Segmentor extension and standalone example for interactively segmenting a volume into multiple body masks. Bodies are defined by polygons drawn from multiple camera views; list order resolves overlaps and an optional remainder body captures unassigned voxels. Labels recompute from definitions, the source isosurface uses body-label coloring, and the UI supports mask previews, dust removal, per-body extend/cosine soft edges, and MRC export with the largest body first.

This is very useful for creating volume segments for multi-body refinement.

The implementation also adds an exact 3D Euclidean distance transform to `mol-math`, fixes `CCP4Writer.writeMrc` for non-canonical axis orders, and adds `volume-segmentor` / `VolumeSegmentor`.

## Demo

[Open the Volume Segmentor demo](https://poko18.github.io/molstar)

https://github.com/user-attachments/assets/ae6eb17b-5cbe-4ca7-987b-5714825ad55f

## Verification

- 27 focused Jest tests pass across the extension and distance transform.
- ESLint passes on the changed source.
- Production example build passes for `volume-segmentor`.
- ESM TypeScript compilation passes; the repository CommonJS build remains blocked by the existing `tsconfig.commonjs.json` / `moduleResolution: bundler` configuration (`TS5095`).

## Actions


src/mol-io/writer/ccp4/ccp4.ts
```diff
@@ -12,9 +12,30 @@ import { Vec3 } from '../../../mol-math/linear-algebra';
...
+            space.getCoords(o, c);
+            out[c[0] + c[1] * nx + c[2] * nx * ny] = data[o];
+        }
+        return out;
+    }
+
     export function writeMrc(grid: Grid, data: Uint8Array | Float32Array): ArrayBuffer {
         const [nx, ny, nz] = grid.cells.space.dimensions as [number, number, number];
         const voxelCount = nx * ny * nz;
+        data = toCanonicalOrder(grid, data);
```
dsehnal:
```suggestion
        toCanonicalOrder(grid, data);
```

no need to re-assign it since you mutate it in place.

src/examples/volume-segmentor/index.html
```diff
@@ -0,0 +1,37 @@
+<!DOCTYPE html>
```
dsehnal, line 1: could you combine this with the volume-mask example UI instead of adding a separate example?

src/mol-io/writer/ccp4/ccp4.ts
Poko18: `toCanonicalOrder` did not mutate in place, but I simplified the voxel permutation and its now done while filling the output buffer, instead of building a canonical copy first

src/examples/volume-segmentor/index.html
Poko18: I’d actually prefer to keep them separate, even though they look similar.

Initially I wanted to just extend the volume-mask tool, but the UI was starting to get too complicated. Since the segmentor has somewhat different functionality, I decided to make it a separate extension and, following the existing repo convention, a separate example as well.

src/mol-io/writer/ccp4/ccp4.ts
dsehnal: Oh, in that case sorry for misleading you.

src/examples/volume-segmentor/index.html
dsehnal: I don't want the examples explode though and this is quite similar.

Maybe you can do this via top level "tabs" or something, so its one parent with two different tasks you can perform?

Maybe can rename the example for this purpose?

Poko18: Okay, both are now under `volume-tools`

dsehnal (commented):
Let's also merge this into a single `volume-tools/{mask, segmentor}` extension. 

This would follow the same pattern as `rcbs`, `model-archive` etc.

otherwise seems fine.

Poko18:
Makes sense, both are now under `volume-tools/{mask, segmentor}` and I also added a **Breaking** CHANGELOG line for the `extensions/volume-mask → extensions/volume-tools/mask` import path.

demo: https://poko18.github.io/molstar/

dsehnal (approved):

PR #1931, opened 2026-09-09
Mol* 6.0 Architecture Proposal

dsehnal:
# Description

Architecture proposal for breaking mol* into multiple packages.

See [summary](https://github.com/molstar/molstar/blob/v6-architecture/docs/v6-summary.md), [full version](https://github.com/molstar/molstar/blob/v6-architecture/docs/v6-architecture), and [WebGPU support](https://github.com/molstar/molstar/blob/v6-architecture/docs/v6-webgpu) (all documents are linked for easier inline feedback as PR review)

### DIFF
--- a/docs/v6-architecture.md
+++ b/docs/v6-architecture.md
@@ -10,6 +10,8 @@ Ship `@molstar/migrate-6` with the release to handle mechanical import changes a
 
 The release also includes a standalone MolViewSpec builder, dependency-cycle removal, maintainer skills, updated developer docs, and workspace CI.
 
+Establish a rendering-backend boundary in `@molstar/graphics` for future WebGPU and other targets, retaining WebGL as the working implementation. The [rendering-backend design](v6-webgpu.md) covers the blast radius, minimal contracts, migration, and validation; a production WebGPU renderer and feature parity are later work.
+
 Out of scope:
 
 - Independent package versions or a package per parser/representation.
@@ -485,6 +487,8 @@ Implement 6.0 on `main`, publish `dev` prereleases, then release stable after th
 ### 10.2 Technical phases
 
 Keep the major workstreams in separate PRs. Each phase ends with a working build; validate compiled packages and source-based app bundles throughout.
+
+Coordinate the [rendering-backend workstream](v6-webgpu.md#6-minimal-implementation-sequence) with dependency cleanup and packaging, and validate its contracts before freezing the v6 graphics API.
 
 | Phase | Work and exit condition |
 | --- | --- |
--- a/docs/v6-architecture.md
+++ b/docs/v6-architecture.md
@@ -97,8 +97,8 @@ Within the plugin layer:
 - Use `import type` for parser result types in `PluginStateObject`, and for `PluginContext` wherever only its type is needed. Extract a context interface where it helps isolate construction; its dependencies must remain type-only.
 - Separate the `PluginBehavior` contract from the classes extending `PluginStateObject`. Behavior detection uses `typeClass === 'Behavior'` without importing those classes.
 - Move default specs out of base spec/context modules into `@molstar/plugin/default-spec` and `@molstar/plugin-ui/default-spec`. Base construction takes an explicit spec.
-- Split transform modules into leaves. Leaves, builders, and managers import individual transformers, never the `StateTransforms` facade.
-- Rebuild that facade as a plain object in `@molstar/plugin/state/transforms` after removing cycles. Delete the lazy getters.
+- Split transform modules into leaves. All transformer consumers import individual defining modules.
+- Remove the `StateTransforms` convenience facade and its lazy getters; do not replace it with another aggregate object.
 
 Enforce an acyclic package graph and no value-import cycles within packages. Type-only cycles inside a package are allowed; they must not conceal a package cycle.
 
@@ -154,7 +154,7 @@ Use the same public package specifiers inside and outside the repository:
 
 ```ts
 // packages/graphics/src/canvas3d/passes/illumination.ts
-import { ValueCell } from '@molstar/core/util';
+import { ValueCell } from '@molstar/core/util/value-cell';
 import type { Texture } from '@molstar/graphics/gl/webgl/texture';
 ```
 
@@ -186,6 +186,19 @@ Each package declares every external package it directly imports. “Owned by co
 Use peers for React/React DOM on UI packages, and optional peers for headless dependencies (`gl`, `canvas`, `pngjs`, `jpeg-js`) and optional cloud storage where used. Do not add React to plugin or core.
 
 Types needed only to build a package belong in its dev dependencies. Types referenced by published declarations must be available to consumers through dependencies or declared peers. In particular, React does not install `@types/react`: UI packages need an explicit consumer-facing type dependency/peer policy, verified with a clean TypeScript consumer.
+
+### 4.4 No barrel files
+
+Do not ship convenience barrel files. Consumers, examples, and internal code import defining modules through supported subpaths. Package export maps provide those paths directly; no root or directory aggregation module is needed. This removes a source of accidental broad imports rather than relying on every caller to choose correctly.
+
+Barrels do not inherently prevent production tree-shaking: ESM bundlers can remove unused exports when usage and side effects permit it. They can still enlarge the graph processed by tooling and retain initialization effects. [Vite's warning](https://vite.dev/guide/performance#avoid-barrel-files) concerns extra development-time fetching and transformation; [esbuild's documentation](https://esbuild.github.io/api/#tree-shaking) describes conditional removal of unused code. The policy avoids the aggregation risk without assuming every barrel defeats every bundler.
+
+- **Remove aggregation modules**, including named re-export collections, `export *` chains, and type-only convenience barrels. Use `import type` from the defining module for types. Exporting locally defined symbols remains normal module structure.
+- **Preserve cohesive implementation entry points.** A file named `index.ts` is not inherently a barrel. Split mixed implementation/re-export modules such as `mol-util/index.ts` into suitable defining modules and record migrated paths. An export-map alias may point directly to an implementation file without adding a wrapper barrel.
+- **Do not substitute aggregate convenience objects.** Remove `StateTransforms` and migrate consumers to transformer leaves. Preserve transformer identifiers and required registration behavior, not the aggregate access syntax.
+- **Keep deliberate composition explicit.** Default specs and registration catalogs assemble complete feature sets for app composition. They are not general symbol-import entry points. Base runtime modules and individual providers must not depend on them; enforce that boundary even within a package.
+
+Enforce the policy through lint/import-graph checks and the public export inventory. Validate that a leaf import cannot reach unrelated providers, default catalogs, or optional backends through re-exports. Reuse the slim-plugin fixture to check both processed modules and production output, with a downstream Vite development smoke/profile case alongside esbuild. Audit real registration and asset side effects before adding purity annotations or `sideEffects: false`; no consumer barrel-rewriting plugin is required.
 
 ## 5. ESM and TypeScript source
 
@@ -257,9 +270,9 @@ A representative core export map:
   "engines": { "node": ">=22.0.0" },
   "exports": {
     "./task": {
-      "types": "./lib/task/index.d.ts",
-      "molstar-src": "./src/task/index.ts",
-      "import": "./lib/task/index.js"
+      "types": "./lib/task/task.d.ts",
+      "molstar-src": "./src/task/task.ts",
+      "import": "./lib/task/task.js"
     },
     "./*": {
       "types": "./lib/*.d.ts",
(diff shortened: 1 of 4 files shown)


docs/v6-architecture.md
```diff
@@ -0,0 +1,506 @@
+# Mol\* 6.0: packages, ESM, and plugin composition
+
+Proposal against the `molstar@5.11.0` tree. The APIs and paths below describe the target, not features available in 5.x. See the [short summary](v6-summary.md) for the main decisions.
+
+## 1. Scope
+
+Mol* 6.0 moves to a pnpm workspace of ESM packages grouped by layer. Parsers, representations, and themes become explicit plugin features; `DefaultPluginSpec()` remains the full built-in composition. Apps keep esbuild, with dependencies and build configuration owned by each app.
```
zachcp, line 7: +1  Moving to ESM rather than having a compile step makes experimentation in HTML via import maps really easy.  So I like the no-build distribution.

docs/v6-architecture.md
```diff
@@ -0,0 +1,506 @@
...
+
+Ship `@molstar/migrate-6` with the release to handle mechanical import changes and report manual work. The unscoped `molstar` package retains the CDN viewer bundle only: no legacy `lib/mol-*` exports, compatibility re-exports, or CommonJS build.
```
zachcp, line 9: Small operation note but if you publish on [JSR](http://jsr.io) and you have no slow types, they auto publish NPM-compatable CommonJS modules as well. 

Edit: to make this more explicit - publication is publication of Typescript source code only.

docs/v6-architecture.md
```diff
@@ -0,0 +1,506 @@
...
+
+The release also includes a standalone MolViewSpec builder, dependency-cycle removal, maintainer skills, updated developer docs, and workspace CI.
```
zachcp, line 11: +1

docs/v6-architecture.md
```diff
@@ -0,0 +1,506 @@
...
+| `@molstar/core` | `util/`, `task/`, `data/`, `math/`, `state/`; no molecular or rendering dependencies |
+| `@molstar/io` | Readers and writers; depends on core |
+| `@molstar/model` | Structures, volumes, particles, domain formats, properties, and script/query code |
+| `@molstar/graphics` | `gl/`, `geo/`, `theme/`, `repr/`, `canvas3d/`, plus rendering code relocated from model/math |
+| `@molstar/plugin` | Plugin and plugin-state together: runtime, spec helpers, and explicit default-spec/catalog entry points |
+| `@molstar/plugin-ui` | React UI, UI-spec types, and an explicit default UI-spec entry point |
+| `@molstar/mvs-builder` | Standalone MolViewSpec schema, builder, serialization, and validation |
+| `@molstar/mvs` | MolViewSpec runtime, feature, loader, and rendering CLI |
+| `@molstar/<name>-extension` | Individual extensions and their dependencies |
+| `@molstar/viewer` | Published Viewer API and app |
```
zachcp, line 60: This is great.

docs/v6-architecture.md
```diff
@@ -0,0 +1,506 @@
...
+| `mol-math/geometry/gaussian-density/gpu.ts` | core → GL | Keep CPU/domain-independent math in core; move GPU computation and its GL-facing API to graphics |
+| `mol-model/shape/shape.ts` | model → geometry, themes, GL | Separate shape data from rendering helpers; place graphics-dependent APIs in graphics |
+| `mol-model-formats/shape/*` | model → mesh builders | Move mesh conversion into graphics, retaining raw file readers in IO |
+| `mol-model-props/**/representations`, `**/themes`, and label helpers | model → representations/themes | Keep property computation in model; move visuals/themes to graphics and remove rendering dependencies from computation |
+| Model particle/unit transform utilities | model → geometry transform helpers | Extract shared matrix/transform primitives into core or move the graphics-specific callers |
+
+This is a starting list, not the full audit. Resolve declaration dependencies too: a type-only edge avoids runtime evaluation, but still requires a package dependency and can create a cycle in `tsc -b` references. Assign exact replacement subpaths during this work and record exceptions in the migration map.
```
zachcp, line 89: this is great. I imagine some additional cycles may show up here but teasing out these dependency chains are fantastic.

docs/v6-architecture.md
```diff
@@ -0,0 +1,506 @@
...
+- Moving Python `molviewspec` into this repository.
+- Publishing a stories library or merging the MolViewStories webapp.
+
+## 2. Starting point
+
+Today one package owns library code, apps, servers, and their npm dependencies. Consumers use deep imports such as `molstar/lib/mol-plugin-ui`; the package has no `exports` map.
+
+| Output | Current build |
+| --- | --- |
+| `lib/` | `tsc`, then `tsc-alias` to complete relative JS paths; includes declarations |
```
zachcp, line 33: As part of this effort I would suggest updating the imports so this step is not needed. E.g. enforce the updated pattern in styler - e.g. eslint/deno fmt etc.

docs/v6-architecture.md
```diff
@@ -0,0 +1,506 @@
...
+
+Publish ESM JavaScript and declarations in `lib/`, plus source in `src/` for bundlers and debugging. Set `"type": "module"` on published packages and retain the existing Node **22 or later** baseline; raise it only if runtime or tooling requirements demand it.
+
+| Consumer | Resolution | Executes |
+| --- | --- | --- |
+| Installed library, CLI, or server | Default `import` condition | `lib/*.js` |
+| TypeScript consumer | `types` condition | Checks `lib/*.d.ts` |
+| In-repo esbuild | `molstar-src` condition | Compiles source, including TSX |
+
+Node executes compiled JavaScript. Keep all published bins on `lib/*.js`; workspace tooling runs as JavaScript or is compiled before execution. `molstar-src` selects source for bundlers and does not promise native Node execution. Publishing source does not require erasable syntax.
```
zachcp, line 202: Is it possible to use a tool that works natively with Typescript? Bun/Deno/webbrowsers /Node > 23 .  Staying ESM + TS would keep everything nice and lean.

docs/v6-architecture.md
```diff
@@ -0,0 +1,506 @@
...
+    "verbatimModuleSyntax": true,
+    "isolatedModules": true,
+    "jsx": "react-jsx",
+    "declaration": true,
+    "declarationMap": true
+  }
+}
+```
+
+Use `.js` relative specifiers in TypeScript source as described in §4.2. No extension-rewriting compiler option is needed. Keep `verbatimModuleSyntax` and explicit `import type` for predictable module dependencies.
```
zachcp, line 225: I'm forgetting the right way to do this. Is it `.js` or `.ts` . Ah this is what I am remembering. There is a difference in how Deno and Node hande TS imports. I'll leave [this here for reference](https://docs.deno.com/runtime/fundamentals/typescript/) even though the above is consistent for Node.

Image

docs/v6-architecture.md
```diff
@@ -0,0 +1,506 @@
...
+## 7. MolViewSpec, extensions, and apps
+
+Split `extensions/mvs` into:
+
+| Package | API and responsibilities |
+| --- | --- |
+| `@molstar/mvs-builder` | Schema, `createMVSBuilder`, `MVSData`, MVSJ/MVSX serialization/validation, `mvs-validate`, schema-printing CLI |
+| `@molstar/mvs` | `loadMVS`/`loadMVSData`, plugin feature, annotations, cameras, runtime representations, `mvs-render` |
+
+The builder replaces [molviewspec-ts](https://github.com/molstar/mol-view-spec/tree/master/molviewspec-ts) and the JSR `@molstar/molviewspec` distribution after API/parity checks. Publish the replacement to npm and JSR. Own `io-ts` and any archive dependencies in the builder; inline or extract its small general helpers without introducing a Mol* runtime dependency. An optional core peer would not make unconditional core imports optional.
```
zachcp, line 403: +1

zachcp:
@dsehnal  I love it!  I think moving to ESM classes is great and that identifying and fixing import cycles will hone the capabiliteis of each of the subpackages and make long-term contributions cleaner.   I made a number of small comments, many of which revolve around some of the compilation steps. As you know, I've been a fan of the [Deno/JSR](https://github.com/molstar/molstar/discussions/1599) ecosystem in part because it is designed to be ESM+TS first so many of the compilation steps "go-away" as long as you meet the compilation requirements around "slow-types".   If you're committed to PNPM/Node then you can discount the specifics of those comments. All in all I think its a great direction. I also like the idea of moing towards WebGPU but I am less familiar with the internals there. Go, David, go!

docs/v6-architecture.md
dsehnal: Ok, I will see how much work would this actually be. We would also have to import the deno listing for this, which I am not sure I want to do currently.

Or we can go the bun way and just rewrite everything in Rust /s :D

docs/v6-architecture.md
dsehnal: Apparently Node can't execute packaged TS. 

The `lib/` stuff is something I need to iterate on more.

docs/v6-architecture.md
zachcp: > We would also have to import the deno listing for this, which I am not sure I want to do currently.

Yep. No Prob. Theres a lot of simultaneous changes.

dsehnal:
@zachcp had a look into Deno support in https://github.com/molstar/molstar/pull/1931/commits/4f5996faa59e1a022a57685d7576ac8f2818eed0 and added [v6-fasttypes](https://github.com/molstar/molstar/blob/v6-architecture/docs/v6-fasttypes.md).

It's not impossible, but would be quite a lot of work currently, so deferring that part for v7. Will at least try a "slow types" release for v6.

docs/v6-fasttypes.md
```diff
@@ -0,0 +1,266 @@
...
+
+- **npm:** compiled ESM JavaScript and declarations in `lib/`, plus `src/` for opt-in bundlers and debugging.
+- **JSR:** TypeScript source and source entry points, allowing slow types. Start with the standalone MVS builder and extend coverage package by package after validation.
+- **Release:** one version and commit, separate publication steps and validation for each registry. Registry availability does not imply support for every runtime.
+
+`lib/` is generated and disposable in the checkout, but its runtime files remain in the npm tarball. JSR source publication does not need that directory. Fast types do not require erasable syntax, native Node TypeScript execution, or replacing the existing build system.
+
+Allowing slow types can make JSR consumer checking slower and leave JSR-generated documentation or npm-compatibility declarations incomplete or imprecise. Accept those limitations for this v6 experiment; native npm consumers keep the full `tsc` declaration output. The flag does not waive normal typechecking, import resolution, or other publication rules. See [JSR's slow-types limitations](https://jsr.io/docs/about-slow-types).
+
+## 2. What fast types require
```
zachcp, line 21: Thanks for looking into this. I agree that this is probably a follow on to V6. Many of the slow types are really about adding explicity types to ARGs/fn returns  which a coding agent + `deno lint` can fix relatively easily. If i recall there are a few factory patterns that make type resolution a bit tricky.  

However, if those can be resolved, you can get TS-source distribution on JSR, and offload the entire compilation process to the JSR backend - e.g. the repo becomes solely about logic.  Just as you don't need to think about distributing `.pyc` files for python, you can publish and use TS source as the first class artifact.

docs/v6-fasttypes.md
```diff
@@ -0,0 +1,266 @@
...
+| Parameters and implicit `undefined` | 9011, 9025 | 110 |
+| Enum initializers, class factories/heritage, computed keys, private names | 9020–9022, 9038–9039 | 100 |
+
+Most edits should be annotations, but the smaller final category and inference-heavy schemas carry disproportionate design risk.
+
+### 3.4 Deno verification
+
+Deno 2.9.6 ran `deno lint` with only `no-slow-types` enabled against a temporary copy of all 1,590 implementation modules from the TypeScript audit. The temporary manifest exported every module explicitly and its lint include list matched those files. No repository configuration was changed, no diagnostics were suppressed, and the lint result contained no processing errors.
+
+| Deno diagnostic | Count |
```
zachcp, line 99: I looked into fixing these last year at some point ( and have a branch somewhere ). The type signatures are easy to fix.  The superclass expression changes needed a bit more thinking through but in almost all cases - seesawing between and agent editing followed by `deno lint && deno test` led to resolution.  And the agents have gotten a lot better in the past year.

docs/v6-fasttypes.md
```diff
@@ -0,0 +1,266 @@
...
+2. **Shared patterns:** agree on parameter-schema composition, typed class-factory bases, and generator output before parallel expansion. Give one owner each shared contract and generator; other agents consume the accepted result.
+3. **Bounded batches:** start around 10–25 related files, or one generator/factory family. Each task states its owned paths, allowed API changes, baseline, and required checks. Integrate completed batches frequently before assigning dependent work.
+4. **Independent validation:** a worker's “checks pass” report is evidence to inspect, not approval by itself. Re-run the relevant checks on the combined branch and review every public declaration change. A separate reviewing agent can assist; the maintainer owns deliberate API decisions.
+
+Do not let workers satisfy diagnostics by adding `any`, broadening schemas, suppressing errors, hiding intended public exports, or introducing circular `typeof` annotations. Existing `any` need not be redesigned as part of this migration, but new loss of precision requires an explicit API decision. Keep runtime refactors and package relocation out of annotation batches.
```
zachcp, line 207: many of these rule could be added to eslint/tsc without requiring Deno at all.  JSR is just stricter about types since it allows them to type check semantically without needing to interpret the code - and therefore serve the bits you need better/faster.

papillot:
Thanks David for opening the discussion!

Shall we consider barrel files as well? At least some bundlers docs assert that it prevents tree-shaking https://vite.dev/guide/performance#avoid-barrel-files

dsehnal:
> Thanks David for opening the discussion!
> 
> 
> 
> Shall we consider barrel files as well? At least some bundlers docs assert that it prevents tree-shaking https://vite.dev/guide/performance#avoid-barrel-files

I would view the barrel files as optional. Although it's quite easy to misuse. Will include this in the design.

PR #1932, opened 2026-09-11
chore: general performance improvements

TaylorHo:
# Description

Optimized the viewer rendering hot path to reduce redundant GPU/CPU work during rendering.

### Changes

* Skip redundant uniform uploads when `ValueCell` values have not changed.
* Batch shared uniform and texture updates per program switch instead of per render object.
* Cache active texture units and bound textures to avoid redundant GL calls.
* Cache culling results when grid/LOD state has not changed.
* Reuse the light-direction buffer and fix its upload size.

### Results

* **71% reduction** in `Canvas3D.render` CPU time.
* **72% reduction** in `DrawPass.render` CPU time.
* **54% reduction** in multisample rendering CPU time.
* **28% reduction** in steady-state CPU mean and **52% reduction** in CPU median.
* **34% reduction** in maximum GPU frame time.
* **48% reduction** in maximum CPU frame time.
* **94.1% of uniform uploads skipped** in the final frame.

These changes significantly reduce CPU rendering overhead and improve worst-case frame hitches, while also providing moderate GPU improvements.

## Actions

### DIFF
--- a/src/mol-gl/webgl/state.ts
+++ b/src/mol-gl/webgl/state.ts
@@ -2,6 +2,7 @@
  * Copyright (c) 2018-2024 mol* contributors, licensed under MIT, See LICENSE file for more info.
  *
  * @author Alexander Rose <alexander.rose@weirdbyte.de>
+ * @author Taylor Hoffmann <taylor@hoffmann.io>
  */
 
 import { GLRenderingContext } from './compat';
@@ -87,6 +88,8 @@ export type WebGLState = {
     clearVertexAttribsState: () => void
     disableUnusedVertexAttribs: () => void
 
+    bindTexture: (unit: number, target: number, texture: WebGLTexture | null) => void
+
     viewport: (x: number, y: number, width: number, height: number) => void
     scissor: (x: number, y: number, width: number, height: number) => void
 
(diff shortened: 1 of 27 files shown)


arose:
Thank you @TaylorHo. This looks interesting. Please split this into multiple PRs that target a specific kind or area of optimization.

I would be especially curious to see what effect tracking texture units in the state has and the extra sorting of scene renderables. Please leave the renderable culling out, I am about to do a PR that handles it a bit more general so it also works for multi-sample (will ping you as a reviewer).

I don't think the uniform optimization works. You will need to re-upload them when ever the program changes and are careful not to do so unnecessarily, see RenderItem.render. What helps with reducing uniform calls is UBOs (webgl2 only). I have a prototype for it and it helps a little here.

src/mol-gl/webgl/state.ts
```diff
@@ -403,7 +420,11 @@ export function createState(gl: GLRenderingContext, e: WebGLExtensions): WebGLSt
         } : undefined,
 
         reset: () => {
-            enabledCapabilities = {};
+            for (const k of Object.keys(enabledCapabilities)) {
```
dsehnal, line 423: Is this really faster than allocating a single new object?

src/mol-state/state.ts
```diff
@@ -805,9 +806,10 @@ function relinkCells(target: StateObjectCell, ctx: UpdateContext): boolean {
 
     // Fast path: same number and all current refs are still in effective.
     if (current.length === effective.length) {
+        const effectiveSet = new Set(effective);
```
dsehnal, line 809: Have a reusable set that you clear?

dsehnal (commented):
Thanks for going through this.

TaylorHo:
> Thank you @TaylorHo. This looks interesting. Please split this into multiple PRs that target a specific kind or area of optimization.

Sure! I'll review your comments and split this into smaller PRs this week. Thanks in advance for the invite to the repo!

PR #1933, opened 2026-09-12
Optimize `Renderable` culling

arose:
# Description

- add `CullCache`
- refactor multi-draw-base data handling

=> helps with any type of multi pass rendering we do: multi-sample, marking, emissive, ...

## Actions


arose:
FYI @TaylorHo

src/mol-gl/renderable.ts
```diff
@@ -68,283 +70,358 @@ function getMdbData(cellCount: number, mdbData?: MultiDrawBaseData): MultiDrawBa
...
+            if (!nearlyEqual(frustumArray[o], pn[0]) || !nearlyEqual(frustumArray[o + 1], pn[1]) || !nearlyEqual(frustumArray[o + 2], pn[2]) || !nearlyEqual(frustumArray[o + 3], p.constant)) return false;
+            o += 4;
+        }
+        return true;
+    }
+
+    return {
+        unchanged(values, cameraPlane, frustum, isOccluded) {
+            if (!has) return false;
+            if (isOccluded !== isOccludedRef) return false;
```
corredD: compares the function, not its answer. hiZ.isOccluded is a class-field arrow, so the reference never changes while its result changes every frame. The cull never re-runs, and occlusion-culled geometry stays hidden after the occluder is gone.

src/mol-gl/renderable.ts
```diff
@@ -68,283 +70,358 @@ function getMdbData(cellCount: number, mdbData?: MultiDrawBaseData): MultiDrawBa
...
-                    const o = l.count;
-
-                    l.counts[o] = lodLevels[j][3];
-                    l.instanceCounts[o] = values.instanceCount.ref.value;
-                    l.baseInstances[o] = 0;
-                    l.count += 1;
-                    break;
-                }
-            }
+            mdb.cullSimple(values, lodLevels, d, radius, scale);
```
corredD, line 338: mdb.prepare(...) runs before if (!lodLevels) return; and zeroes every list[j].count. The old code returned first. cullEnabled is still true from the last cull(), so nothing is drawn.

src/mol-gl/renderable.ts
```diff
@@ -68,283 +70,358 @@ function getMdbData(cellCount: number, mdbData?: MultiDrawBaseData): MultiDrawBa
...
-    const mdbDataList: MultiDrawBaseData[] = [];
-    let cullEnabled = false;
-    let lodLevelsVersion = -1;
+/** The minimal set of values needed to cull a render item. */
+type CullValues = {
+    readonly drawCount: ValueCell<number>
+    readonly instanceCount: ValueCell<number>
+    readonly uLod: ValueCell<Vec4>
+    readonly instanceGrid: ValueCell<InstanceGrid>
+    readonly boundingSphere: ValueCell<Sphere3D>
```
corredD, line 97: boundingSphere in CullValues is never read — not versioned in ownValuesUnchanged, not used by cull/cullSimple.

src/mol-gl/renderable.ts
```diff
@@ -68,283 +70,358 @@ function getMdbData(cellCount: number, mdbData?: MultiDrawBaseData): MultiDrawBa
...
+            lodCount[j] = lodLevels![j][3];
+        }
 
-            const lodLevels: [minDistance: number, maxDistance: number, overlap: number, count: number, sizeFactor: number][] | undefined = values.lodLevels?.ref.value;
+        // caller only invokes this when the instance grid is usable (cellSize > 1)
+        const { cellOffsets, cellSpheres, batchOffsets, batchSpheres, batchCount, batchSize } = grid;
+        const checkCellOccludedDistance = 2 * batchSize;
+
+        for (let k = 0; k < batchCount; ++k) {
+            // batchCell is serial after the grid reorder, cells of a batch are consecutive
```
corredD: The new comment is right that batchCell is serial — and this was its last consumer instance-grid.ts:128 still allocates and fillSerials a Uint32Array(cellCount) per grid build that nothing reads.

corredD (commented):
few comments. lgtm

src/mol-gl/renderable.ts
arose: right, will set `cullEnabled = false;` at the start

src/mol-gl/renderable.ts
arose: will remove the CullCache and use a "frame" based concept instead

src/mol-gl/renderable.ts
arose: will remove the CullCache and use a "frame" based concept instead

corredD (commented):
nice optimization, lgtm

PR #1934, opened 2026-09-12
Support merging render-objects

arose:
# Description

- add support for merging render-objects: handles spheres, mesh, cylinders, lines, points
- merge representation render-objects: structure units/particle target representation
- prefer merging render-object over structure level visuals

Transparently merges render-objects with minimal code changes to the representation code.

For example, greatly speeds-up rendering of structures with many units without falling back to structure-level visuals (that don't support unit-based clipping). Try PDB entry 7y7a.

## Actions


dsehnal (approved):
Skimming the code everything looks reasonable.
Tried the recommended structure and few others, feels very snappy, GJ.

corredD (commented):
Great improvement ! I found one issue : "aSegmentLod is a vec4, so a 5th lod level
  silently renders at level 4's scale : worth rejecting lodLevels.length > 4 in
  canMergeValues.

arose:
> Great improvement ! I found one issue : "aSegmentLod is a vec4, so a 5th lod level silently renders at level 4's scale : worth rejecting lodLevels.length > 4 in canMergeValues.

Would not reject it, because I think the performance it hit is worse than the degraded quality. We could pack to 16 bit floats if need be. Have you ever used more than 4 levels?

PR #1935, opened 2026-09-12
Improve dynamic trackball controls and show param

arose:
# Description

Improve dynamic trackball controls and show param. Some tweaks and show param in UI.

## Actions

### DIFF
--- a/src/mol-canvas3d/controls/trackball.ts
+++ b/src/mol-canvas3d/controls/trackball.ts
@@ -85,8 +85,8 @@ export const TrackballControlsParams = {
         }, { description: 'Rock the 3D scene around an axis in camera space' })
     }),
 
-    staticMoving: PD.Boolean(true, { isHidden: true }),
-    dynamicDampingFactor: PD.Numeric(0.2, {}, { isHidden: true }),
+    staticMoving: PD.Boolean(true),
+    dynamicDampingFactor: PD.Numeric(0.2, { min: 0, max: 1, step: 0.01 }),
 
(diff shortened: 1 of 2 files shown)


src/mol-canvas3d/controls/trackball.ts
```diff
@@ -221,7 +221,31 @@ namespace TrackballControls {
...
         const rotMoveDir = Vec3();
 
-        function rotateCamera() {
+        // `dynamicDampingFactor` decays/lags inertia per-call; without normalizing by elapsed
+        // time, slower frame rates (e.g. very large scenes) decay far slower in wall-clock
+        // time, so residual camera motion
+        const REFERENCE_DT = 1000 / 60;
+        /** raises a factor tuned for one `REFERENCE_DT` step to the equivalent for `deltaT` */
+        function timeNormalized(perFrameFactor: number, deltaT: number) {
+            return Math.pow(perFrameFactor, deltaT / REFERENCE_DT);
```
dsehnal, line 230: I am not sure normalizing to 60 fps is the right call. `requestAnimationFrame` can tick slower of faster depending on platform.

> On a 120 Hz display (like an iPhone's ProMotion screen or modern gaming monitor), rAF will trigger roughly 120 times per second (~8.33 ms intervals) if the main thread is unblocked.

Maybe at least make it a parameter? Or measure this and don't let this fall below 30 fps?

arose: that is intentional, when slower than 60fps it dampens more per step and on faster it dampens less, so the user experiences roughly the same.

dsehnal (approved):

Issue #1084, opened 2024-04-16
Bumpiness not working on ball and stick representation


giagitom:
It does not work using both meshes or impostors

arose:
That is intended, the default `bumpFrequency` is set to zero for some representations where is does not look good. See src\mol-repr\structure\representation\ball-and-stick.ts

giagitom:
Hmm.. It seems that also varying bump frequency does not produce any effect on the representation. Maybe it is disabled somewhere else?

![image](https://github.com/molstar/molstar/assets/20627977/7387f3b3-d68b-4e89-8b3e-fedc682964b5)

arose:
I find the artifact quite strong without specular highlights

![image](https://github.com/molstar/molstar/assets/272250/213f70f3-b0d0-4888-9e72-acd38124ad5b)

arose:
If you can fix the artifact, I am happy to reconsider

giagitom:
@arose I have found an other related issue with bumpiness on impostor spheres and cylinders that is most visible when xRayShaded is on. Here is a test state to reproduce the issue [test.molj.txt](https://github.com/molstar/molstar/files/15178904/test.molj.txt)

![image](https://github.com/molstar/molstar/assets/20627977/29eb3d5e-80d1-42cf-989d-ee47b43920da)

The issue seems to be due to this piece of code inside ```perturbNormal```
https://github.com/molstar/molstar/blob/9604b89ee0716442b0bc8b862cfe712468dc6c95/src/mol-gl/shader/chunks/common-frag-params.glsl.ts#L114

Handling edge cases in which det == 0 like this ```if (det == 0.0) return vec3(normal);``` seems to fix the problem and also reduces the effect of the before mentioned artifact (that is still present though).

I suspect there is an other similar edge case breaking the perturbed normal.

arose:
thanks, applied the fix in fda3481d1504213dbabf70ec264580df2dfb22ff

PR #1936, opened 2026-09-13
Fix bumpines on impostors

giagitom:
# Description

This is a fix for #1084 

## Actions


arose:
Nice, this does more calculations than using derivatives but probably not a bottleneck. Got some explanation for the math?

giagitom:
Here is a brief explanation:

The slope is measured on the surface itself, from this fragment only.

1. Build two directions in the surface from the normal `n`: `t1 = normalize(cross(n, axis))`, `t2 = cross(n, t1)`.
2. Sample the noise at the point and one small step `e` along each direction, and take rise over run:
   `s1 = (h(p + e t1) - h(p)) / e`, `s2 = (h(p + e t2) - h(p)) / e`.
3. Combine into the surface gradient: `g = s1 t1 + s2 t2`.
4. Tilt the normal: `n' = normalize(n - scale * g)`, with `scale = amplitude * bumpiness / frequency` as before.

Notes:
- The noise lives in model space, so `t1` and `t2` are rotated with `uInvView` before stepping.
- `e` is one pixel footprint at the fragment depth, floored at 1% of the noise wavelength, so the bump scale and strength match the old result.
- No screen derivatives are used, so seams, cuts and silhouettes cannot leak in and the 2x2 quad blockiness is gone. Cost: two extra `fbm` calls per bumpy fragment.

arose (changes requested):
very nice, and code looks good, thanks

can you also change the bumpFrequency defaults of the ball&stick (5) and backbone (4) ?

giagitom:
Done

arose (approved):

PR #1937, opened 2026-09-14
Support solid interior caps for meshes at the near plane and clip planes, and for impostors at any clip object

giagitom:
# Description

## Add `solidInterior` option for mesh visuals

image

Adds a `solidInterior` param to mesh and texture-mesh geometries, mirroring the option that impostor spheres and cylinders already have. When the camera near plane cuts into a closed surface, the cut is closed with a solid cap instead of showing the hollow interior.

Off by default, WebGL2 only (falls back to the current behaviour on WebGL1).

### How it works
- A stencil parity pass counts front and back faces per pixel; the stencil is non-zero where the near plane is inside the solid.
- The cap is then drawn from the back faces, snapped to the near plane, where the stencil is set.
- Works with opaque, blended, WBOIT and DPOIT transparency and with the illumination (tracing) pass. The cap is also written to the transparent depth so SSAO, outlines and DoF see it.

### Changes
- `depth-stencil` texture format and a `depthStencil` option for render targets.
- `solidInterior` param and cap passes for mesh / texture-mesh.
- DPOIT cap composited once after depth peeling.

## Actions

- [X] Added description of changes to the `[Unreleased]` section of `CHANGELOG.md`
- [X] Updated headers of modified files


corredD:
this is really nice ! could this be done with all the clipping object ? Or at least with the plane object ?

giagitom:
Thanks @corredD, 

> could this be done with all the clipping object ? Or at least with the plane object?

Now it should work with plane clip objects too

image

src/mol-gl/renderer.ts
```diff
@@ -404,6 +413,98 @@ namespace Renderer {
...
+            globalUniformsNeedUpdate = true;
+            if (hwDepthTest) {
+                state.enable(gl.DEPTH_TEST);
+                state.depthFunc(gl.LEQUAL);
+            }
+            state.colorMask(true, true, true, true);
+            state.depthMask(writeDepth);
+            state.stencilFunc(gl.NOTEQUAL, 0, 0xff);
+            state.stencilOp(gl.KEEP, gl.KEEP, gl.ZERO);
+            renderObject(r, variant, Flag.BlendedBack);
```
corredD, line 448: The cap pass draws the cap from back faces, so mesh.frag:55 marks every cap fragment interior, and the shared check-transparency chunk then discards it whenever the mesh is transparent, which is why the slice vanishes in every transparency mode. In mesh.frag (line 39 already has bool capPass), you could exempt cap fragments from the interior-backface discard while leaving interior set so the cap keeps its interior colour.

src/mol-canvas3d/passes/draw.ts
```diff
@@ -267,6 +267,15 @@ export class DrawPass {
             // evaluate dpoit
             target.bind();
             this.dpoit.render();
+
+            const capStencil = isPostprocessingEnabled ? this.colorTarget.depthRenderbuffer : null;
```
corredD, line 271: why do you gate on isPostprocessingEnabled ?

src/mol-gl/shader/mesh.frag.ts
```diff
@@ -20,9 +21,36 @@ precision highp int;
...
+        }
+        bool capPass = uSolidInteriorPass != 0;
+        vec3 viewPosition = vViewPosition;
+        vec3 modelPosition = vModelPosition;
+        if (capPass) {
+            viewPosition = vCapPosition.xyz / vCapPosition.w;
+            modelPosition = (uInvView * vec4(viewPosition, 1.0)).xyz;
+        }
+        #if defined(dClipVariant_pixel) && dClipObjectCount != 0
+            if (clipTest(modelPosition)) discard;
```
corredD, line 47: clip_pixel was dividing by uModelScale and this replacement doesn't, is that deliberate ?

src/mol-gl/renderer.ts
```diff
@@ -744,6 +851,9 @@ namespace Renderer {
                 const r = renderables[i];
                 if (checkOpaque(r)) {
                     renderObject(r, 'color', Flag.None);
+                    if (hasSolidInteriorCap(r)) {
+                        renderSolidInteriorCap(r, 'color', 'opaque');
```
corredD, line 855: The cap is only rendered in the depth, tracing and color passes not in renderPick or marking. What should be the mouse over/picking behaviour for a mesh cap ?

src/mol-gl/webgl/texture.ts
```diff
@@ -132,6 +136,7 @@ function getFormatSize(format: TextureFormat) {
         case 'rgb': return 3;
         case 'rgba': return 4;
         case 'depth': return 4;
+        case 'depth-stencil': return 4;
```
corredD, line 139: getFormatSize is a component count (rgba→4), so depth should be 1 and depth-stencil should be 2 as written it reports 16 B/px for an 8 B/px format, and since 'depth' also reports 16 for 4, the real 4→8 increase doesn't show in getByteCount(). The 'depth' half is pre-existing, so drop it if you'd rather keep scope tight.

src/mol-gl/shader/chunks/common-clip.glsl.ts
```diff
@@ -108,5 +111,131 @@ float getSignedDistance(const in vec3 center, const in int type, const in vec3 p
...
+
+    float clipExit(const in vec3 pFront, const in vec3 pBack, const in float sStart) {
+        #if defined(dClipping)
+            int flag = int(floor(vClipping * 255.0 + 0.5));
+        #else
+            int flag = 0;
+        #endif
+
+        float s = sStart;
+        for (int k = 0; k < 8; ++k) {
```
corredD, line 219: why 8 ?

corredD (commented):
This is excellent and it looks great @giagitom ! Would it be hard / usefull to support all clipping object ? What do you think @arose ? 
Also, should the cap use the color of the representation clipped ?

src/mol-gl/renderer.ts
giagitom: you need `transparentBackfaces: 'on'` to see the cap on transparent objects since it is considered as interior.

src/mol-canvas3d/passes/draw.ts
giagitom: the condition is now written as `target !== this.colorTarget`. The effect should be the same.

src/mol-gl/shader/mesh.frag.ts
giagitom: Thanks, fixed in commit.

src/mol-gl/renderer.ts
giagitom: Cap pixels now write the null picking group, so any getLoci resolves them to the whole instance

src/mol-gl/webgl/texture.ts
giagitom: Fixed both.Thanks

src/mol-gl/shader/chunks/common-clip.glsl.ts
giagitom: It was a safety margin I forgot to replace. Fixed, thanks

giagitom:
> Would it be hard / usefull to support all clipping object 

It should be possible, I'm working on this. 


> Also, should the cap use the color of the representation clipped ?

It would be nice to have a different granularity for the interior. I would like to explore that separately.

giagitom:
Full support on meshes for every clip object in last commit

image

arose:
looks awesome, will review later

src/mol-gl/renderer.ts
```diff
@@ -404,6 +411,112 @@ namespace Renderer {
...
+            const type = values.uClipObjectType.ref.value;
+            const invert = values.uClipObjectInvert.ref.value;
+            const position = values.uClipObjectPosition.ref.value;
+            const rotation = values.uClipObjectRotation.ref.value;
+            const scale = values.uClipObjectScale.ref.value;
+            const transform = values.uClipObjectTransform.ref.value;
+            for (let i = 0; i < count; ++i) {
+                Mat4.fromArray(solidInteriorTransform, transform, i * 16);
+                Vec3.fromArray(solidInteriorPosition, position, i * 3);
+                if (type[i] === Clip.Type.plane) {
```
arose: why is the plane check so complex here? generally can we add helper to Clip and/or math?

src/mol-gl/renderer.ts
```diff
@@ -528,8 +641,12 @@ namespace Renderer {
 
             const { renderables } = group;
             for (let i = 0, il = renderables.length; i < il; ++i) {
-                if (!renderables[i].state.colorOnly) {
-                    renderObject(renderables[i], variant, Flag.None);
+                const r = renderables[i];
+                if (!r.state.colorOnly) {
+                    renderObject(r, variant, Flag.None);
+                    if (hasSolidInteriorCap(r)) {
+                        renderSolidInteriorCap(r, variant, 'opaque');
```
arose: rendering the cap here produces many webgl state changes and uniform churn. can we split renderSolidInteriorPass  into two passes and do full loops over the renderables to minimize state changes?

arose (changes requested):
so cool, two renderer comments, will look more into it

src/mol-gl/renderer.ts
giagitom: Yeah, that was too much for the renderer. A clip plane is stored as position + rotation plus a `transform` applied to the point before testing, so to cull on the CPU I had to bring the plane back to world space by hand. I've moved that into `Clip.getPlane()` (returns a `Plane3D`) and `Clip.canIntersectSphere()`, with a small spec. The loop in the renderer is just two checks now.

src/mol-gl/renderer.ts
giagitom: Good point, the churn was real. Caps now render in their own loop after the normal one, so the main loop isn't interrupted anymore. Stencil state is set once per loop, the pass is split into a mark and a fill function, and the pass ints are set directly on the program instead of re-uploading all the globals. On GroEL (35 units), CPU time in `renderOpaque` dropped from about 1.3 to 0.5 ms with one clip plane, and the images are pixel-identical to before.

I couldn't go all the way to "mark everything, then fill everything" with one shared stencil, though. The stencil doesn't know which object marked a pixel, so with transparency an object sitting behind would paint the front object's cap with its own colour. Clip objects are also per renderable, so the cut indices don't line up between them. So it's still one stencil per renderable, just without the interleaving. Blended stays interleaved because it depends on draw order. Let me know if you see a way around that.

src/mol-gl/webgl/context.ts
```diff
@@ -288,7 +289,7 @@ export interface WebGLContext {
     /** Cache for textures, managed by consumers */
     readonly namedTextures: { [name: string]: Texture }
 
-    createRenderTarget: (width: number, height: number, depth?: boolean, type?: 'uint8' | 'float32' | 'fp16', filter?: TextureFilter, format?: 'rgba' | 'alpha') => RenderTarget
+    createRenderTarget: (width: number, height: number, depth?: boolean, type?: 'uint8' | 'float32' | 'fp16', filter?: TextureFilter, format?: 'rgba' | 'alpha', depthStencil?: boolean) => RenderTarget
```
arose: let's not have an extra param for depthStencil but rename depth to depthStencil with the three option (stencil alone is not a real thing)

src/mol-gl/renderer.ts
```diff
@@ -405,6 +412,108 @@ namespace Renderer {
...
+        const endSolidInteriorCaps = (mode: 'opaque' | 'blended' | 'oit') => {
+            state.disable(gl.STENCIL_TEST);
+            state.stencilOp(gl.KEEP, gl.KEEP, gl.KEEP);
+            if (mode !== 'oit') state.depthFunc(gl.LESS);
+            state.frontFace(gl.CCW);
+            state.cullFace(gl.BACK);
+        };
+
+        const renderSolidInteriorCaps = (renderables: ReadonlyArray<GraphicsRenderable>, check: (r: GraphicsRenderable) => boolean, variant: GraphicsRenderVariant, mode: 'opaque' | 'oit') => {
+            if (!isWebGL2) return;
```
arose: why is webgl2 required? would webgl1 plus depthTexture ext work too?

src/mol-gl/webgl/context.ts
giagitom: Done, it's now `depthStencil: 'none' | 'depth' | 'depth-stencil'` instead of the two flags. On WebGL1 the `'depth-stencil'` targets go from depth16 to depth24+stencil8 (core `DEPTH_STENCIL` renderbuffer).

src/mol-gl/renderer.ts
giagitom: You're right, it isn't needed, I had just picked the 32f+8 formats. Stencil and `DEPTH_STENCIL` renderbuffers are core in WebGL1, `WEBGL_depth_texture` gives the depth-stencil texture and `EXT_frag_depth` is what the cap depth needs, so that's the only gate now. Tested on WebGL1 with and without the depth texture extension, caps and cap picks match WebGL2. On the way I found that the packed depth pass dropped all back faces of opaque objects, so outlines and occlusion were off wherever the interior is visible (camera inside a surface, clipped meshes, impostor near caps); the discard is now limited to the transparent mask, where it matches the colour pass.

arose:
A call to `renderSolidInteriorCaps` is missing from renderDepth (although it seems we don't use that anymore, so either remove or fix), `renderDepthOpaqueBack`, `renderEmissiveOpaque`, `renderEmissiveTransparent`. With emissive, the issue is easy to spot by just making a single residue emissive. I haven't look to find an example for depth opaque back in illumination mode. In `renderBlendedTransparent`, `renderSolidInteriorCap` is still called interleaved.

Please add `timer.mark` to `renderSolidInteriorCaps`.

Would be nice to have a `solidInterior` option in `StructureComponentManager`

giagitom:
Thanks! Added the caps to `renderDepth`, `renderDepthOpaqueBack` and both emissive passes, plus the `timer.mark` and a `solidInterior` option in `StructureComponentManager`. `renderBlendedTransparent` stays interleaved on purpose, since blended depends on draw order and writes no depth.

Also fixed: the canvas had no stencil, so caps broke when drawing straight to it, and the mesh param had switched `solidInterior` off for backbone, dots and particles. Cartoon is now off on purpose, since it's mostly meshes.

PR #1938, opened 2026-09-16
Interior coloring improvements

giagitom:
# Description

## Optional interior color from the color theme

Interior faces and solid interior caps currently show the flat `interior.color` tint over whatever back face happens to rasterize, which on a cut surface means the far wall's per-atom colors. A color theme can now provide its own color for interiors.

image

### API

`ColorTheme` gains two optional fields:

- `interiorColor?: LocationColor`
- `interiorGranularity?: ColorTypeLocation` 

Three themes provide one, each handing over the sub-theme it already carries: `element-symbol` its carbon theme, `illustrative` its style theme, `cartoon` its mainchain theme. A cut surface then keeps its per-atom colors outside and shows one flat color per chain or entity inside.

### Params

`interior.themeColor` (default off) enables it. The flat `interior.color` still
applies on top at `interior.colorStrength`, which was lowered from 1 to 0.5 so
the underlying colors stay visible.

### Implementation

The interior colors reuse `createColors` through a view on the interior value
cells, so every granularity and the palette behave exactly like the main color.
They are rebuilt on the same trigger as the main colors, and nothing is
allocated when a theme provides no interior color. The GLSL is compiled out for
geometries without interiors and for the emissive variant.

## Actions


giagitom:
I added also the solid interior commits to give an idea of how it could be at the end. the last is the new commit [80594ec](https://github.com/molstar/molstar/pull/1938/commits/80594ecdee02c1c1270ebe2024e52738e92b1e66)

arose:
will have a look, can you change the target from master to [giagitom:solidInterior-meshes](https://github.com/giagitom/molstar/tree/solidInterior-meshes) so it is easier to see changes

giagitom:
@arose GitHub doesn't allow changing the base repository of an open PR, only the base branch within `molstar/molstar` — so I've re-opened this against `solidInterior-meshes` in my fork instead: giagitom/molstar#1 (https://github.com/giagitom/molstar/pull/1). The diff there is just the single new commit.

I also merged `solidInterior-meshes` into the branch first, otherwise `fc4fcb9` (nearest back face for the caps) would have shown up as a revert.

Closing this one; I'll open a fresh PR here once #1937 is merged.

PR #1939, opened 2026-09-17
Bonds provider registry POC

papillot:
# Description
This branch is a POC of the implementation of a bond provider as discussed in this pull request https://github.com/molstar/molstar/pull/1855

It makes minimal changes to the current Mol* core code: adds a call to BondProviderRegistry (optionally populated) in Unit.Atomic.bonds getter, and exports functions that are required for computing the bond graph.

The rest of the logic is in the extension. For readability, none of the bond perception actual code is in this PR. The POC only transforms every C-C bond in ligands to double bond.

The extension contains a preset, which applies a new transform decorator (PerceiveBondOrders) to a Structure. This decorator registers the Bond provider on the model using the structure's units model property. When the bond provider is present on the model, it owns the computations and the caching for the unit.bonds getter.

```mermaid
flowchart TD
  D["PerceiveBondOrders decorator"] --> R["registry.add(provider)"]
  R --> A["registry.select(provider.name)"]
  A --> L["later: unit.bonds"]
  L --> P{"provider returns graph?"}
  P -- yes --> O["return provider graph directly"]
  P -- no --> C{"unit/element-set cache hit?"}
  C -- yes --> O2["return cached built-in graph"]
  C -- no --> B["computeIntraUnitBonds"]
  B --> S["cache built-in graph"]
  S --> O2
```

IMO, this results in minimal changes in Mol* core. I think there is still some complexity with the bond provider registry concept: because it is agnostic of the bond order provider implementation, it can register multiple providers. I don't really see a use for that, and the complexity of having a list of providers registered, one being active, seems dubious, but I may miss something there.

Disclaimer: used an LLM for iterating through the concept.

## Actions


src/mol-model-formats/structure/property/bonds/struct_conn.ts
```diff
@@ -179,6 +181,8 @@ export namespace StructConn {
                 case 'metalc': flags = BondType.Flag.MetallicCoordination; break;
             }
 
+            if (orderUnknown && type === 'covale') flags |= BondType.Flag.Computed;
```
arose, line 184: is this so you know they where not given? we probably need a separate flag for orders

arose (changes requested):
Good call making a POC, this looks promising, thanks.

Let's get rid of the global BondProviderRegistry and move it to the PluginContext. Then add a CustomModelProperty and register it with PluginContext.customModelProperties. The CustomModelProperty shows the available bond providers in a param and returns the provider as a value which can be used in the bond calculation. In principle this could allow switching bond providers at runtime but need to think about how to trigger state re-evaluation.

src/mol-model-formats/structure/property/bonds/struct_conn.ts
papillot: Yes, that was the intent. I'll add a new flag `UnknownOrder`

src/mol-model/structure/structure/unit.ts
```diff
@@ -307,6 +309,9 @@ namespace Unit {
         }
 
         get bonds() {
+            const provided = BondProviderRegistry.get(this.model).active?.getBonds(this);
+            if (provided !== undefined) return provided;
```
dsehnal: I don't think this is a good idea. This means that two calls to `Unit.bonds` can return a different result.

The bond provider needs to be a parameter of the `Unit` constructor and remain static.

The registry needs to be to on `PluginContext`, and all the transform does is resolve the provider instance from the registry on the context, and passes it to Structure constructor, that passes it down to Units.

